PR #8, opened 2019-02-26
Refresh tower-http

carllerche:
This is a work in progress to bring `tower-http` up to date.


tower-http-service/src/lib.rs
```diff
@@ -33,37 +44,20 @@ pub trait HttpService<RequestBody>: Sealed<RequestBody> {
     /// `HttpService` instance needs to be used where a `T: Service` is
     /// required, it must be wrapped with a type that provides that
     /// implementation. `LiftService` does this.
-    fn lift(self) -> LiftService<Self> where Self: Sized {
-        LiftService { inner: self }
+    fn into_service(self) -> util::IntoService<Self> where Self: Sized {
```
LucioFranco: why not just import these instead of using the full path?

tower-http-service/src/util/as_service.rs
```diff
@@ -0,0 +1,36 @@
+use HttpService;
+use futures::Poll;
+use http::{Request, Response};
+use tower_service::Service;
+
+/// Wraps an `HttpService` reference, implementing `tower_service::Service`.
+///
+/// See [`lift`] function documentation for more details.
```
LucioFranco: lift no longer exists?

LucioFranco: prob should be `AsService` instead

tower-http-service/src/lib.rs
carllerche: 🤷‍♂️

Cargo.toml
```diff
@@ -1,16 +1,5 @@
-[package]
-name = "tower-http"
```
GeorgeHahn: I think this broke the `tower-grpc` build - that project still contains references to this package.

carllerche: Yep... I messed that up. Fixes are coming.

carllerche: The recent churn is an effort to get the whole thing on crates.io :) so sorry about that.

src/service.rs
```diff
@@ -1,123 +0,0 @@
...
-        LiftServiceRef { inner: self }
-    }
-}
-
-/// Wraps an `HttpService` instance, implementing `tower_service::Service`.
-///
-/// See [`lift`] function documentation for more details.
-///
-/// [`lift`]: #
-pub struct LiftService<T> {
```
hntd187: Some downstream services like @LucioFranco 's tower-consul breaks with lift service being removed, lift should just be replaced with straight service impls correct?

LucioFranco: lift just got renamed to `AsService` so just need to change the functions we call in tower-consul

hntd187: There is more to it than that, LiftService only has T, where AsService is `AsService<'a, T>`

carllerche: * `LiftService` -> `IntoService`
* `LiftServiceRef` -> `AsService`.

It should have just been a rename. If not, I messed up.

PR #10, opened 2019-03-05
Add Body::is_end_stream

carllerche:
Provides a hint indicating if the Body is done streaming. Being done
implies that the Body value will no longer yield data or trailers.

False negatives are permitted, but not false positives.

### DIFF
--- a/tower-http-service/src/body.rs
+++ b/tower-http-service/src/body.rs
@@ -21,6 +21,17 @@ pub trait Body {
 
     /// Poll for an optional **single** `HeaderMap` of trailers.
     fn poll_trailers(&mut self) -> Poll<Option<HeaderMap>, Self::Error>;
+
+    /// Returns `true` when the end of stream has been reached.
+    ///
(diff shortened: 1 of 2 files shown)


tower-http-service/src/body.rs
```diff
@@ -21,6 +21,17 @@ pub trait Body {
...
     /// Poll for an optional **single** `HeaderMap` of trailers.
     fn poll_trailers(&mut self) -> Poll<Option<HeaderMap>, Self::Error>;
+
+    /// Returns `true` when the end of stream has been reached.
+    ///
+    /// An end of stream means that both `poll_buf` and `poll_trailers` will
+    /// return `None`.
+    ///
+    /// A return value of `false` **does not** guarantee that a value will be
+    /// returend from `poll_stream` or `poll_trailers`.
```
seanmonstar:
```suggestion
    /// returned from `poll_stream` or `poll_trailers`.
```

PR #11, opened 2019-03-06
add-origin: use tower-service from crates.io

seanmonstar:


carllerche:
CI should be happy once https://github.com/tower-rs/tower/pull/186 is merged.

PR #12, opened 2019-03-21
Add middleware for constant token based auth

GeorgeHahn:
Add support for constant token based auth - this is useful for scenarios where the auth token is exchanged out of band.

This does not currently support scenarios where the token needs to change periodically, like login and token refresh, but should be adaptable to those cases when the need arises.


seanmonstar:
Just noting here: it seems this could be about adding a constant *header*, not just token, ya?

GeorgeHahn:
That's a good point! Nothing here makes this token-specific.

I should also note that it could be desirable to make the header value mutable (maybe even hold a whole `HeaderMap` and offer ref/mut access?). That's well beyond what I need, but I think that ability could be beneficial in some scenarios.

carllerche:
Thanks for the PR, great stuff!

At the risk of blowing up scope... it seems like the feature if adding a const header could be merged with “add-origin” to have a single middleware to modify the request in a constant way :-)

GeorgeHahn:
I like that idea - a middleware for request modification would be useful.

What do you think about a builder that accepts a `fn(Request) -> Request`? This should give a lot of dev flexibility.

The new `AddOrigin` implementation would look fairly similar to the existing one. It would be nice if we could eliminate the `impl Service` entirely, but I'm not quite sure how to build that in a way that would continue to expose `AddOrigin` and its functions. I'll make some sketches and see what I can get working. I think the fallback option would be to use a boilerplate `impl Service` wrapper around the interceptor's.

```rust
let conn = tower_request_interceptor::Builder::new()
    .with(|req| /* do something interesting with req */ req)
    .build(conn)
    .unwrap();
```

carllerche:
The problem w/ a fn is that the type then cannot be named, thus cannot be used as part of other trait associated types.

I was thinking having a struct that had a declarative list of modifications to make.

Initially, it could start w/ `Vec<Modification>` where:

```rust
enum Modification {
    SetScheme(Scheme),
    SetAuthority(Authority),
    AddHeader(HeaderName, HeaderValue),
}
```

seanmonstar:
You'd rather we have a `Vec` instead of just chaining the middleware multiple times?

carllerche:
@seanmonstar unknown. The vec would slightly simplify the chained result type. Maybe we don’t care about that anymore?

GeorgeHahn:
I sketched out a `Vec<Modification>` implementation. Currently failing the actual behavior tests, but I'm out of time to look into this for tonight.

Here's what the API ends up looking like - what do you think?

```
let conn = Builder::new()
    .set_origin("http://www.example.com/foo")
    .add_header("authorization", "Bearer ee2c2e06-0254-441d-b885-5bade6d7f3b2")
    .add_header("User-Agent", "Mozilla/<version>")
    .build(conn)
    .unwrap();
```

tower-request-modifier/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+extern crate tower_service;
+
+use futures::Poll;
+use http::header::{HeaderName, HeaderValue};
+use http::{Request, HttpTryFrom};
+use http::uri::{self, Authority, Scheme, Uri};
+use tower_service::Service;
+
+#[derive(Debug, Clone)]
+pub enum Modification {
```
carllerche: I would try to keep this enum private. Doing so would allow us to add modifications without a breaking change.

tower-request-modifier/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Create a new `RequestModifier`
+    pub fn new(inner: T, modifications: Vec<Modification>) -> Self {
+        RequestModifier {
+            inner: inner,
+            modifications: modifications,
+        }
+    }
+
+    /// Return a reference to the modifications that are performed on all requests.
+    pub fn modifications(&self) -> &Vec<Modification> {
```
carllerche: I would probably not include this function in the public API. This ties the struct to the `Vec<Modification>` internal representation. I also don't think we need this.

davidbarsky:
@carllerche 

> Also, looking at the implementation... I wonder if we could generalize it further. What if the modification is represented as Fn(&mut Request<_>) + Send + Sync. Then, the modifiers could be stored as:
>
> `modifiers: Arc<Vec<Box<Fn(&mut Request<_>) + Send + Sync>>>`

I'm a fan of this approach—this is something that I think I'd find _very_ useful.
```

**Comment by GeorgeHahn** [comment]
I just pushed a `Fn(Request<B>) -> Request<B>` implementation. I kept commit history around for comparison, but I can squash it all together if this approach looks good.

I'm pretty happy with how this turned out, though I'm not certain I handled the generics as well as I could have.

File: tower-request-modifier/tests/request_modification.rs [hunk 3]
```diff
@@ -0,0 +1,136 @@
...
+
+    // Make everything happy:
+    let response = Response::builder().status(204).body(());
+
+    send_response.respond(response);
+}
+
+#[test]
+fn run_arbitrary_modifier() {
+    let (mock, mut handle) = Mock::new();
```
davidbarsky: (Mobile, sorry!) Tower just switched to Tower-test, with mock being exported. Mock::new’s been replaced by Mock::pair (with some friendlier  method names, too). Consider delete your cargo.lock and pulling in those (breaking) changes?

carllerche: (sorry)

tower-request-modifier/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl<T, B> std::fmt::Debug for RequestModifier<T, B> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        writeln!(f, "RequestModifier with {} modifiers", self.modifiers.len())
+    }
+}
+
+/// Configure an `RequestModifier` instance
+#[derive(Default)]
+pub struct Builder<B: Default> {
```
carllerche: I don't think the `B: Default` bound is needed. Instead, don't derive `Default`, but manually implement.

tower-request-modifier/tests/request_modification.rs
GeorgeHahn: Ah, thank you!

GeorgeHahn:
Thanks for the reviews! Made fixes and changes as requested and squashed.

PR #13, opened 2019-04-04
Spike `Client`

carllerche:
Work in progress... Experimenting with an HTTP client shim.

### DIFF
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -1,0 +1,44 @@
+extern crate futures;
+extern crate http;
+extern crate tower_http_service;
+extern crate tower_util;
+
+use futures::{Future, Poll};
+use http::{Request, Response};
+use http::uri::Uri;
+use tower_http_service::HttpService;
+use tower_http_service::util::{IntoService};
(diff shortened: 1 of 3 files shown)


tower-http/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl<T> Client<T> {
+    pub fn get<RequestBody>(&self, uri: Uri) -> ResponseFuture<T, RequestBody>
+    where
+        T: HttpService<RequestBody> + Clone,
+        RequestBody: Default
+    {
+        let request = Request::get(uri).body(Default::default()).unwrap();
+        let inner = Oneshot::new(self.service.clone().into_service(), request);
```
davidbarsky: I'm somewhat curious why we want a Oneshot service here—is it expected that we'd receive a `MakeService`, which will implement `IntoService`?

carllerche: The strategy used here is to clone the service from the client shim and move it into the response future. `Oneshot` does this.

davidbarsky: Fair enough, thanks!

PR #17, opened 2019-04-10
Fix request modifier build

GeorgeHahn:
I accidentally left the `Default` restriction on `Builder`'s generic parameter. This pushed a `Default` requirement down to the request body (`BoxBody` for `tower-grpc`). This is fixed in 62b96cf81cb21870039df15c0bd5860144ef8c34.

I'm also seeing an issue where the grpc client connection (`GrpcConnection<RequestModifier<...>>` (full type below) cannot be cloned. I'm trying to figure out how to fix this but have not found a solution.

Full type of the grpc client connection:
`GrpcConnection<RequestModifier<tower_h2::client::connection::Connection<tokio_tcp::stream::TcpStream, tokio::runtime::threadpool::task_executor::TaskExecutor, tower_grpc::body::BoxBody>, tower_grpc::body::BoxBody>>`

### DIFF
--- a/tower-request-modifier/src/lib.rs
+++ b/tower-request-modifier/src/lib.rs
@@ -10,14 +10,12 @@ use std::sync::Arc;
 use tower_service::Service;
 
 /// Wraps an HTTP service, injecting authority and scheme on every request.
-pub struct RequestModifier<T, B>
-{
+pub struct RequestModifier<T, B> {
     inner: T,
     modifiers: Arc<Vec<Box<dyn Fn(Request<B>) -> Request<B> + Send + Sync>>>,
 }
 
-impl<T, B> std::fmt::Debug for RequestModifier<T, B>
-{
+impl<T, B> std::fmt::Debug for RequestModifier<T, B> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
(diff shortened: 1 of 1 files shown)


tower-request-modifier/src/lib.rs
```diff
@@ -10,13 +10,20 @@ use std::sync::Arc;
 use tower_service::Service;
 
 /// Wraps an HTTP service, injecting authority and scheme on every request.
-#[derive(Clone)]
-pub struct RequestModifier<T, B> {
+pub struct RequestModifier<T, B>
+where
+    T: Clone,
```
carllerche: You should not need `where` bounds here. The `Clone` bound can be isolated to the `Clone` impl

tower-request-modifier/src/lib.rs
```diff
@@ -94,9 +106,22 @@ where
     }
 }
 
+impl<T, B> Clone for RequestModifier<T, B>
+where
+    T: Clone,
+    B: Clone,
```
carllerche: `B` should not require a `Clone` bound.

carllerche:
Looks like it needs a `rustfmt` pass.

GeorgeHahn:
Whoops, sorry about that. I can also squash this down if it looks good. I'm still having trouble cloning the client in my repo, but I think that's going to end up being a type issue in my code.

carllerche:
No need to squash. I can do that on merge.

PR #19, opened 2019-04-29
Introduce tower-http-tls

LucioFranco:
This is the initial workings of providing a `TlsConnector` for `tower-http` transports. It introduces three new things:

- `HttpConnection` which provides http based info about a connection. @seanmonstar you may know of more items to add to this trait?
- `HttpMakeConnection` is a trait alias for `MakeConnection` that produces `HttpConnection` connections.
- `TlsConnector` is a `HttpMakeConnection` impl that provides a `TlsStream` wrapper that implements `HttpConnection`.

### DIFF
--- a/tower-http-tls/Cargo.toml
+++ b/tower-http-tls/Cargo.toml
@@ -20,6 +20,6 @@ http = "0.1"
 tower = "0.1"
 tokio-buf = "0.1"
 tower-hyper = { path = "../../tower-hyper" }
-# tower-http = { path = "../tower-http" }
-tower-http = { git = "https://github.com/tower-rs/tower-http" }
+tower-http = { path = "../tower-http" }
+# tower-http = { git = "https://github.com/tower-rs/tower-http" }
 hyper = "0.12"
--- a/tower-http-tls/Cargo.toml
+++ b/tower-http-tls/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "tower-http-tls"
+version = "0.1.0"
+authors = ["Lucio Franco <luciofranco14@gmail.com>"]
+
+[dependencies]
+futures = "0.1"
+tower-http-util = { path = "../tower-http-util" }
+rustls = "0.15"
+tokio-rustls = "0.10.0-alpha.2"
+tokio-tcp = "0.1"
+tokio-io = "0.1"
+tower-service = "0.2"
+webpki = "0.19"
+webpki-roots = "0.16"
+http = "0.1"
+# trust-dns = "0.16"
+
+[dev-dependencies]
+tower = "0.1"
+tokio-buf = "0.1"
+tower-hyper = { path = "../../tower-hyper" }
+# tower-http = { path = "../tower-http" }
+tower-http = { git = "https://github.com/tower-rs/tower-http" }
+hyper = "0.12"
--- a/tower-http-tls/src/lib.rs
+++ b/tower-http-tls/src/lib.rs
@@ -1,0 +1,130 @@
+extern crate futures;
+extern crate http;
+extern crate rustls;
+extern crate tokio_io;
+extern crate tokio_rustls;
+extern crate tokio_tcp;
+extern crate tower_http_util;
+extern crate tower_service;
+extern crate webpki;
+extern crate webpki_roots;
+
+use futures::{Future, Poll};
+use http::Version;
+use rustls::{ClientConfig, Session};
+use std::net::{SocketAddr, ToSocketAddrs};
+use std::sync::Arc;
+use tokio_io::{AsyncRead, AsyncWrite};
+use tokio_rustls::client::TlsStream;
+use tokio_tcp::TcpStream;
+use tower_http_util::HttpConnection;
+use tower_service::Service;
+use webpki::DNSNameRef;
+use webpki_roots::TLS_SERVER_ROOTS;
+
+const ALPN_H2: &str = "h2";
+
+pub struct TlsConnector {
+    config: Arc<ClientConfig>,
+}
+
+pub struct TlsConnection {
+    inner: TlsStream<TcpStream>,
+}
+
+impl TlsConnector {
+    pub fn new(config: ClientConfig) -> Self {
+        let config = Arc::new(config);
+        TlsConnector { config }
+    }
+
+    pub fn with_root(h2: bool) -> Self {
+        let mut config = ClientConfig::new();
+        config
+            .root_store
+            .add_server_trust_anchors(&TLS_SERVER_ROOTS);
+
+        if h2 {
+            config.alpn_protocols.push(Vec::from(&ALPN_H2[..]));
+        }
+
+        TlsConnector::new(config)
+    }
+}
+
+impl<Target> Service<Target> for TlsConnector
+where
+    Target: AsRef<str> + 'static,
+{
+    type Response = TlsConnection;
+    type Error = std::io::Error;
+    type Future = Box<Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        Ok(().into())
+    }
+
+    fn call(&mut self, target: Target) -> Self::Future {
+        let addr = target.as_ref().to_socket_addrs().unwrap().next().unwrap();
+        // TODO(lucio): how do we get a generic target that can extract the DNS from it
+        // and still provide a host:port combo to get the TcpStream?
+        let dns = DNSNameRef::try_from_ascii_str("http2.akamai.com") //(target.as_ref())
+            .unwrap()
+            .to_owned();
+        let config = self.config.clone();
+
+        let connect = TcpStream::connect(&addr)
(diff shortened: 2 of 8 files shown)


LucioFranco:
cc @carllerche @seanmonstar @quinner

LucioFranco:
and also cc @quininer

LucioFranco:
Related [`tower-hyper`](https://github.com/tower-rs/tower-hyper/pull/24) changes for this.

carllerche:
Looks good :+1:.

I'm really going to have to defer to @seanmonstar on the details: re TLS.

I do think that `HttpConnection` should probably exist in its own crate. That can happen later though.

tower-http-tls/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    inner: TlsStream<TcpStream>,
+}
+
+impl TlsConnector {
+    pub fn new(config: ClientConfig) -> Self {
+        let config = Arc::new(config);
+        TlsConnector { config }
+    }
+
+    pub fn with_root(h2: bool) -> Self {
```
brndnmtthws: Should probably allow an option to specify your own root cert, and also a way to pass a client cert/key pair.

LucioFranco: The `new` takes a ClientConfig so you can do it that way. I believe this just pulls from the default Mozilla root cert. Though we could prob add more until constructors around that.

brndnmtthws: Fair enough. I thought I'd mention it anyway since I assume this is for use with gRPC, and MTLS is quite common with gRPC nowadays (for good reason).

tower-http-util/src/connection.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn make_connection(&mut self, target: Target) -> Self::Future;
+}
+
+/// Represents a HTTP aware connection.
+pub trait HttpConnection {
+    /// Returns the version that this stream is set too.
+    fn version(&self) -> Version;
+
+    /// Returns the remote address that this connection is connected to.
+    fn remote_addr(&self) -> std::io::Result<SocketAddr>;
```
seanmonstar: I've always found the `Result` part of this unnecessary, how about just `Option<SocketAddr>`? As an example, a Unix domain socket should be a valid transport, but won't have a `SocketAddr`, so what should it return?

tower-http-util/src/connection.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn poll_ready(&mut self) -> Poll<(), Self::Error>;
+
+    /// Connect and return a transport asynchronously
+    fn make_connection(&mut self, target: Target) -> Self::Future;
+}
+
+/// Represents a HTTP aware connection.
+pub trait HttpConnection {
+    /// Returns the version that this stream is set too.
+    fn version(&self) -> Version;
```
seanmonstar: Is this meant to represent a version determined by ALPN? Maybe it should be `Option<Version>`, then? Since it could be confusing if a connection was made with HTTP2 Prior Knowledge, and yet this says `Version::HTTP_11`.

tower-http-tls/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+                    443
+                } else {
+                    panic!("not a valid http scheme")
+                }
+            } else {
+                panic!("No port provided and no scheme provided")
+            }
+        };
+
+        let addr = (host, port).to_socket_addrs().unwrap().next().unwrap();
```
quininer: This doesn't sound too good, I prefer to have a custom DNS Resolver.

tower-http-tls/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    pub fn with_root(h2: bool) -> Self {
+        let mut config = ClientConfig::new();
+        config
+            .root_store
+            .add_server_trust_anchors(&TLS_SERVER_ROOTS);
+
+        if h2 {
+            config.alpn_protocols.push(Vec::from(&ALPN_H2[..]));
```
quininer: Can we allow both http/1.1 and h2?

tower-http-tls/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    config: Arc<ClientConfig>,
+}
+
+pub struct TlsConnection {
+    inner: TlsStream<TcpStream>,
+}
+
+impl TlsConnector {
+    pub fn new(config: ClientConfig) -> Self {
+        let config = Arc::new(config);
```
quininer: I prefer it.

```rust
impl From<ClientConfig> for TlsConnector {}

impl From<Arc<ClientConfig>> for TlsConnector {}
```

tower-http-tls/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+                panic!("No port provided and no scheme provided")
+            }
+        };
+
+        let addr = (host, port).to_socket_addrs().unwrap().next().unwrap();
+
+        let dns = DNSNameRef::try_from_ascii_str(host).unwrap().to_owned();
+        let config = self.config.clone();
+
+        let connect = TcpStream::connect(&addr)
```
quininer: Should we have a `TcpConnector`? so that we can implement Happy Eyeballs for it.

LucioFranco: Yeah, I think it would be a good idea to actually take a `MakeConnection`? Probably we could provide a `HttpConnector` in `tower-http-util`, then you can pass your own in to do happy eyeballs. I _think_ that would work, thoughts?

tower-http-tls/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+            config.alpn_protocols.push(Vec::from(&ALPN_H2[..]));
+        }
+
+        TlsConnector::new(config)
+    }
+}
+
+impl<Target> Service<Target> for TlsConnector
+where
+    Target: AsRef<str> + 'static,
```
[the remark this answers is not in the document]
LucioFranco: Hmm yeah, should this just take a `Into<Uri>` or a `HttpTryFrom<Uri>` bound?

tower-http-tls/src/lib.rs
quininer: Good idea, so this means that DNS resolve cannot be placed in `TlsConnector`.

tower-http-tls/src/lib.rs
LucioFranco: Yeah, I was holding off on a custom dns for this. I think we may want to introduce a `MakeAddress` trait that is a `Serivce<Target>` where `Service::Response: Into<IpAddr>`?

tower-http-tls/src/lib.rs
LucioFranco: Could you explain why this over just take a `ClientConfig`?

quininer: I have a global `Arc<ClientConfig>` which avoids having to need `Clone` once `ClientConfig` every time a request is create.

tower-http-tls/src/lib.rs
quininer: `Service::Response: Stream<Item=IpAddr>` would be better, or `Iterator<Item=IpAddr>`

LucioFranco:
So some updates, spent some time this evening on the train trying to see if I could make some progress.

First thing, is locally I now have a `HttpConnector` which takes a `D: Service<String, Response = IpAddr>`. We might want to type alias this to a `DnsResolver` trait.

Second, is the `HttpConnector` currently has a custom type `IntoUri` but I am thinking creating a `http-destination` crate that will be similar to `hyper::client::connect::Destination` but for use with general http libraries. This crate will also provide a `DestinationTryInto` trait that will allow the user to pass in a `String`, `&str` etc type into `make_connection`.

Any thoughts around this? I will need a bit more time to finish this as I didnt get as far as I wanted too.

seanmonstar:
> `D: Service<String, Response = IpAddr>`. We might want to type alias this to a `DnsResolver` trait.

Typically with DNS, a DNS name could resolve to multiple addresses. There's `tower-discover` which may be useful for this eventually, but I'd suggest punting worrying about this design *too* much at first.

tower-http-tls/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+tower-service = "0.2"
+webpki = "0.19"
+webpki-roots = "0.16"
+http = "0.1"
+# trust-dns = "0.16"
+
+[dev-dependencies]
+tower = "0.1"
+tokio-buf = "0.1"
+tower-hyper = { path = "../../tower-hyper" }
```
brndnmtthws: This path should probably point to the git repo, yes? i.e.:

```
tower-hyper = { git = "https://github.com/tower-rs/tower-hyper" }
```

I was just trying to test and bumped into this.

LucioFranco:
I'm starting to extract parts of this PR into smaller ones see #22

tower-http-tls/Cargo.toml
LucioFranco: Yeah, should be fixing this soon. once #22 is merged we can depend on the git repo and then work on this independently.

PR #22, opened 2019-05-08
Add `HttpMakeConnection`

LucioFranco:
This is extracted from #19.

This adds just `HttpConnection` and `HttpMakeConnection` based on comments given by @seanmonstar.


LucioFranco:
@seanmonstar I find that name fits better when writing higher level applications like something that would consume tower-grpc 🤷‍♂ .

Also note, I am moving the `HttpConnection` trait to https://github.com/hyperium/http-connection and will update this PR. There are no code changes, it just belongs better in that crate.

PR #30, opened 2019-12-22
Update to `http` v0.2, `http-body` v0.3, `tower-service` v0.3 and…

tesaguri:
This removes `BodyExt::into_buf_stream` since `tokio-buf` crate was removed since `tokio` 0.2, and consequently, `BodyExt` has no methods now. I think there should be an alternative method like `BodyExt::into_stream` or otherwise `BodyExt` should be removed, but that is beyond the scope of this PR.


tower-http-util/Cargo.toml
```diff
@@ -21,10 +21,8 @@ Utilities for working with HTTP services.
...
-futures = "0.1.25"
-http = "0.1.16"
-http-body = "0.1"
+http = "0.2"
+http-body = "0.3"
 http-connection = "0.1.0"
-tokio-buf = "0.1.0"
-tokio-io = "0.1"
-tower-service = "0.2.0"
+tokio = "0.2"
```
LucioFranco:
```suggestion
tokio = { version = "0.2", features = ["io"] }
```

tower-request-modifier/tests/request_modification.rs
```diff
@@ -3,15 +3,17 @@ extern crate tower_request_modifier;
 extern crate tower_service;
 extern crate tower_test;
 
+use futures_test::task::noop_context;
```
LucioFranco: we should attempt to use `tokio-test` here, should be possible to get a context, you can peek at the other tower tests for that.

tower-http-util/Cargo.toml
tesaguri: `tokio` v0.2 does not have have a feature named `io`. Also, ` tower-http-util` only uses `AsyncRead` and `AsyncWrite`, which do not depend on any Cargo features.

tesaguri:
Updated to use `tokio-test`.

tower-http-util/Cargo.toml
LucioFranco: You're right!

PR #33, opened 2021-01-26
Extract middlewares from Embark Studios

davidpdrsn:
- Extracts the middlewares we had written at Embark Studios.
- Setup GH actions.
- Setup cargo deny.
- Write very basic example using hyper.

Might not be feasible to have all the code reviewed in depth but feel free to tear it to shreds.

---

From the readme:

> These are the middlewares included in this crate:
> 
> - `AddExtension`: Stick some shareable value in [request extensions].
> - `Compression`: Apply compression to responses. Uses the `Accept-Encoding` header to pick an encoding. Supports gzip, deflate, brotli, and zstd. Requires hyper because it uses `hyper::Body`.
> - `DebugEnterLeave`: Print when entering and leaving a middleware. Useful for verifying the order in which your middlewares run.
> - `Metrics`: Record high level metrics. Uses the [metrics] crate.
> - `PropagateHeader`: Propagate a header from the request to the response.
> - `SensitiveHeader`: Marks a given header as [sensitive] so it wont show up in logs.
> - `SetResponseHeader`: Set a header on the response.
> - `Trace`: High level tracing of requests and responses.
> - `WrapInSpan`: Wrap response futures in a `tracing::Span`.
> 
> Middlewares uses the [`http`] crate as the HTTP interface so they're compatible with any library or framework that also uses [`http`]. For example hyper and actix.
> 
> All middlewares are disabled by default and can be enabled using a cargo feature. The feature `full` turns on everything.

[`http`]: https://crates.io/crates/http
[sensitive]: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.set_sensitive
[request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
[metrics]: https://crates.io/crates/metrics

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,100 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - master
+  pull_request: {}
+
+jobs:
+  check:
+    # Run `cargo check` first to ensure that the pushed code at least compiles.
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust: [stable, 1.40.0]
+    steps:
+    - uses: actions/checkout@master
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ matrix.rust }}
+        profile: minimal
+    - name: Check
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --all-targets --all-features
+
+  cargo-hack:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        profile: minimal
+    - name: Install cargo-hack
+      run: |
+        curl -LsSf https://github.com/taiki-e/cargo-hack/releases/latest/download/cargo-hack-x86_64-unknown-linux-gnu.tar.gz | tar xzf - -C ~/.cargo/bin
+    - name: cargo hack check
+      working-directory: ${{ matrix.subcrate }}
+      run: cargo hack check --each-feature --no-dev-deps --all
+
+  test-versions:
+    # Test against the stable, beta, and nightly Rust toolchains on ubuntu-latest.
+    needs: check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust: [stable, beta, nightly, 1.40.0]
+    steps:
+    - uses: actions/checkout@master
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ matrix.rust }}
+        profile: minimal
+    - name: Run tests
+      uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --all --all-features
+
+  style:
+    # Check style.
+    needs: check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        components: rustfmt
+        profile: minimal
+    - name: rustfmt
+      uses: actions-rs/cargo@v1
+      with:
+        command: fmt
+        args: --all -- --check
+  
+  # warnings:
+  #   # Check for any warnings. This is informational and thus is allowed to fail.
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #   - uses: actions/checkout@master
+  #   - uses: actions-rs/toolchain@v1
+  #     with:
+  #       toolchain: stable
+  #       components: clippy
+  #       profile: minimal
+  #   - name: Clippy
+  #     uses: actions-rs/clippy-check@v1
+  #     with:
+  #       token: ${{ secrets.GITHUB_TOKEN }}
+  #       args: --all --all-targets --all-features -- -D warnings
+
+  deny-check:
+    name: cargo-deny check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: EmbarkStudios/cargo-deny-action@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,8 +22,13 @@ jobs:
     - name: Check
       uses: actions-rs/cargo@v1
       with:
-        command: check
+        command: clippy
         args: --all --all-targets --all-features
+    - name: rustfmt
+      uses: actions-rs/cargo@v1
+      with:
+        command: fmt
+        args: --all -- --check
 
   cargo-hack:
     runs-on: ubuntu-latest
@@ -58,39 +63,6 @@ jobs:
       with:
         command: test
         args: --all --all-features
-
-  style:
-    # Check style.
-    needs: check
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@master
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        components: rustfmt
-        profile: minimal
-    - name: rustfmt
-      uses: actions-rs/cargo@v1
-      with:
-        command: fmt
-        args: --all -- --check
-  
-  # warnings:
-  #   # Check for any warnings. This is informational and thus is allowed to fail.
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #   - uses: actions/checkout@master
-  #   - uses: actions-rs/toolchain@v1
-  #     with:
-  #       toolchain: stable
-  #       components: clippy
-  #       profile: minimal
-  #   - name: Clippy
-  #     uses: actions-rs/clippy-check@v1
-  #     with:
-  #       token: ${{ secrets.GITHUB_TOKEN }}
-  #       args: --all --all-targets --all-features -- -D warnings
 
   deny-check:
     name: cargo-deny check
--- a/tower-http-examples/Cargo.toml
+++ b/tower-http-examples/Cargo.toml
@@ -3,6 +3,7 @@ name = "tower-http-examples"
 version = "0.1.0"
 authors = ["David Pedersen <david.pdrsn@gmail.com>"]
 edition = "2018"
+license = "MIT"
 publish = false
 
 [dependencies]
--- a/tower-http-examples/Cargo.toml
+++ b/tower-http-examples/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "tower-http-examples"
+version = "0.1.0"
+authors = ["David Pedersen <david.pdrsn@gmail.com>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+hyper = { version = "0.14", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
+tower = { version = "0.4", features = ["util"] }
+tower-http = { path = "../tower-http", features = ["full"] }
+uuid = { version = "0.8", features = ["v4"] }
+tracing = "0.1"
+tracing-subscriber = "0.2"
--- a/tower-http-examples/README.md
+++ b/tower-http-examples/README.md
@@ -1,0 +1,3 @@
+# tower-http-examples
+
+This crate contains a complete (currently work in progress) example of how to use tower-http.
--- a/tower-http-examples/src/main.rs
+++ b/tower-http-examples/src/main.rs
@@ -1,0 +1,90 @@
+#![allow(dead_code, unused_imports)]
+
+use hyper::{
+    header::{self, HeaderName, HeaderValue},
+    service::{make_service_fn, service_fn},
+    Body, Request, Response, Server,
+};
+use std::{convert::Infallible, sync::Arc};
+use tower::ServiceBuilder;
+use tower_http::{
+    add_extension::AddExtensionLayer,
+    compression::CompressionLayer,
+    metrics::MetricsLayer,
+    propagate_header::PropagateHeaderLayer,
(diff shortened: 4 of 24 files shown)


.github/workflows/CI.yml
```diff
@@ -0,0 +1,100 @@
...
+    - uses: actions/checkout@master
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ matrix.rust }}
+        profile: minimal
+    - name: Check
+      uses: actions-rs/cargo@v1
+      with:
+        command: check
+        args: --all --all-targets --all-features
```
Jake-Shadle, line 26: I'd use clippy here instead.

Jake-Shadle: Also no need to use the cargo action, can just call cargo directly.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,100 @@
...
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        components: rustfmt
+        profile: minimal
+    - name: rustfmt
+      uses: actions-rs/cargo@v1
+      with:
+        command: fmt
+        args: --all -- --check
```
Jake-Shadle, line 77: Seems excessive to have a completely different runner for something that will take more time to spinup than actually run, I usually put it as part of the same job that runs the clippy check.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,100 @@
...
+  #   - uses: actions-rs/toolchain@v1
+  #     with:
+  #       toolchain: stable
+  #       components: clippy
+  #       profile: minimal
+  #   - name: Clippy
+  #     uses: actions-rs/clippy-check@v1
+  #     with:
+  #       token: ${{ secrets.GITHUB_TOKEN }}
+  #       args: --all --all-targets --all-features -- -D warnings
```
Jake-Shadle, line 93: I guess this is the clippy check, but should either be removed or uncommented.

Jake-Shadle (changes requested):
Don't really have any comments on the code itself since I'm not familiar with it yet and don't have strong opinions.

.github/workflows/CI.yml
davidpdrsn: 👍 

I copied this setup from tower, maybe I should clean that up as well 🤔

.github/workflows/CI.yml
davidpdrsn: Makes sense 👍

.github/workflows/CI.yml
davidpdrsn: 👍

lpil (approved):
I dig it! Very clear and easy to read. It would be grand to get documentation for each middleware, but that could perhaps come later.

tower-http/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus {
+        self(result)
+    }
+}
+
+mod common {
+    //! Common types that middleware modules use. Just so we don't have to import them all the
+    //! time.
+
+    #![allow(unreachable_pub)]
```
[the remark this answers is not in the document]
davidpdrsn: 😅

tower-http/src/metrics.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        InFlightRequests::decrement();
+
+        Poll::Ready(result)
+    }
+}
+
+struct InFlightRequests;
+
+static IN_FLIGHT_REQUESTS: AtomicU32 = AtomicU32::new(0);
```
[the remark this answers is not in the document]
davidpdrsn: Ah good suggestion. I think that makes sense to support. Noted!

davidpdrsn:
> It would be grand to get documentation for each middleware, but that could perhaps come later.

I imagine once we've agreed on the general interfaces and patterns then we can start writing docs. Also have 0 tests so far.

tower-http/src/metrics.rs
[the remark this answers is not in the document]
davidpdrsn: Fixed in 6d77b1c5a4346. No more global state 😊

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,131 @@
+use crate::common::*;
+
+// TODO(david): don't derive Debug. M is likely to be a closure
```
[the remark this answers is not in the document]
davidpdrsn: Fixed in 5f3e2c5

tower-http/src/compression.rs
```diff
@@ -0,0 +1,214 @@
...
+                Encoding::Zstd => {
+                    let stream = FramedRead::new(
+                        ZstdEncoder::new(StreamReader::new(stream)),
+                        BytesCodec::new(),
+                    );
+                    Body::wrap_stream(stream)
+                }
+                Encoding::Identity => Body::wrap_stream(stream),
+            }
+        });
```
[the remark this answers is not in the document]
lpil: Alas! Thanks for trying

tower-http-examples/README.md
```diff
@@ -0,0 +1,3 @@
+# tower-http-examples
+
+This crate contains a complete (currently work in progress) example of how to use tower-http.
```
LucioFranco, line 3:
```suggestion
This crate contains a complete (currently work in progress) example of how to use `tower-http`.
```

tower-http-examples/src/main.rs
```diff
@@ -0,0 +1,90 @@
...
+            .override_existing(false)
+            .debug_enter_leave("set-response-header"),
+        )
+        .layer(
+            PropagateHeaderLayer::new(HeaderName::from_static(REQUEST_ID))
+                .debug_enter_leave("request-id"),
+        )
+        .layer(CompressionLayer::new().debug_enter_leave("compression"))
+        .layer(AddExtensionLayer::new(state).debug_enter_leave("state"))
+        .service(service_fn(handle));
```
LucioFranco, line 67: I wonder if we should add a `service_fn` to `ServiceBuilder`?

tower-http-examples/src/main.rs
```diff
@@ -0,0 +1,90 @@
+#![allow(dead_code, unused_imports)]
```
LucioFranco, line 1: Why do we need this?

tower-http-examples/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "tower-http-examples"
+version = "0.1.0"
+authors = ["David Pedersen <david.pdrsn@gmail.com>"]
```
LucioFranco, line 4: should we use the same authors here as in `tower-http`?

tower-http/Cargo.toml
```diff
@@ -0,0 +1,63 @@
...
+name = "tower-http"
+description = "Tower middlewares and utilities for HTTP clients and servers"
+version = "0.1.0"
+authors = ["Tower Maintainers <team@tower-rs.com>"]
+edition = "2018"
+license = "MIT"
+readme = "README.md"
+repository = "https://github.com/tower-rs/tower-http"
+homepage = "https://github.com/tower-rs/tower-http"
+documentation = "https://docs.rs/tower-http"
```
LucioFranco, line 11: This should point to the correct version

tower-http/Cargo.toml
```diff
@@ -0,0 +1,63 @@
...
+categories = ["asynchronous", "network-programming", "web-programming"]
+keywords = ["io", "async", "non-blocking", "futures", "service", "http"]
+
+[dependencies]
+futures-util = "0.3"
```
LucioFranco, line 16: remove default features here, and only pull in what we need?

tower-http/src/add_extension.rs
```diff
@@ -0,0 +1,51 @@
+use crate::common::*;
```
LucioFranco, line 1: I would like to avoid any `*` imports

tower-http/src/add_extension.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Clone, Copy, Debug)]
+pub struct AddExtension<S, T> {
+    inner: S,
+    value: T,
+}
+
+impl<ResBody, S, T> Service<Request<ResBody>> for AddExtension<S, T>
+where
+    S: Service<Request<ResBody>>,
+    T: 'static + Clone + Send + Sync,
```
LucioFranco, line 37: minor nit: but I prefer this style 
```suggestion
    T: Clone + Send + Sync + 'static,
```

The order I like is Non auto trait > auto trait > lifetime

tower-http/src/compression.rs
```diff
@@ -0,0 +1,214 @@
+use crate::common::*;
+use async_compression::tokio::bufread::BrotliEncoder;
+use async_compression::tokio::bufread::DeflateEncoder;
+use async_compression::tokio::bufread::GzipEncoder;
+use async_compression::tokio::bufread::ZstdEncoder;
```
LucioFranco, line 5: combine these?

tower-http/src/compression.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Debug)]
+pub struct ResponseFuture<F> {
+    #[pin]
+    future: F,
+    encoding: Encoding,
+}
+
+impl<F, E> Future for ResponseFuture<F>
+where
+    F: Future<Output = Result<Response<Body>, E>>,
```
LucioFranco, line 71: I think this is fine for now but I want to consider if we could maybe impl this over `impl Body<Buf = Bytes>` instead of `hyper::Body`. This way we could eventually add this into tonic quite easily. That is because in tonic we impl body like so https://docs.rs/tonic/0.4.0/tonic/body/struct.BoxBody.html#impl-Body

tower-http/src/lib.rs
```diff
@@ -0,0 +1,145 @@
+// #![doc(html_root_url = "https://docs.rs/tower-http/0.1.0")]
```
LucioFranco, line 1: why is this commented out?

tower-http/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum LatencyUnit {
+    Millis,
+    Nanos,
+}
+
+/// You might always want to log the literal HTTP status. gRPC for example its own status and
+/// always uses `200 OK` even for errors.
+// TODO(david): can we come up with a better name for this?
+pub trait GetTraceStatus<T, E> {
```
LucioFranco, line 92: @eliza and @davidbarsky

tower-http/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// TODO(david): can we come up with a better name for this?
+pub enum TraceStatus {
+    Status(u16),
+    Error,
+}
+
+#[derive(Copy, Clone)]
+pub struct GetTraceStatusFromHttpStatus;
```
LucioFranco, line 103: We likely want to make sure these can't be manually constructed.

```suggestion
pub struct GetTraceStatusFromHttpStatus(());
```

olix0r:
This is a pretty large PR. Is it possible to add modules individually so we can better review what problem each module solves? Or are these modules all interdependent?

tower-http/src/metrics.rs
```diff
@@ -0,0 +1,337 @@
+use crate::{common::*, GetTraceStatus, GetTraceStatusFromHttpStatus, LatencyUnit, TraceStatus};
+use http::{Method, Version};
+use metrics::{gauge, histogram, increment_counter, SharedString};
```
LucioFranco, line 3: I'd be curious if we could implement this with a trait and optionally pull in metrics? That way I can implement it for some of our internal metrics stuff at work :)

LucioFranco (commented):
I only got to `metrics.rs` file but overall looking quite good! 

A few things:

1) I would like to avoid glob imports
2) There are some missing docs which we should add before we publish.

I agree with what @olix0r says and this would be good to break up into smaller PRs (also so firefox stops freezing one me 😆 ).

Anyways, thanks so much for donating this code! Much love to Embark!

davidpdrsn:
Only `Trace` and `Metrics` share a few types, the others have no dependencies on each other.

tower-http/src/metrics.rs
olix0r: Similarly, we don't use the metrics library. I think it would be unfortunate if a general `tower-http` crate set of utilities pulled in specific external implementations like this. This seems better suited to its own crate, imo.

davidpdrsn:
I'm cool with submitting PRs for each middleware but `Trace` and `Metrics` is where most of the complexity is. The others are fairly small. What would you prefer @LucioFranco @olix0r ?

LucioFranco:
@davidpdrsn if the PRs are small I don't mind. I would say maybe break this one up into chunks of 500 loc PRs? Might be easier to review

@olix0r yeah, I think maybe a trait that you can implement on the consumer side could work.

olix0r:
@davidpdrsn I don't want to create more work for you, but I generally have a preference that a single PR address relatively few PRs. The main incentive for you might be that we can merge uncontroversial things while we hash out the details of some of the hairier components like metrics.

davidpdrsn:
It’s no big deal. I’ll address the feedback from @LucioFranco and break stuff into smaller PRs. ETA tomorrow.

tower-http/src/lib.rs
lpil: ooh nice trick!

PR #35, opened 2021-01-26
Add `SetSensitiveHeader` middleware

davidpdrsn:
## What it does

Calls [`HeaderValue::set_sensitive`](https://docs.rs/http/0.2.3/http/header/struct.HeaderValue.html#method.set_sensitive) for some header. Useful for example with `Authorization`.

## Things to consider for other middlewares

I haven't yet written any docs. My initial idea was to get things merged first and then do follow up PRs that add documentation. Otherwise we might spend time writing docs for something only to have it changed during review. I totally wouldn't release things without all public items having documentation.

Another thing I ran into while building things: In how many different ways should we allow users to build/customize their middlewares? I started having both `SensitiveHeaderLayer::new(header).layer(svc)` and `SensitiveHeader::new(svc, header)` for construction.

However for middlewares with lots of customization I would like a builder API a la 

```rust
MetricsLayer::default()
    .record_path_label(true)
    .record_method_label(true)
    .layer(svc)
```

But that kinda means we need to have identical builder methods for those who don't use layers and instead do `Metrics::new(svc)`. That turned into a lot of duplicate builder methods which I got a bit tired of. So I decided that layers are the only way you can wrap services in middlewares. So I stopped defining `new` methods for the services themselves. That removes a bunch of questions/decisions.

What do you think? Is that too opinionated? I personally dig layers and `ServiceBuilder` but I also understand why direct service constructors are nice.

We could have some fairly simple `macro_rules!` macros for defining builder methods but macros tend to be divisive.

### DIFF
--- a/tower-http/src/sensitive_header.rs
+++ b/tower-http/src/sensitive_header.rs
@@ -1,0 +1,88 @@
+use futures_util::ready;
+use http::{header::HeaderName, Request, Response};
+use pin_project::pin_project;
+use std::future::Future;
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Clone, Debug)]
+pub struct SensitiveHeaderLayer {
+    header: HeaderName,
+}
+
+impl SensitiveHeaderLayer {
+    pub fn new(header: HeaderName) -> Self {
+        Self { header }
+    }
+}
+
+impl<S> Layer<S> for SensitiveHeaderLayer {
+    type Service = SensitiveHeader<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        SensitiveHeader {
+            inner,
+            header: self.header.clone(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct SensitiveHeader<S> {
+    inner: S,
+    header: HeaderName,
+}
+
+impl<ReqBody, ResBody, S> Service<Request<ReqBody>> for SensitiveHeader<S>
+where
+    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
+        if let Some(value) = req.headers_mut().get_mut(&self.header) {
+            value.set_sensitive(true);
+        }
+
+        ResponseFuture {
+            future: self.inner.call(req),
+            header: Some(self.header.clone()),
+        }
+    }
+}
+
+#[pin_project]
+#[derive(Debug)]
+pub struct ResponseFuture<F> {
+    #[pin]
+    future: F,
+    header: Option<HeaderName>,
+}
+
+impl<F, ResBody, E> Future for ResponseFuture<F>
+where
+    F: Future<Output = Result<Response<ResBody>, E>>,
+{
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let mut res = ready!(this.future.poll(cx)?);
+
+        if let Some(value) = res.headers_mut().get_mut(this.header.take().unwrap()) {
+            value.set_sensitive(true);
+        }
+
+        Poll::Ready(Ok(res))
+    }
+}
--- a/tower-http/src/sensitive_header.rs
+++ b/tower-http/src/sensitive_header.rs
@@ -165,7 +165,7 @@ pub struct SetSensitiveResponseHeader<S> {
 }
 
 impl<S> SetSensitiveResponseHeader<S> {
-    /// Create a new [`SetSensitiveResponseHeader`].
+    /// Create a new [`SetSensitiveResponseHeader`] service.
     pub fn new(inner: S, header: HeaderName) -> Self {
         Self { inner, header }
     }
--- a/tower-http/src/sensitive_header.rs
+++ b/tower-http/src/sensitive_header.rs
@@ -1,3 +1,7 @@
+//! Middlewares that mark headers as [sensitive].
+//!
+//! [sensitive]: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.set_sensitive
+
 use futures_util::ready;
 use http::{header::HeaderName, Request, Response};
 use pin_project::pin_project;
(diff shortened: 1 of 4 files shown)


LucioFranco (approved):

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::future::Future;
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Clone, Debug)]
+pub struct SensitiveHeaderLayer {
```
olix0r, line 13: question: should this be called SetSensitiveHeaderLayer? Might we have an UnsetSensitiveHeader?

Is it always appropriate to do this on both requests and responses? Should this layer hold a list of headers? Maybe a list for requests and a list for responses?

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl<ReqBody, ResBody, S> Service<Request<ReqBody>> for SensitiveHeader<S>
+where
+    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
olix0r, line 48:
```suggestion
    #[inline]
    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```

tower-http/src/sensitive_header.rs
davidpdrsn: I think renaming it to `SetSensitiveHeaderLayer` and `SetSensitiveHeader` is a good idea. I'll change that.

Not sure `UnsetSensitiveHeader` would be very useful as that is the default state and I don't quite see why you'd wanna make a header sensitive and then revert that later.

I think its also fine setting it on both the request and the response in one go. I'm thinking if some header is sensitive it'll be so regardless if its on request or response.

I actually thought setting it on the response is a bit odd, since I this is likely to be outer most middleware such that request headers are immediately marked as sensitive, before they reach logging/tracing. However if its the outer most one it'll also be the last to see the response which make marking response header less useful. I do think its slightly safer to keep it like this though.

davidpdrsn: @olix0r wanna have another look?

hawkw: > Not sure `UnsetSensitiveHeader` would be very useful as that is the default state and I don't quite see why you'd wanna make a header sensitive and then revert that later.

I think that what @olix0r meant was a layer that _removes a header_ rather than a layer that unsets the sensitive flag on that header.

hawkw: > question: should this be called SetSensitiveHeaderLayer? Might we have an UnsetSensitiveHeader?
> 
> Is it always appropriate to do this on both requests and responses? Should this layer hold a list of headers? Maybe a list for requests and a list for responses?

IMO it might be worth having separate middleware for setting sensitive headers on requests and on responses? There could be a layer that produces each middleware, and a layer that produces both in one go...

davidpdrsn: Ah yeah thats a good idea. I made some changes.

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl SetSensitiveHeaderLayer {
+    /// Create a new [`SetSensitiveHeaderLayer`].
+    pub fn new(header: HeaderName) -> Self {
+        Self { header }
+    }
+}
+
+impl<S> Layer<S> for SetSensitiveHeaderLayer {
+    type Service = SetSensitiveRequestHeader<SetSensitiveResponseHeader<S>>;
```
hawkw, line 32: nit/TIOLI: we could also expose a type alias
```rust
pub type SetSensitiveHeader<S> = SetSensitiveRequestHeader<SetSensitiveResponseHeader<S>>;
```
so that users can more easily refer to this without having to type out both service's names...

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn layer(&self, inner: S) -> Self::Service {
+        SetSensitiveRequestHeader::new(
+            SetSensitiveResponseHeader::new(inner, self.header.clone()),
+            self.header.clone(),
+        )
+    }
+}
+
+/// Mark a request header as [sensitive].
```
hawkw, line 42: it might be nice if these docs included a link to the service this layer produces? not a big deal though.

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn layer(&self, inner: S) -> Self::Service {
+        SetSensitiveResponseHeader {
+            inner,
+            header: self.header.clone(),
+        }
+    }
+}
+
+/// Mark a response header as [sensitive].
+///
```
hawkw, line 148: similarly, a link to the produced service could be nice

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,224 @@
...
+/// [sensitive]: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.set_sensitive
+#[derive(Clone, Debug)]
+pub struct SetSensitiveResponseHeader<S> {
+    inner: S,
+    header: HeaderName,
+}
+
+impl<S> SetSensitiveResponseHeader<S> {
+    /// Create a new [`SetSensitiveResponseHeader`].
```
hawkw, line 157: nit, take it or leave it:
```suggestion
    /// Create a new [`SetSensitiveResponseHeader`] service.
```

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+/// Response future for [`SetSensitiveResponseHeader`].
+#[pin_project]
+#[derive(Debug)]
+pub struct SetSensitiveResponseHeaderResponseFuture<F> {
+    #[pin]
+    future: F,
+    header: Option<HeaderName>,
```
hawkw, line 205: i'm not sure if i get why this needs to be an option. we only need to borrow it to call `get_mut`, we don't need to move it by value --- can't this just be a `HeaderName`? if we did that, we could avoid the `take()` and `unwrap()` calls in the `Future` impl, which would make the code a little shorter (and not generate a panic that never actually occurs)...

hawkw (commented):
Overall this looks about right! I had a couple of docs nits, and I think the future implementation could be simplified a little bit.

tower-http/src/sensitive_header.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    /// Gets a mutable reference to the underlying service.
+    pub fn get_mut(&mut self) -> &mut S {
+        &mut self.inner
+    }
+
+    /// Consumes `self`, returning the underlying service.
+    pub fn into_inner(self) -> S {
+        self.inner
+    }
```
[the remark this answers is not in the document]
davidpdrsn: Fun fact. I actually had such a macro way back but I decided to remove it since macros aren't everyones cup of tea. But I think tidying this up with a macro would be cool. I'm gonna make a mental note of that.

tower-http/src/sensitive_header.rs
davidpdrsn: 👍

tower-http/src/sensitive_header.rs
davidpdrsn: Good catch. `res.headers_mut().get_mut(&*this.header)` did the trick.

PR #37, opened 2021-01-26
Add `Compression` middleware

davidpdrsn:
Suggested by @LucioFranco in https://github.com/tower-rs/tower-http/pull/33:

> I think this is fine for now but I want to consider if we could maybe impl this over impl Body<Buf = Bytes> instead of hyper::Body. This way we could eventually add this into tonic quite easily. That is because in tonic we impl body like so https://docs.rs/tonic/0.4.0/tonic/body/struct.BoxBody.html#impl-Body

I totally agree but I couldn't quite figure out how to make it work. `Body::wrap_stream` is quite convenient. I could give it another shot.

## TODO

- [x] Add feature toggles for each compression algorithm a la https://github.com/tower-rs/tower-http/pull/41/commits/090053ae3056996e29228f8a93739227c16d01ab#r566686758.


davidpdrsn:
I've put this back into draft mode. https://github.com/tower-rs/tower-http/pull/41 shows how to make decompression work without depending on `hyper::Body`. So I'm gonna copy that approach for this PR.

davidpdrsn:
I've updated this to support any `B: http_body::Body` rather than being hardcoded to `hyper::Body`. Its highly inspired by https://github.com/tower-rs/tower-http/pull/41 so thanks to @tesaguri!

Before merging this we should decide something regarding the questions raised [here](https://github.com/tower-rs/tower-http/pull/41#pullrequestreview-577107333) though.

tower-http/src/compression.rs
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl<B: Body> Stream for Adapter<B, B::Error> {
+    type Item = io::Result<Bytes>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.project();
+
+        match ready!(this.body.poll_data(cx)) {
+            Some(Ok(mut data)) => Poll::Ready(Some(Ok(data.copy_to_bytes(data.remaining())))),
```
LucioFranco, line 272: I wonder if there is a way to optimize this if it impls bytes :(

tower-http/src/compression.rs
```diff
@@ -0,0 +1,385 @@
...
+        }
+    }
+
+    fn poll_trailers(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Option<http::HeaderMap>, Self::Error>> {
+        match self.project().inner.project() {
+            BodyInnerProj::Gzip(framed_read) => framed_read
+                .get_pin_mut()
```
LucioFranco, line 214: lol

LucioFranco (approved):
This looks great!

tower-http/src/compression.rs
tesaguri: I think `<Bytes as Buf>::copy_to_bytes` is already optimized enough. It will become a mere `mem::replace` call if the argument equals `data.remaining()`, without even incrementing the reference count.

<https://github.com/tokio-rs/bytes/blob/master/src/bytes.rs#L551-L559>

PR #38, opened 2021-01-26
Add `PropagateHeader` middleware

davidpdrsn:
Another small one. Propagate a header from the request to the response.

### DIFF
--- a/tower-http/src/propagate_header.rs
+++ b/tower-http/src/propagate_header.rs
@@ -41,6 +41,35 @@ impl<S> Layer<S> for PropagateHeaderLayer {
 ///
 /// So if the header is present on the request it'll be applied to the response as well. This could
 /// for example be used to propagate headers such as `X-Request-Id`.
+///
+/// # Example
+///
+/// ```rust
+/// use http::{Request, Response, header::HeaderName};
+/// use std::convert::Infallible;
+/// use tower::{Service, ServiceExt};
+/// use tower_http::propagate_header::PropagateHeader;
+///
+/// # #[tokio::main]
+/// # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// let echo_service = tower::service_fn(|request: Request<String>| async move {
+///     Ok::<_, Infallible>(Response::new(request.into_body()))
+/// });
+///
+/// // This will copy `x-request-id` headers on requests onto responses.
+/// let mut svc = PropagateHeader::new(echo_service, HeaderName::from_static("x-request-id"));
+///
+/// let request = Request::builder()
+///     .header("x-request-id", "1337")
+///     .body(String::new())?;
+///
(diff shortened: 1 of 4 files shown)


tower-http/src/propagate_header.rs
```diff
@@ -0,0 +1,87 @@
...
+use tower_layer::Layer;
+use tower_service::Service;
+
+#[derive(Clone, Debug)]
+pub struct PropagateHeaderLayer {
+    header: HeaderName,
+}
+
+impl PropagateHeaderLayer {
+    pub fn new(header: HeaderName) -> Self {
```
hawkw, line 18: would be nice to have at least a simple example for this...

tower-http/src/propagate_header.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let value = req.headers().get(&self.header).cloned();
+
+        ResponseFuture {
+            future: self.inner.call(req),
+            header_and_value: Some(self.header.clone()).zip(value),
```
hawkw, line 58: clever :)

tower-http/src/propagate_header.rs
```diff
@@ -0,0 +1,87 @@
...
+    fn layer(&self, inner: S) -> Self::Service {
+        PropagateHeader {
+            inner,
+            header: self.header.clone(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct PropagateHeader<S> {
```
hawkw, line 35: nit: would be nice if this had a constructor for building the service directly. it could also be good to add accessors for the inner service, and maybe also a `PropagateHeader::layer` function?

hawkw (approved):
This looks good to me overall!

It would be nice to have some docs for this one, to make the difference between `SetHeader` and `PropagateHeader` a little clearer. It's not _immediately_ obvious to me what this does from the name --- I kind of assumed it was something related to trace ID header propagation...

tower-http/src/propagate_header.rs
davidpdrsn: Yes totally. Once a few PRs have been merged I'm gonna do a pass where I add inner service accessors and align the layers and services. I just got super tired of writing the same mostly boilerplate code over and over and wanted to initially focus more on the interesting part of the APIs.

tower-http/src/propagate_header.rs
davidpdrsn: I've one now.

Before shipping anything I want to sure the docs are at least decent and that all middlewares have examples.

PR #39, opened 2021-01-26
Add `SetResponseHeader` middleware

davidpdrsn:
We use this to set `X-Request-Id` with something like:

```rust
.layer(
    SetResponseHeaderLayer::new(
        HeaderName::from_static("x-request-id"),
        |_res: &Response<Body>| HeaderValue::from_str(&Uuid::new_v4().to_string()).unwrap(),
    )
    .override_existing(false),
);

// or even

.layer(
    SetResponseHeaderLayer::new(
        HeaderName::from_static("content-type"),
        HeaderValue::from_static("text/html"),
    )
    .override_existing(false),
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ shouldn't be using it yet!).
 
 These are the middlewares included in this crate:
 
-<!-- Other PRs will list things here -->
+- `SetResponseHeader`: Set a header on the response.
 
 Middlewares uses the [`http`] crate as the HTTP interface so they're compatible with any library or framework that also uses [`http`]. For example hyper and actix.
 
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Tower HTTP
 
-Tower utilities for HTTP clients and servers
+Tower middlewares and utilities for HTTP clients and servers
 
-[![Build Status](https://travis-ci.org/tower-rs/tower-http.svg?branch=master)](https://travis-ci.org/tower-rs/tower-http)
+[![Build status](https://github.com/tower-rs/tower-http/workflows/CI/badge.svg)](https://github.com/tower-rs/tower-http/actions)
 
-More information about this crate can be found in the [crate documentation][dox]
+More information about this crate can be found in the [crate documentation][dox].
 
 [dox]: https://tower-rs.github.io/tower-http/tower_http
 
@@ -13,3 +13,17 @@ environment or you will regret it!** This crate is still under active
 development and there has not yet been any focus on documentation (because you
 shouldn't be using it yet!).
 
+## Middlewares
+
+These are the middlewares included in this crate:
+
+<!-- Other PRs will list things here -->
+
+Middlewares uses the [`http`] crate as the HTTP interface so they're compatible with any library or framework that also uses [`http`]. For example hyper and actix.
+
+The middlewares were originally extracted from one of [@EmbarkStudios] internal projects.
+
+All middlewares are disabled by default and can be enabled using a cargo feature. The feature `full` turns on everything.
+
+[`http`]: https://crates.io/crates/http
+[@EmbarkStudios]: https://github.com/EmbarkStudios
--- a/README.md
+++ b/README.md
@@ -20,7 +20,12 @@ These are the middlewares included in this crate:
 - `AddExtension`: Stick some shareable value in [request extensions].
 - `Compression`: Compression response bodies.
 - `Decompression`: Decompress response bodies.
+- `PropagateHeader`: Propagate a header from the request to the response.
+- `SensitiveHeader`: Marks a given header as [sensitive] so it wont show up in logs.
 - `SetResponseHeader`: Set a header on the response.
+- `SetSensitiveRequestHeader`: Marks a given request header as [sensitive].
+- `SetSensitiveResponseHeader`: Marks a given response header as [sensitive].
+- `SetSensitiveeHeader`: Mark a header as [sensitive] on both requests and responses.
 
 Middlewares uses the [`http`] crate as the HTTP interface so they're compatible with any library or framework that also uses [`http`]. For example hyper and actix.
 
@@ -30,3 +35,4 @@ All middlewares are disabled by default and can be enabled using a cargo feature
 
 [`http`]: https://crates.io/crates/http
 [@EmbarkStudios]: https://github.com/EmbarkStudios
+[sensitive]: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.set_sensitive
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ shouldn't be using it yet!).
 
 These are the middlewares included in this crate:
 
+- `AddExtension`: Stick some shareable value in [request extensions].
+- `Compression`: Compression response bodies.
+- `Decompression`: Decompress response bodies.
 - `SetResponseHeader`: Set a header on the response.
 
 Middlewares uses the [`http`] crate as the HTTP interface so they're compatible with any library or framework that also uses [`http`]. For example hyper and actix.
--- a/README.md
+++ b/README.md
@@ -25,7 +25,6 @@ These are the middlewares included in this crate:
 - `SetResponseHeader`: Set a header on the response.
 - `SetSensitiveRequestHeader`: Marks a given request header as [sensitive].
 - `SetSensitiveResponseHeader`: Marks a given response header as [sensitive].
-- `SetSensitiveeHeader`: Mark a header as [sensitive] on both requests and responses.
 
 Middlewares uses the [`http`] crate as the HTTP interface so they're compatible with any library or framework that also uses [`http`]. For example hyper and actix.
 
@@ -36,3 +35,4 @@ All middlewares are disabled by default and can be enabled using a cargo feature
 [`http`]: https://crates.io/crates/http
 [@EmbarkStudios]: https://github.com/EmbarkStudios
 [sensitive]: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.set_sensitive
+[request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
--- a/tower-http/src/set_response_header.rs
+++ b/tower-http/src/set_response_header.rs
@@ -222,7 +222,7 @@ where
 impl<ReqBody, ResBody, S, M> Service<Request<ReqBody>> for SetResponseHeader<S, M>
 where
     S: Service<Request<ReqBody>, Response = Response<ResBody>>,
-    M: MakeHeaderValue<ResBody> + Clone,
+    M: MakeHeaderValue<Response<ResBody>> + Clone,
 {
     type Response = S::Response;
     type Error = S::Error;
@@ -257,7 +257,7 @@ pub struct ResponseFuture<F, M> {
 impl<F, ResBody, E, M> Future for ResponseFuture<F, M>
 where
     F: Future<Output = Result<Response<ResBody>, E>>,
-    M: MakeHeaderValue<ResBody>,
+    M: MakeHeaderValue<Response<ResBody>>,
 {
     type Output = F::Output;
 
@@ -279,7 +279,7 @@ where
     }
 }
 
-/// Trait for producing header values from responses.
+/// Trait for producing header values.
 ///
 /// Used by [`SetResponseHeader`].
 ///
@@ -291,28 +291,28 @@ where
 /// [`SetResponseHeaderLayer`].
 ///
 /// [`HeaderValue`]: https://docs.rs/http/0.2.3/http/header/struct.HeaderValue.html
-pub trait MakeHeaderValue<B> {
-    /// Try to create a header value from the response.
-    fn make_header_value(&mut self, response: &Response<B>) -> Option<HeaderValue>;
+pub trait MakeHeaderValue<T> {
+    /// Try to create a header value from the request or response.
+    fn make_header_value(&mut self, message: &T) -> Option<HeaderValue>;
 }
 
-impl<F, B> MakeHeaderValue<B> for F
+impl<F, T> MakeHeaderValue<T> for F
 where
-    F: FnMut(&Response<B>) -> Option<HeaderValue>,
+    F: FnMut(&T) -> Option<HeaderValue>,
 {
-    fn make_header_value(&mut self, response: &Response<B>) -> Option<HeaderValue> {
-        self(response)
+    fn make_header_value(&mut self, message: &T) -> Option<HeaderValue> {
+        self(message)
     }
 }
 
-impl<B> MakeHeaderValue<B> for HeaderValue {
-    fn make_header_value(&mut self, _response: &Response<B>) -> Option<HeaderValue> {
+impl<T> MakeHeaderValue<T> for HeaderValue {
+    fn make_header_value(&mut self, _message: &T) -> Option<HeaderValue> {
         Some(self.clone())
     }
 }
 
-impl<B> MakeHeaderValue<B> for Option<HeaderValue> {
-    fn make_header_value(&mut self, _response: &Response<B>) -> Option<HeaderValue> {
+impl<T> MakeHeaderValue<T> for Option<HeaderValue> {
+    fn make_header_value(&mut self, _message: &T) -> Option<HeaderValue> {
         self.clone()
     }
 }
--- a/tower-http/src/set_response_header.rs
+++ b/tower-http/src/set_response_header.rs
@@ -55,7 +55,6 @@
 
 use futures_util::ready;
 use http::{header::HeaderName, HeaderValue, Request, Response};
-use http_body::Body;
 use pin_project::pin_project;
 use std::{
     fmt,
@@ -186,7 +185,6 @@ impl<ReqBody, ResBody, S, M> Service<Request<ReqBody>> for SetResponseHeader<S, 
 where
     S: Service<Request<ReqBody>, Response = Response<ResBody>>,
     M: MakeHeaderValue<ResBody> + Clone,
-    ResBody: Body,
 {
     type Response = S::Response;
     type Error = S::Error;
@@ -222,7 +220,6 @@ impl<F, ResBody, E, M> Future for ResponseFuture<F, M>
 where
     F: Future<Output = Result<Response<ResBody>, E>>,
     M: MakeHeaderValue<ResBody>,
-    ResBody: Body,
 {
     type Output = F::Output;
 
@@ -255,9 +252,7 @@ where
 /// can suply one directly to [`SetResponseHeaderLayer`].
 pub trait MakeHeaderValue<B> {
     /// Try to create a header value from the response.
-    fn make_header_value(&mut self, response: &Response<B>) -> Option<HeaderValue>
-    where
-        B: Body;
+    fn make_header_value(&mut self, response: &Response<B>) -> Option<HeaderValue>;
 }
 
 impl<F, B> MakeHeaderValue<B> for F
--- a/tower-http/src/set_response_header.rs
+++ b/tower-http/src/set_response_header.rs
@@ -27,7 +27,8 @@
 //!         // We have to add `::<_, Body>` since Rust cannot infer the body type when
 //!         // we don't use a closure to produce the header value.
 //!         //
-//!         // `if_not_present` will only insert the header if does not already have a value.
+//!         // `if_not_present` will only insert the header if it does not already
+//!         // have a value.
 //!         SetResponseHeaderLayer::<_, Body>::if_not_present(
 //!             header::CONTENT_TYPE,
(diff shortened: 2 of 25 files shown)


hawkw:
> ```rust
> .layer(
>     SetResponseHeaderLayer::new(|_res: &Response<Body>| {
>         let header = HeaderName::from_static("x-request-id");
>         let value = HeaderValue::from_str(&Uuid::new_v4().to_string()).unwrap();
>         (header, value)
>     })
>     .override_existing(false)
> )
> ```

This is an interesting API --- I would have probably started with something a little simpler, where the header name is provided once when the layer is created, and only the value is returned by the function. The approach you've used is more flexible, sicne it allows the header to be determined dynamically based on the content of the response. However, it also might introduce a bit more when the header name isn't constructed using `from_static`. 

Do you think there's much of a use case for determining which header name should be set dynamically? Or would it make more sense to make this slightly more efficient by providing it statically when the layer is constructed? I'm open to being convinced either way, just thought I'd bring that up.

davidpdrsn:
Thats a good point. I actually think its slightly nicer to only have the value be produced by the function. Off the top of my head I cannot think of a use case where you'd wanna add dynamic headers based on the response.

The only immediate downside is that you have to clone the `HeaderName` to move it from the layer into the service and again into the response future. However from poking around in hyper it seems that `HeaderName` is basically a `Bytes` which is cheap to clone. So I think thats fine.

I'll update this with some docs.

hawkw:
Yeah, `HeaderName` should be cheap to clone — in the case where it's constructed from a string at runtime, it should actually be *much* to clone it than cheaper than to call `HeaderName::from_lowercase` in every request!

davidpdrsn:
I've added a more things:

- Docs and an example
- `MakeHeaderValue` is now generic over just the body type instead of the whole request type. I guess it makes sense to fix the request type to `Request` as that was required by the service impl regardless.
- `MakeHeaderValue::make_header_value` now returns an `Option<HeaderValue>`. So if for some reason you cannot produce the header value from the response you can return `None`. For example for trying to set `Content-Length` when the body doesn't have a known size.
- `MakeHeaderValue` is implemented directly for `HeaderValue` for the cases you want to set the header to a static value, for example for `Content-Type: text/html`.

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,271 @@
...
+    {
+        Self {
+            make,
+            header_name,
+            override_existing: true,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Should the header be overriden if the response already contains it?
```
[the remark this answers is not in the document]
hawkw: also, since HTTP headers can have multiple values, I wonder if it's worth having a third "mode" for this, which calls [`HeaderMap::append`](https://docs.rs/http/0.2.3/http/header/struct.HeaderMap.html#method.append) instead of `insert`, which adds a new value for a header regardless of whether any previous values are present or not? that way, there's also a mechanism to add _additional_ values to a header.

(FWIW, an append mode is probably particularly useful for proxies --- you might be forwarding a response from an upstream server, and you want to add your own values to the headers set by that upstream. For example, a response might have come via a _different_ proxy, and already have a `Via: 1.1 router-proxy` or something, and I might want to add `Via: 1.1 tls-proxy`, or whatever)

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,271 @@
...
+///
+/// You shouldn't normally have to implement this trait since its implemented for closures with the
+/// correct type.
+///
+/// It is also implemented directly for `HeaderValue` so if you just want to add a fixed value you
+/// can suply one directly to [`SetResponseHeaderLayer`].
+pub trait MakeHeaderValue<B> {
+    /// Try to create a header value from the response.
+    fn make_header_value(&mut self, response: &Response<B>) -> Option<HeaderValue>;
+}
```
hawkw, line 256: TIOLI: it occurs to me that we _could_ make this a little more generic, like
```rust
pub trait MakeHeaderValue<T> {
    /// Try to create a header value from a request or response.
    fn make_header_value(&mut self, message: T) -> Option<HeaderValue>;
}
```

That way, a similar middleware for setting request headers could take a `MakeHeaderValue<Request<B>>`, and we could use `MakeHeaderValue<Response<B>>` here.

What do you think?

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,271 @@
...
+
+/// Trait for producing header values from responses.
+///
+/// Used by [`SetResponseHeader`].
+///
+/// You shouldn't normally have to implement this trait since its implemented for closures with the
+/// correct type.
+///
+/// It is also implemented directly for `HeaderValue` so if you just want to add a fixed value you
+/// can suply one directly to [`SetResponseHeaderLayer`].
```
hawkw, line 252: in general i think it's good to prefer to avoid second-person ("You ...") in docs. maybe:
```suggestion
/// This trait is implemented for closures with the correct type signature. Typically
/// users will not have to implement this trait for their own types.
///
/// It is also implemented directly for [`HeaderValue`]. When a fixed header value
/// should be added to all responses, it can be  supplied directly to 
/// [`SetResponseHeaderLayer`].
///
/// [`HeaderValue`]: https://docs.rs/http/0.2.3/http/header/struct.HeaderValue.html
```
or something?

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,271 @@
...
+    fn make_header_value(&mut self, response: &Response<B>) -> Option<HeaderValue> {
+        self(response)
+    }
+}
+
+impl<B> MakeHeaderValue<B> for HeaderValue {
+    fn make_header_value(&mut self, _response: &Response<B>) -> Option<HeaderValue> {
+        Some(self.clone())
+    }
+}
```
hawkw, line 271: nit/TIOLI: for completeness sake, what do you think about also adding
```suggestion
}

impl<B> MakeHeaderValue<B> for Option<HeaderValue> {
    fn make_header_value(&mut self, _response: &Response<B>) -> Option<HeaderValue> {
        self.clone()
    }
}
```

this might be handy in cases where there's a fixed value that may or may not be added, and whether or not it's added is determined statically. For example, I might have a config file option turns off adding [`Server`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server) or [`Via`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Via) headers for security reasons. These would be fixed values, but whether or not they're added is determined at runtime.

README.md
```diff
@@ -22,6 +22,7 @@ These are the middlewares included in this crate:
 - `Decompression`: Decompress response bodies.
 - `PropagateHeader`: Propagate a header from the request to the response.
 - `SensitiveHeader`: Marks a given header as [sensitive] so it wont show up in logs.
```
hawkw, line 24: just noticed this --- is this a duplicate of the entry for `SetSensitiveHeader`? we can fix that separately

hawkw (commented):
This is looking really close. I had a few notes. Let me know what you think?

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,271 @@
+//! Set a header on the response.
+//!
+//! # Example
```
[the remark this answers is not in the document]
davidpdrsn: I like it :)

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,271 @@
...
+//! use tower::{Service, ServiceExt, ServiceBuilder};
+//! use tower_http::set_response_header::SetResponseHeaderLayer;
+//! use http_body::Body as _; // for `Body::size_hint`
+//! use hyper::Body;
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! let echo_service = tower::service_fn(|request: Request<Body>| async move {
+//!     Ok::<_, Infallible>(Response::new(request.into_body()))
+//! });
```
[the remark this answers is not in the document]
davidpdrsn: That makes sense. Keeps things a bit more tidy.

tower-http/src/set_response_header.rs
davidpdrsn: I actually write it like that first but changed it. I didn't consider that it might be nice to re-use the trait in another middleware. Makes sense so I'll change it back.

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,318 @@
...
+/// Used by [`SetResponseHeader`].
+///
+/// This trait is implemented for closures with the correct type signature. Typically
+/// users will not have to implement this trait for their own types.
+///
+/// It is also implemented directly for [`HeaderValue`]. When a fixed header value
+/// should be added to all responses, it can be  supplied directly to
+/// [`SetResponseHeaderLayer`].
+///
+/// [`HeaderValue`]: https://docs.rs/http/0.2.3/http/header/struct.HeaderValue.html
```
tesaguri, line 293:
```suggestion
```

This can be an intra-doc link as well.

tower-http/src/set_response_header.rs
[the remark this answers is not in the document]
davidpdrsn: Yeah thats a good idea. I kinda think it makes sense to change the API a bit then. I would find it a bit odd to have `override_existing` and `append`. Those two are mutually exclusive in a way. I changed this a bit to instead take an `InsertHeaderMode` enum.

Let me know what you think 😊

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,340 @@
...
+        self
+    }
+}
+
+impl<S, M> fmt::Debug for SetResponseHeader<S, M>
+where
+    S: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SetResponseHeaderLayer")
```
tesaguri, line 225:
```suggestion
impl<S, M> fmt::Debug for SetResponseHeader<S, M>
where
    S: fmt::Debug,
{
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("SetResponseHeader")
```

Maybe a copy and paste error?

davidpdrsn: Ah yes. Good catch!

davidpdrsn:
Alright I think this is good to go now, given the new `InsertHeaderMode` API is approved.

I made a few more minor changes:

- Changed so it works for any request type. Previously it required `Request<B>` but not much reason for that.
- Added tests for each of the modes.

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+    {
+        Self {
+            make,
+            header_name,
+            mode: InsertHeaderMode::OverrideExisting,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Set which mode to use when inserting the header.
```
hawkw, line 140: nit: "sets which mode" isn't terribly descriptive. I might reword this to focus on what it actually configures, rather than just saying the name of the type.

how about something like this:
```suggestion
    /// Configures how existing header values are handled.
    ///
    /// This takes an [`InsertHeaderMode`] which configures the service's
    /// behavior when other values have previously been set for the same
    /// header. The available options are:
    ///
    /// - `InsertHeaderMode::OverrideExisting` (the default): if a previous
    ///   value exists for the same header, it is removed and replaced with
    ///   the new header value.
    /// - `InsertHeaderMode::SkipIfPresent`: if a previous value exists for
    ///   the header, the new value is not inserted.
    /// - `InsertHeaderMode::Append`: the new header is always added,
    ///   preserving any existing values. If previous values exist, the header
    ///   will have multiple values.
```

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+    ///
+    /// Defaults to [`InsertHeaderMode::OverrideExisting`].
+    pub fn mode(mut self, mode: InsertHeaderMode) -> Self {
+        self.mode = mode;
+        self
+    }
+
+    define_inner_service_accessors!();
+
+    /// Returns a new [`Layer`] that wraps services with a `SetResponseHeaderLayer` middleware.
```
hawkw, line 221:
```suggestion
    /// Returns a new [`Layer`] that wraps services with a `SetResponseHeader` middleware.
```

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+        Self {
+            inner,
+            header_name,
+            make,
+            mode: InsertHeaderMode::OverrideExisting,
+        }
+    }
+
+    /// Set which mode to use when inserting the header.
+    ///
```
hawkw, line 212: as above, maybe something like
```suggestion
    /// Configures how existing header values are handled.
    ///
    /// This takes an [`InsertHeaderMode`] which configures the service's
    /// behavior when other values have previously been set for the same
    /// header. The available options are:
    ///
    /// - `InsertHeaderMode::OverrideExisting` (the default): if a previous
    ///   value exists for the same header, it is removed and replaced with
    ///   the new header value.
    /// - `InsertHeaderMode::SkipIfPresent`: if a previous value exists for
    ///   the header, the new value is not inserted.
    /// - `InsertHeaderMode::Append`: the new header is always added,
    ///   preserving any existing values. If previous values exist, the header
    ///   will have multiple values.
```

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+#[derive(Clone)]
+pub struct SetResponseHeader<S, M> {
+    inner: S,
+    header_name: HeaderName,
+    make: M,
+    mode: InsertHeaderMode,
+}
+
+impl<S, M> SetResponseHeader<S, M> {
+    /// Create a new [`SetResponseHeader`].
```
hawkw, line 201: This should probably say something like
```suggestion
    /// Create a new [`SetResponseHeader`].
    ///
    /// By default, the layer will construct services configured with
    /// [`InsertHeaderMode::OverrideExisting`]. This will replace any
    /// previously set values for that header. This behavior can be
    /// changed using the [`mode`] method.
```

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+        }
+    }
+
+    /// Set which mode to use when inserting the header.
+    ///
+    /// Defaults to [`InsertHeaderMode::OverrideExisting`].
+    pub fn mode(mut self, mode: InsertHeaderMode) -> Self {
+        self.mode = mode;
+        self
+    }
```
hawkw, line 146: here's a thought, let me know what you think: it seems like with this API, where the layer is always created in the default mode and then a new mode is set, it's possible for users to get surprising behavior by not reading the docs carefully. we *might* want to consider an API design where the user _can't_ easily construct a new `SetResponseHeaderLayer` without being aware of the behavior of the service it produces.

One approach would be to always require the mode to be explicitly provided in `new`, which is kind of wordy and annoying --- you'd have to write out `InsertHeaderMode::OverrideExisting` every time you create a `SetResponseHeaderLayer`. Alternatively, we could replace the single `new` constructor with three constructors that build layers in each mode --- `new_overwrite`, `new_append`, and `new_if_not_present` or something (or, drop the `new_` prefix)? this is a little less wordy than taking the mode as a parameter, but it still makes it impossible to construct a layer without being aware of the behavior you're going to get.

let me know if you think this is a good idea, or just an additional speed bump that makes the API more complicated? i could be convinced either way.

hawkw: the other advantage of having three constructors is that we can make the enum private, reducing the number of types in the API surface...which is not the _biggest_ deal...

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+/// The mode to use when inserting a header into a request or response.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum InsertHeaderMode {
+    /// Insert the header, overriding any previous values the header might have.
+    OverrideExisting,
```
hawkw, line 154: nit/TIOLI: i think we could make this a little more concise by dropping the "existing" from the name --- i think that it's kind of implied, and "Override" (or "Overwrite"?) would be just as clear on its own. up to you though!

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,457 @@
...
+    ///
+    /// Defaults to [`InsertHeaderMode::OverrideExisting`].
+    pub fn mode(mut self, mode: InsertHeaderMode) -> Self {
+        self.mode = mode;
+        self
+    }
+}
+
+/// The mode to use when inserting a header into a request or response.
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
hawkw: it could _probably_ also be `Copy` --- I don't imagine us making any future changes that would make it not able to be `Copy`...

tower-http/src/set_response_header.rs
hawkw, line 154: nit, TIOLI: since `OverrideExisting` is the default mode, it _might_ also be nice to add a `Default` impl for this that returns it? not terribly important at all, just a thought.

hawkw (commented):
Looking good! I had some docs suggestions, as well as a few more API design thoughts I'd love to hear what you think about. None of them are hard blockers though.

tower-http/src/set_response_header.rs
davidpdrsn: Yeah I think thats cool. Added a Copy impl as well.

davidpdrsn: Good point 👍 

I'm honestly not sure what the difference is between override and overwrite. My impression is that languages most commonly uses "override". According to `define` on Google

- Overwrite: write on top of (other writing).
- Override: use one's authority to reject or cancel (a decision, view, etc.).

I guess both fit here but override seems more specific 🤷

tower-http/src/set_response_header.rs
davidpdrsn: I think its a good thing to make users more conscious about. Could be a source of bugs otherwise. And I don't think it makes the API more complicated.

I agree including the mode in `new` is a bit much. I think perhaps skipping the `new_` prefix could work. So we would have:

- `SetResponseHeader::overriding` as `override` is a keyword in Rust.
- `SetResponseHeader::appending` to match the tense of `overriding`
- `SetResponseHeader::if_not_present`. A shame this one doesn't match, but `inserting_if_not_present` is perhaps a bit much 😅 Could also be `unless_present` but I'd be fine with `if_not`. I've done a lot of ruby with its unless-statements.

hawkw: hmm, this seems reasonable...it would be nice if the tense for all three matched but i can't think of a good naming scheme where they all line up :( so this is good in my book!

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,454 @@
...
+//! # });
+//! #
+//! let mut svc = ServiceBuilder::new()
+//!     .layer(
+//!         // Layer that sets `Content-Type: text/html` on responses.
+//!         //
+//!         // We have to add `::<_, Body>` since Rust cannot infer the body type when
+//!         // we don't use a closure to produce the header value.
+//!         //
+//!         // `if_not_present` will only insert the header if it is not present already.
```
hawkw, line 30: nit: maybe
```suggestion
//!         // `if_not_present` will only insert the header if does not already have a value.
```
or something? just to make sure it's clear that this won't insert the header if _any_ value for that header is already present, not the specific value.

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,454 @@
...
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! # let render_html = tower::service_fn(|request: Request<Body>| async move {
+//! #     Ok::<_, std::convert::Infallible>(Response::new(Body::from("1234567890")))
+//! # });
+//! #
+//! let mut svc = ServiceBuilder::new()
+//!     .layer(
+//!         // Layer that sets `Content-Length` if the body has a known size.
+//!         // Bodies with streaming responses wont have a known size.
+//!         SetResponseHeaderLayer::overriding(
```
hawkw, line 67: could be nice if this comment explained the behavior of `overriding`?

tower-http/src/set_response_header.rs
```diff
@@ -0,0 +1,454 @@
...
+            make,
+            header_name,
+            mode,
+            _marker: PhantomData,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[non_exhaustive]
```
hawkw, line 162: nit/TIOLI: since this is no longer public we can just remove the non_exhaustive, which is only necessary when the enum is part of the public API stability guarantee.
```suggestion
```

hawkw (approved):
looks good to me — I had a few small suggestions, but this looks right!

PR #40, opened 2021-01-26
Add `Trace` middleware

davidpdrsn:
A bit more to this one 😅

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -21,7 +21,7 @@ tower-layer = "0.3"
 tower-service = "0.3"
 
 # optional dependencies
-# will be filled out in other PRs
+tracing = { version = "0.1", optional = true }
 
 [dev-dependencies]
 futures = "0.3"
@@ -31,7 +31,9 @@ tower = "0.4"
 [features]
 default = []
 full = [
+    "trace",
 ]
+trace = ["tracing"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -21,7 +21,7 @@ tower-layer = "0.3"
 tower-service = "0.3"
 
 # optional dependencies
-tracing = { version = "0.1", optional = true }
+tracing = { version = "0.1", optional = true, default-features = false }
 
 [dev-dependencies]
 futures = "0.3"
--- a/tower-http/src/classify_response.rs
+++ b/tower-http/src/classify_response.rs
@@ -1,0 +1,82 @@
+use http::{Response, StatusCode};
+use std::fmt;
+
+/// Classify requests results as either success or failure.
+///
+/// Not all services can rely entirely on the status code to determine if its a success or failure.
+/// gRPC for example has its own way of communicating that and always uses `200 OK`.
+pub trait ClassifyResponse<Body, Err> {
+    /// The type used to classify successful responses.
+    type OkClass;
+
+    /// The type used to classify failed responses.
+    type ErrClass;
+
+    /// Do the classification.
+    fn classify_request_result(
+        &self,
+        request_result: &Result<Response<Body>, Err>,
+    ) -> ResponseClassification<Self::OkClass, Self::ErrClass>;
+}
+
+/// Was a response a success or a failure?
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ResponseClassification<T, E> {
+    /// The response was a success.
+    ///
+    /// `T` is a value produced by doing the classification. Could for example be the status code
+    /// that a middleware uses.
+    Ok(T),
+
+    /// The response was a failure.
+    ///
+    /// `E` is a value produced by doing the classification. Could for example be an error.
+    Err(E),
+}
+
+/// The default [`ClassifyResponse`] that some middlewares use.
+///
+/// It classifies `Ok(response)` as success unless the status code is `5xx`. Errors are always
+/// classified as errors.
+#[derive(Debug, Clone, Default)]
+pub struct DefaultHttpResponseClassifier {
+    _priv: (),
+}
+
+impl DefaultHttpResponseClassifier {
+    /// Create a new [`DefaultHttpResponseClassifier`].
+    pub fn new() -> Self {
+        Self { _priv: () }
+    }
+}
+
+impl<B, E> ClassifyResponse<B, E> for DefaultHttpResponseClassifier
+where
+    E: fmt::Display,
+{
+    type OkClass = StatusCode;
+    type ErrClass = DefaultErrorClassification;
+
+    fn classify_request_result(
+        &self,
+        request_result: &Result<Response<B>, E>,
+    ) -> ResponseClassification<StatusCode, DefaultErrorClassification> {
+        match request_result {
+            Ok(res) if res.status().is_server_error() => {
+                let err = format!("Server error. status={}", res.status());
+                let err = DefaultErrorClassification { err };
+                ResponseClassification::Err(err)
+            }
+            Ok(res) => ResponseClassification::Ok(res.status()),
+            Err(err) => ResponseClassification::Err(DefaultErrorClassification {
+                err: err.to_string(),
+            }),
+        }
+    }
+}
+
+/// The error classification used by [`DefaultHttpResponseClassifier`].
+#[derive(Debug, Clone)]
+pub struct DefaultErrorClassification {
+    pub(crate) err: String,
+}
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -58,12 +58,14 @@ pub trait GetTraceStatus<T, E> {
 }
 
 // TODO(david): can we come up with a better name for this?
+#[derive(Debug)]
+#[non_exhaustive]
 pub enum TraceStatus {
     Status(u16),
     Error,
 }
 
-#[derive(Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub struct GetTraceStatusFromHttpStatus(());
 
 impl<B, E> GetTraceStatus<Response<B>, E> for GetTraceStatusFromHttpStatus {
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -38,51 +38,16 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(test, allow(clippy::float_cmp))]
 
-use http::Response;
-
 #[cfg(feature = "trace")]
 #[cfg_attr(docsrs, doc(cfg(feature = "trace")))]
 pub mod trace;
 
+pub mod classify_response;
+
+/// The unit some middlewares will use when dealing with latencies.
 #[derive(Debug, Clone, Copy)]
 #[non_exhaustive]
 pub enum LatencyUnit {
     Millis,
     Nanos,
-}
-
-/// You might not always want to log the literal HTTP status. gRPC for example has its own status
-/// and always uses `200 OK` even for errors.
-// TODO(david): can we come up with a better name for this?
-pub trait GetTraceStatus<T, E> {
-    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus;
-}
-
-// TODO(david): can we come up with a better name for this?
-#[derive(Debug)]
-#[non_exhaustive]
-pub enum TraceStatus {
-    Status(u16),
-    Error,
-}
-
-#[derive(Debug, Copy, Clone)]
-pub struct GetTraceStatusFromHttpStatus(());
-
-impl<B, E> GetTraceStatus<Response<B>, E> for GetTraceStatusFromHttpStatus {
-    fn trace_status(&self, result: &Result<Response<B>, E>) -> TraceStatus {
-        match result {
-            Ok(res) => TraceStatus::Status(res.status().as_u16()),
-            Err(_) => TraceStatus::Error,
-        }
-    }
-}
-
-impl<T, E, F> GetTraceStatus<T, E> for F
-where
-    F: Fn(&Result<T, E>) -> TraceStatus,
-{
-    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus {
-        self(result)
-    }
 }
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -64,7 +64,7 @@ pub enum TraceStatus {
 }
 
 #[derive(Copy, Clone)]
-pub struct GetTraceStatusFromHttpStatus;
+pub struct GetTraceStatusFromHttpStatus(());
 
 impl<B, E> GetTraceStatus<Response<B>, E> for GetTraceStatusFromHttpStatus {
     fn trace_status(&self, result: &Result<Response<B>, E>) -> TraceStatus {
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -37,3 +37,49 @@
 #![forbid(unsafe_code)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(test, allow(clippy::float_cmp))]
+
+use http::Response;
+
+#[cfg(feature = "trace")]
+#[cfg_attr(docsrs, doc(cfg(feature = "trace")))]
+pub mod trace;
+
+#[derive(Debug, Clone, Copy)]
+pub enum LatencyUnit {
+    Millis,
+    Nanos,
+}
+
+/// You might not always want to log the literal HTTP status. gRPC for example has its own status
+/// and always uses `200 OK` even for errors.
+// TODO(david): can we come up with a better name for this?
+pub trait GetTraceStatus<T, E> {
+    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus;
+}
+
+// TODO(david): can we come up with a better name for this?
+pub enum TraceStatus {
+    Status(u16),
+    Error,
+}
+
+#[derive(Copy, Clone)]
+pub struct GetTraceStatusFromHttpStatus;
+
+impl<B, E> GetTraceStatus<Response<B>, E> for GetTraceStatusFromHttpStatus {
+    fn trace_status(&self, result: &Result<Response<B>, E>) -> TraceStatus {
+        match result {
+            Ok(res) => TraceStatus::Status(res.status().as_u16()),
+            Err(_) => TraceStatus::Error,
+        }
+    }
+}
+
+impl<T, E, F> GetTraceStatus<T, E> for F
+where
+    F: Fn(&Result<T, E>) -> TraceStatus,
+{
+    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus {
+        self(result)
+    }
+}
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -45,6 +45,7 @@ use http::Response;
 pub mod trace;
 
 #[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
 pub enum LatencyUnit {
     Millis,
     Nanos,
--- a/tower-http/src/trace.rs
+++ b/tower-http/src/trace.rs
@@ -11,52 +11,67 @@ use std::{
 };
 use tower_layer::Layer;
 use tower_service::Service;
-use tracing::{
-    field::{debug, display},
-    Level, Span,
-};
+use tracing::{field::debug, Level, Span};
 
 #[derive(Clone, Debug)]
-pub struct TraceLayer<T = GetTraceStatusFromHttpStatus> {
+pub struct TraceLayer<MakeSpan, T = GetTraceStatusFromHttpStatus> {
     record_headers: bool,
-    span: Option<Span>,
     latency_unit: LatencyUnit,
     get_trace_status: T,
     record_full_uri: bool,
+    make_span: MakeSpan,
 }
 
-impl Default for TraceLayer<GetTraceStatusFromHttpStatus> {
+impl<B> Default for TraceLayer<fn(&Request<B>) -> Span, GetTraceStatusFromHttpStatus> {
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl TraceLayer<GetTraceStatusFromHttpStatus> {
+impl<B> TraceLayer<fn(&Request<B>) -> Span, GetTraceStatusFromHttpStatus> {
     pub fn new() -> Self {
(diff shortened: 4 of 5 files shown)


tower-http/src/lib.rs
```diff
@@ -37,3 +37,49 @@
...
+}
+
+// TODO(david): can we come up with a better name for this?
+pub enum TraceStatus {
+    Status(u16),
+    Error,
+}
+
+#[derive(Copy, Clone)]
+pub struct GetTraceStatusFromHttpStatus(());
```
LucioFranco, line 67: debug?

tower-http/src/lib.rs
```diff
@@ -37,3 +37,49 @@
...
+
+/// You might not always want to log the literal HTTP status. gRPC for example has its own status
+/// and always uses `200 OK` even for errors.
+// TODO(david): can we come up with a better name for this?
+pub trait GetTraceStatus<T, E> {
+    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus;
+}
+
+// TODO(david): can we come up with a better name for this?
+pub enum TraceStatus {
```
LucioFranco, line 61: non_exhaustive?

LucioFranco (commented):
Missing some doc comments :D

tower-http/src/lib.rs
davidpdrsn: 👍

tower-http/src/lib.rs
davidpdrsn: 👍

davidpdrsn:
> Missing some doc comments

My plan was to add docs in follow up PRs when we've agreed in the APIs.

LucioFranco:
Works for me, since this deals with tracing I'd like @hawkw or @davidbarsky to review it.

tower-http/src/lib.rs
```diff
@@ -37,3 +37,51 @@
...
+    Millis,
+    Nanos,
+}
+
+/// You might not always want to log the literal HTTP status. gRPC for example has its own status
+/// and always uses `200 OK` even for errors.
+// TODO(david): can we come up with a better name for this?
+pub trait GetTraceStatus<T, E> {
+    fn trace_status(&self, result: &Result<T, E>) -> TraceStatus;
+}
```
hawkw, line 58: This seems like it might be worth generalizing into a more generic notion of ["response classification"], which might be useful for other purposes as well (e.g. retries). In Linkerd, we have [our own version of this](https://github.com/linkerd/linkerd2-proxy/blob/e0271b4ee4841ef86fcbdb34f86b0fe725c502e2/linkerd/http-classify/src/lib.rs), it might be worthwhile to try to design a general-purpose response classification trait for `tower-http`?

tower-http/src/lib.rs
```diff
@@ -37,3 +37,51 @@
...
 #![cfg_attr(test, allow(clippy::float_cmp))]
+
+use http::Response;
+
+#[cfg(feature = "trace")]
+#[cfg_attr(docsrs, doc(cfg(feature = "trace")))]
+pub mod trace;
+
+#[derive(Debug, Clone, Copy)]
+pub enum LatencyUnit {
```
hawkw, line 48: this should definitely be non-exhaustive

tower-http/src/lib.rs
```diff
@@ -37,3 +37,51 @@
...
+    Millis,
+    Nanos,
+}
+
+/// You might not always want to log the literal HTTP status. gRPC for example has its own status
+/// and always uses `200 OK` even for errors.
+// TODO(david): can we come up with a better name for this?
+pub trait GetTraceStatus<T, E> {
```
hawkw, line 56: nit/TIOLI: IMO, idiomatic rust tends not to use `get` except for collections that can look up arbitrary keys. I might drop the `Get` (but see my other comment about maybe replacing this with a more generic response-classifier design?)

tower-http/src/trace.rs
```diff
@@ -0,0 +1,230 @@
...
+#[derive(Clone, Debug)]
+pub struct TraceLayer<T> {
+    record_headers: bool,
+    span: Option<Span>,
+    latency_unit: LatencyUnit,
+    get_trace_status: T,
+    record_full_uri: bool,
+}
+
+impl Default for TraceLayer<GetTraceStatusFromHttpStatus> {
```
hawkw, line 28: should this maybe be a default parameter on `TraceLayer` as well?

tower-http/src/trace.rs
```diff
@@ -0,0 +1,230 @@
...
+        self.latency_unit = latency_unit;
+        self
+    }
+
+    pub fn record_full_uri(mut self, record_full_uri: bool) -> Self {
+        self.record_full_uri = record_full_uri;
+        self
+    }
+
+    pub fn get_trace_status<K>(self, get_trace_status: K) -> TraceLayer<K> {
```
hawkw, line 72: same naming quibble as above. also, why `K`?

tower-http/src/trace.rs
```diff
@@ -0,0 +1,230 @@
...
+        self.record_headers = record_headers;
+        self
+    }
+
+    /// Provide a custom span. The span is expected to at least have the fields `method`, `path`,
+    /// and `headers`. If any of the fields are missing from the span they'll also be missing from
+    /// whatever output you may have configured.
+    ///
+    /// The default span uses `DEBUG` level and is called `http-request`.
+    pub fn span(mut self, span: Span) -> Self {
```
hawkw, line 57: instead of taking a specific `Span` and cloning it, what do you think about taking a `Fn(&http::Request<B>) -> Span`? that way, the user can customize the span however they want, but get a new span created for _each_ request, which is probably what we'd want (and what this code does by default). also, we could avoid the use of `span.record`, which is a little less efficient than recording all the fields at span creation time.

if we did this, the default behavior could even be a function pointer, so we wouldn't need to special-case it.

tower-http/src/trace.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub fn record_headers(mut self, record_headers: bool) -> Self {
+        self.record_headers = record_headers;
+        self
+    }
+
+    /// Provide a custom span. The span is expected to at least have the fields `method`, `path`,
+    /// and `headers`. If any of the fields are missing from the span they'll also be missing from
+    /// whatever output you may have configured.
+    ///
+    /// The default span uses `DEBUG` level and is called `http-request`.
```
hawkw, line 56: looks like this changed to `INFO`?

tower-http/src/trace.rs
```diff
@@ -0,0 +1,230 @@
...
+                tracing::info!(
+                    message = "request failed",
+                    latency_ns = %time.as_nanos(),
+                );
+            }
+            (TraceStatus::Error, LatencyUnit::Millis) => {
+                tracing::info!(
+                    message = "request failed",
+                    latency_ms = %time.as_millis(),
+                );
```
hawkw, line 224: this feels a bit opinionated about what the output looks like...do we want to consider providing some way of allowing the user to customize these events further?

hawkw (commented):
overall, this feels like a good start. i had some thoughts, let me know what you think?

tower-http/src/lib.rs
davidpdrsn: Good point. I'll keep this in mind when I think about [your other comment](https://github.com/tower-rs/tower-http/pull/40#discussion_r566474374).

tower-http/src/trace.rs
davidpdrsn: No particular reason. Its called `T` on the `impl` and `K` comes after `T` 😅

tower-http/src/trace.rs
davidpdrsn: Good idea!

tower-http/src/trace.rs
davidpdrsn: Yeah true. I guess you're probably the one to ask about this. I struggled with making things customizable since the name of span and the fields (afaik) must be declared up front.

I suppose we could do what we do wrt "response classification", have a trait for customizing it and provide a default impl. Off of the top of my head that trait would have a method like `fn emit_event(&self, trace_status: TraceStatus, latency_unit: LatencyUnit)`. `TraceStatus` would most likely be renamed though but thats separate.

What do you think about this?

tower-http/src/trace.rs
davidpdrsn: Good idea 👍

davidpdrsn:
@hawkw I've added some new response classification stuff and generally made the middleware more configurable. Please have another look.

tower-http/src/classify_response.rs
```diff
@@ -0,0 +1,82 @@
...
+    type OkClass = StatusCode;
+    type ErrClass = DefaultErrorClassification;
+
+    fn classify_request_result(
+        &self,
+        request_result: &Result<Response<B>, E>,
+    ) -> ResponseClassification<StatusCode, DefaultErrorClassification> {
+        match request_result {
+            Ok(res) if res.status().is_server_error() => {
+                let err = format!("Server error. status={}", res.status());
```
hawkw, line 66: this will allocate a string every time a server error is returned. that feels like a _lot_ of overhead that I don't think is really necessary. what do you think about changing `DefaultErrorClassification` to something more like this:
```rust
pub struct DefaultErrorClassification<E> {
    inner: DefaultErrorKind<E>,
}

enum DefaultErrorKind<E> {
    Err(E),
    Status(http::Status),
}

impl<E: fmt::Display> fmt::Display for DefaultErrorClassification<E> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self.inner {
            DefaultErrorKind::Status(s) => write!(f, "server error. status={}", s),
            DefaultErrorKind::Err(e) => write!(f, "{}", e),
        }
    }
}
```
or something like that. This way, we don't allocate strings at all, and only format the errors/status codes when needed. Also, you could easily add an `Error` impl for `DefaultErrorClassification` that allows downcasting to the original `Error`.

tower-http/src/classify_response.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    /// The type used to classify failed responses.
+    type ErrClass;
+
+    /// Do the classification.
+    fn classify_request_result(
+        &self,
+        request_result: &Result<Response<Body>, Err>,
+    ) -> ResponseClassification<Self::OkClass, Self::ErrClass>;
+}
```
hawkw, line 20: I'm not sure if this API for response classification is really sufficient. 

For example, you mentioned gRPC in the docs. However, gRPC responses should be classified based on [the `grpc-status` trailer](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md), which is only sent _after_ a streaming body completes. Therefore, we can't classify gRPC responses based on the `http::Response` type --- we would need to wrap the body, returning a new response, and run the classification logic only when the body type's [`poll_trailers` method](https://docs.rs/http-body/0.4.0/http_body/trait.Body.html#tymethod.poll_trailers) returns trailers.

This is why linkerd's response-classification code has [two separate traits, `Classify` and `ClassifyEos`](https://github.com/linkerd/linkerd2-proxy/blob/e0271b4ee4841ef86fcbdb34f86b0fe725c502e2/linkerd/http-classify/src/lib.rs#L27-L53). A `Classify` can classify errors immediately, but in order to classify a response, it needs to produce an end-of-stream classifier that waits until the response's trailers are received. A design more like this is probably necessary in order to properly support gRPC...

tower-http/src/classify_response.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Classify requests results as either success or failure.
+///
+/// Not all services can rely entirely on the status code to determine if its a success or failure.
+/// gRPC for example has its own way of communicating that and always uses `200 OK`.
+pub trait ClassifyResponse<Body, Err> {
+    /// The type used to classify successful responses.
+    type OkClass;
+
+    /// The type used to classify failed responses.
+    type ErrClass;
```
hawkw, line 13: nit: I might change this naming to `SuccessClass` and `FailureClass`, or even just `Success` and `Failure` — using `Ok` and `Err` feels like it kind of conflates these with whether the `Result` passed into the classifier is `Ok` or `Err`, but an `Ok` result might still be a _failure_ at the HTTP level, if it contains a 5xx status code.

hawkw: also, do we really need a class type for successful responses? if a response succeeded, do we expect users to want to classify that success into different kinds of success, or is it just "not an error"? IMO, it would _probably_ be fine for the success side of this to be `()`...but maybe leaving it as a type param is more future-proof? i could be convinced either way...

tower-http/src/trace.rs
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+}
+
+/// Middleware that add high level [tracing] to a [`Service`].
+///
+/// [tracing]: https://crates.io/crates/tracing
+/// [`Service`]: tower_service::Service
+#[derive(Clone, Debug)]
+pub struct Trace<S, Classifier, EventEmitter> {
```
hawkw, line 148: nit: since the classifier and event emitter have defaults on the layer type, they probably should here as well?

tower-http/src/trace.rs
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+}
+
+/// The [`Future`] produced by [`Trace`] services.
+///
+/// [`Future`]: std::future::Future
+#[pin_project]
+#[derive(Debug)]
+pub struct ResponseFuture<F, Classifier, EventEmitter, ReqBody> {
```
hawkw, line 200: nit: since the classifier and event emitter parameters have defaults on the `Layer` type, they probably should here as well? this would mean moving the `ReqBody` param before the `Classifier` and `EventEmitter` params, but I think that's fine...

tower-http/src/trace.rs
```diff
@@ -0,0 +1,348 @@
...
+    /// This mirrors [`ClassifyResponse::ErrClass`].
+    type ErrClass;
+
+    /// Make a new [`tracing::Span`] from a request.
+    ///
+    /// [`tracing::Span`]: https://docs.rs/tracing/latest/tracing/struct.Span.html
+    fn make_span(&self, req: &Request<ReqBody>) -> Span;
+
+    /// Emit an event when a new request has been received.
+    fn on_request_received(&self, req: &Request<ReqBody>);
```
hawkw, line 257: This doesn't actually care about the request body type, does it? I think we don't _need_ the trait to be generic over body types, and could just make the individual methods generic. That might simplify the usage of the trait a bit:
```suggestion
pub trait EmitTracingEvents {
    /// The type used to classify successful responses.
    ///
    /// This mirrors [`ClassifyResponse::OkClass`].
    type OkClass;

    /// The type used to classify failed responses.
    ///
    /// This mirrors [`ClassifyResponse::ErrClass`].
    type ErrClass;

    /// Make a new [`tracing::Span`] from a request.
    ///
    /// [`tracing::Span`]: https://docs.rs/tracing/latest/tracing/struct.Span.html
    fn make_span<ReqBody>(&self, req: &Request<ReqBody>) -> Span;

    /// Emit an event when a new request has been received.
    fn on_request_received<ReqBody>(&self, req: &Request<ReqBody>);
```

tower-http/src/trace.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    /// Emit an event when a request has been processed and the result has been classified.
+    fn on_classified_result(
+        &self,
+        result: ResponseClassification<Self::OkClass, Self::ErrClass>,
+        latency: Duration,
+        latency_unit: LatencyUnit,
+    );
```
hawkw, line 265: not totally sure how i feel about putting all these callbacks on one trait. it seems to me like it might be better to take a separate `Fn` for the span, request start, and response classification callbacks. that way, a user could override one of them while using the defaults for the others?

tower-http/src/trace.rs
```diff
@@ -0,0 +1,348 @@
...
+{
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+
+        let _guard = this.span.enter();
+
+        let result = ready!(this.future.poll(cx));
+        let latency = this.start.elapsed();
```
hawkw, line 227: it seems like this always records latency in terms of time until the beginning of the response. i can also see use-cases that might want to record latency in terms of time to first byte received, or time to response body EOS. we might want to think about allowing the user to customize this — but, it's probably out of scope for this initial PR; just something to think about later.

hawkw (commented):
Okay, I had some more feedback, mostly related to the response classifier trait. I don't know if the current approach will work for use-cases like gRPC.

I don't think that we need to use _exactly_ [the same response classifier design we use in Linkerd](https://github.com/linkerd/linkerd2-proxy/blob/e0271b4ee4841ef86fcbdb34f86b0fe725c502e2/linkerd/http-classify/src/lib.rs) but i _would_, eventually, like to be able to replace our implementation with a version from `tower-http`, and i don't think the current impl will work for our use-case. 

If you like, I'd be happy to open a PR to add _just_ a response-classifier trait, and we could rebase this branch onto that once it merges?  Or, we could punt on getting the design exactly right in this branch and continue working on response classification after this merges. Let me know what works best for you?

tower-http/Cargo.toml
```diff
@@ -21,7 +21,7 @@ tower-layer = "0.3"
 tower-service = "0.3"
 
 # optional dependencies
-# will be filled out in other PRs
+tracing = { version = "0.1", optional = true }
```
hawkw, line 24: nit: we should probably set 
```suggestion
tracing = { version = "0.1", optional = true, default-features = false }
```
so that users who are depending on tracing with its default features disabled can still use this without having them all be turned on — we don't use anything that's feature-flagged in `tracing`.

tower-http/src/classify_response.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Do the classification.
+    fn classify_request_result(
+        &self,
+        request_result: &Result<Response<Body>, Err>,
+    ) -> ResponseClassification<Self::OkClass, Self::ErrClass>;
+}
+
+/// Was a response a success or a failure?
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ResponseClassification<T, E> {
```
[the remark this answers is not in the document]
davidpdrsn: My thought was that using `Result` makes it seem like the classification itself failed when actually it was processing the request that failed.

tower-http/src/classify_response.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// `E` is a value produced by doing the classification. Could for example be an error.
+    Err(E),
+}
+
+/// The default [`ClassifyResponse`] that some middlewares use.
+///
+/// It classifies `Ok(response)` as success unless the status code is `5xx`. Errors are always
+/// classified as errors.
+#[derive(Debug, Clone, Default)]
+pub struct DefaultHttpResponseClassifier {
```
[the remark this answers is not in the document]
davidpdrsn: I'm generally also a fan of naming things in terms of what they do. `ServerErrorsAsFailures` sounds good 😊

tower-http/src/classify_response.rs
davidpdrsn: Yeah I agree having to allocate a string sucks. However since the classify method receives `&Result<Response<_>, E>` we cannot get an owned `E` to put into the return value. That introduces a lifetime which spread everywhere. Not sure if there is an elegant solution to that.

tower-http/src/classify_response.rs
davidpdrsn: Uhh thats intersting. I didn't know gRPC could do that. Makes sense why this doesn't work then.

Seeing as you have lots of experience with this area would you perhaps want to make an initial version that you think would work? Could be as a PR to this PR or a separate new PR.

I'm not against copying what you have linkerd-proxy but my initial impression was that it was pretty complicated. Perhaps some docs and usage examples would help. Seems like the complexities are warranted though 😜

tower-http/src/classify_response.rs
davidpdrsn: `SuccessClass` and `FailureClass` sound better 👍

Regarding yielding a value from the classification that was a way to get a status code to the thing that emits the events.

tower-http/src/trace.rs
davidpdrsn: That was such that if users wanted to do something interesting with the body they could implement this trait for whatever body they have. Could be someone `POST`ing JSON that you want included in your logs. Not sure you can do that when `ReqBody` is declared on the function.

However having the type on the trait does lead to the generic param spreading to a bunch of places which is kinda annoying.

tower-http/src/trace.rs
davidpdrsn: Yeah I did consider that as well. However then we end up with like 4-5 type params on `Trace` which is quite a lot. Adding default implementations in the trait itself is hard because then we cannot pass configuration elsewhere than as direct arguments.

tower-http/src/trace.rs
davidpdrsn: Uhh yeah thats a good point. Didn't consider that.

davidpdrsn:
@hawkw Thanks for providing such detailed and thoughtful feedback! I think its important that we aspire to build something that is useable both for simple toy web apps and complex setups like linkerd-proxy.

tower-http/src/classify_response.rs
hawkw: > I'm not against copying what you have linkerd-proxy but my initial impression was that it was pretty complicated.

It is pretty complicated, but I think a lot of that complexity is kind of inherently necessary if you want to be able to include trailers in classification, which is kind of necessary to support gRPC. 

While we can't get away with not having it, I think it _might_ be possible to design an interface for `tower-http` where that additional complexity is _optional_...we might be able to have an interface where you can write a simple classifier based only on HTTP status codes without having to write two trait impls?

tower-http/src/trace.rs
hawkw: it *is* just one generic parameter so it isn't a huge amount of complexity. the future proofing might be worth it.

tower-http/src/classify_response.rs
davidpdrsn: I'm not really worried about things being complex if thats what it takes to also be correct. I think its about making "easy things easy and hard things possible". So ideally users who just want the defaults should be able to do `.layer(TraceLayer::new())` and move on. Whereas if deep customization requires implementing two traits then I don't think thats necessarily bad.

davidpdrsn: Also, we use gRPC at Embark so I have some vested interested the gRPC classification working 😅

davidpdrsn:
Superseded by https://github.com/tower-rs/tower-http/pull/49

PR #42, opened 2021-01-28
Bring back `HttpService` trait alias

tesaguri:
This brings back the `HttpService` trait from the old `tower-http-util`.

With `HttpService`, you can rewrite trait bounds like this:

```rust
fn foo<S, ReqBody, ResBody>(_: S)
where
    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
    ResBody: Body,
```

to this:

```rust
fn foo<S, B>(_: S)
where
    S: HttpService<B>,
```

The code is mostly copy and pasted from 926a64f262919a559624b6b99c2f38498ec3ba38.

### DIFF
--- a/tower-http/src/service.rs
+++ b/tower-http/src/service.rs
@@ -1,0 +1,86 @@
+//! Types and utilities for working with `HttpService`
+
+mod as_service;
+mod into_service;
+
+pub use self::as_service::AsService;
+pub use self::into_service::IntoService;
+
+use http::{Request, Response};
+use http_body::Body;
+use std::future::Future;
+use std::task::{Context, Poll};
+use tower_service::Service;
+
+/// An HTTP service
+///
+/// This is not intended to be implemented directly. Instead, it is a trait
+/// alias of sorts. Implements the [`tower_service::Service`] trait using
+/// [`http::Request`] and [`http::Response`] types.
+pub trait HttpService<RequestBody>: sealed::Sealed<RequestBody> {
+    /// Response payload.
+    type ResponseBody: Body;
+
+    /// Errors produced by the service.
+    type Error;
+
+    /// The future response value.
+    type Future: Future<Output = Result<Response<Self::ResponseBody>, Self::Error>>;
+
+    /// Returns `Ready` when the service is able to process requests.
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+
+    /// Process the request and return the response asynchronously.
+    fn call(&mut self, request: Request<RequestBody>) -> Self::Future;
+
+    /// Wrap the `HttpService` so that it implements `tower_service::Service`
+    /// directly.
+    ///
+    /// Since `HttpService` does not directly implement `Service`, if an
+    /// `HttpService` instance needs to be used where a `T: Service` is
+    /// required, it must be wrapped with a type that provides that
+    /// implementation. `IntoService` does this.
+    fn into_service(self) -> IntoService<Self>
+    where
+        Self: Sized,
+    {
+        IntoService::new(self)
+    }
+
+    /// Same as `into_service` but operates on an `HttpService` reference.
+    fn as_service(&mut self) -> AsService<'_, Self>
+    where
+        Self: Sized,
+    {
+        AsService::new(self)
+    }
+}
+
+impl<T, ReqBody, ResBody> HttpService<ReqBody> for T
+where
+    T: Service<Request<ReqBody>, Response = Response<ResBody>>,
+    ResBody: Body,
+{
+    type ResponseBody = ResBody;
+    type Error = T::Error;
+    type Future = T::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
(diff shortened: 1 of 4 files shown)


tower-http/src/service.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// implementation. `IntoService` does this.
+    fn into_service(self) -> IntoService<Self>
+    where
+        Self: Sized,
+    {
+        IntoService::new(self)
+    }
+
+    /// Same as `into_service` but operates on an `HttpService` reference.
+    fn as_service(&mut self) -> AsService<'_, Self>
```
tesaguri, line 51: This could take `?Sized` if the method were to be separated to another trait like `HttpServiceExt`, but I doubt if `dyn HttpService` would be useful in practice.

davidpdrsn (changes requested):
I'm struggling to understand when this trait would be useful for us as maintainers and users as well. But I'm up for discussing it.

I have a couple of questions:

Am I right that the only thing this trait does is simplify the trait bounds on functions that takes a specific kind of service, as you show?

Would we start implementing this trait for middlewares rather than `tower_service::Service`? I'm guessing that would be annoying because the trait is sealed.

Would this work nicely with the rest of the tower ecosystem or would users have to call `.into_service()` all the time?

Unless this trait unlocks some good use cases I kinda feel it isn't worth the additional overhead in terms of understand tower-http for users. I feel the same way about the `MakeService` and `MakeConnection` traits that in tower today but the ship has already sailed on those 😜

tesaguri:
@davidpdrsn

> Am I right that the only thing this trait does is simplify the trait bounds on functions that takes a specific kind of service

Yes, the main purpose of the trait is for brevity of trait bounds of functions (and, to be exact, type declarations and inherent/trait implementations). But, as type parameters are part of public API, it actually affects ergonomics of users of the downstream library when used with turbofish syntax (`foo::<S, B1, B2>(..);` vs `foo::<S, B>(..);`). Additionally, the trait may also offer other convenience methods like `map_body`, but they can be defined in another trait like `HttpServiceExt`.

> Would we start implementing this trait for middlewares rather than tower_service::Service?

No, you will implement `Service<Request<B1>, Response = Response<B2>>` as always and get an `HttpService` implementation for free thanks to the blanket implementation `impl<S: Service<..>, B1, B2: Body> HttpService<B1> for S`. However, it would still be great to implement the middlewares for `impl HttpService` rather than implement `HttpService` for the middlewares as you implied, that is, `impl<S: HttpService<B>, B> Service<..> for Middleware<S>` rather than `impl<S: Service<..>, B1, B2> HttpService<B1> for Middleware<S>`, since the former allows users of the middlewares use both `impl Service` and `impl HttpService` without using `.into_service()`.

> would users have to call .into_service() all the time?

I guess you are asking this on the assumption that the answer to the second question is "yes"? Since `HttpService` is an opt-in trait alias, users can opt against using it if they want to avoid `into_service`. This works the same way as the [`TryFuture`] and [`TryStream`] trait aliases.

[`TryFuture`]: https://docs.rs/futures-core/0.3.12/futures_core/future/trait.TryFuture.html
[`TryStream`]: https://docs.rs/futures/0.3.12/futures/stream/trait.TryStream.html

As a prior art, `hyper` has an internal [`HttpService`] trait alias similar to the one in this PR, which is apparently used 33 times in its code base. I also have custom `HttpService` in my code (I have [three][twitter-stream] [of][pipitor] [them][twitter-client]!), which is used 39 times in total. Writing `<S, B1, B2> ... where S: Service<Request<B1>, Response = Response<B2>>, B2: Body,` dozens of times is not the most pleasant experience. On the other hand, `hyper` does not seem to be using something like `into_service` (maybe it doesn't using Tower combinators at all) and my crates use it only 4 times. In general, I guess `.into_service()` will be simpler and shorter than the trait bound without `HttpService` in most cases because, usually, you will need to call `.into_service()` only once in each function at most since results of subsequent combinator calls implement `Service`.

[`HttpService`]: https://github.com/hyperium/hyper/blob/2c8121f1735aa8efeb0d5e4ef595363c373ba470/src/service/http.rs#L8-L27
[twitter-stream]: https://github.com/tesaguri/twitter-stream-rs/blob/40bc93098b72ac628ef5c9f4c9c8e18afb6b13fb/src/service.rs#L14-L17
[pipitor]: https://github.com/tesaguri/pipitor/blob/7ea1304a930129d857a3ef8e6ff756508754b632/src/util/http_service.rs#L17-L31
[twitter-client]: https://nest.pijul.com/tesaguri/twitter-client-rs:main/FRMT7RO7G2QNM.6YDQA

Having said that, users could define their own `HttpService` trait in the same way as `hyper` does already even if `tower-http` does not provide one. However, one universal standard `HttpService` rather than many similar traits in each crate will provide maximum interoperability for the ecosystem (*remembers <https://xkcd.com/927/>* 🙄).

davidpdrsn:
I would like to hear what others think. cc @LucioFranco @olix0r @hawkw

davidpdrsn:
Sorry but I'm going to close this. I asked what people thought in Discord last week and it seemed people were generally on the fence about it. You're welcome to open an issue if you want to discuss it further.

PR #43, opened 2021-01-29
Add some docs and tests

davidpdrsn:
- Initial root module docs.
- Docs and tests for `AddExtension`.

### DIFF
--- a/tower-http/src/add_extension.rs
+++ b/tower-http/src/add_extension.rs
@@ -1,4 +1,4 @@
-//! Middleware for adding some shareable value to [request extensions].
+//! Middleware that clones a value into each request's [extensions].
 //!
 //! [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
 
--- a/tower-http/src/add_extension.rs
+++ b/tower-http/src/add_extension.rs
@@ -1,14 +1,22 @@
+//! Middleware for adding some shareable value to [request extensions].
+//!
+//! [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
+
 use http::Request;
 use std::task::{Context, Poll};
 use tower_layer::Layer;
 use tower_service::Service;
 
+/// [`Layer`] for adding some shareable value to [request extensions].
+///
+/// [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
 #[derive(Clone, Copy, Debug)]
 pub struct AddExtensionLayer<T> {
     value: T,
 }
 
 impl<T> AddExtensionLayer<T> {
+    /// Create a new [`AddExtensionLayer`].
     pub fn new(value: T) -> Self {
         AddExtensionLayer { value }
     }
@@ -28,10 +36,35 @@ where
     }
 }
 
+/// Middleware for adding some shareable value to [request extensions].
+///
+/// [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
 #[derive(Clone, Copy, Debug)]
 pub struct AddExtension<S, T> {
     inner: S,
     value: T,
+}
+
+impl<S, T> AddExtension<S, T> {
+    /// Create a new [`AddExtension`].
+    pub fn new(inner: S, value: T) -> Self {
+        Self { inner, value }
+    }
+
+    /// Gets a reference to the underlying service.
+    pub fn get_ref(&self) -> &S {
+        &self.inner
+    }
+
+    /// Gets a mutable reference to the underlying service.
+    pub fn get_mut(&mut self) -> &mut S {
+        &mut self.inner
+    }
+
+    /// Gets a mutable reference to the underlying service.
+    pub fn into_inner(self) -> S {
+        self.inner
+    }
 }
 
 impl<ResBody, S, T> Service<Request<ResBody>> for AddExtension<S, T>
@@ -51,5 +84,37 @@ where
     fn call(&mut self, mut req: Request<ResBody>) -> Self::Future {
         req.extensions_mut().insert(self.value.clone());
         self.inner.call(req)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[allow(unused_imports)]
+    use super::*;
+    use http::Response;
+    use hyper::Body;
+    use std::{convert::Infallible, sync::Arc};
+    use tower::{service_fn, ServiceBuilder, ServiceExt};
+
+    struct State(i32);
+
+    #[tokio::test]
+    async fn basic() {
+        let state = Arc::new(State(1));
+
+        let svc = ServiceBuilder::new()
+            .layer(AddExtensionLayer::new(state))
+            .service(service_fn(|req: Request<Body>| async move {
+                let state = req.extensions().get::<Arc<State>>().unwrap();
+                Ok::<_, Infallible>(Response::new(state.0))
+            }));
+
+        let res = svc
+            .oneshot(Request::new(Body::empty()))
+            .await
+            .unwrap()
+            .into_body();
+
+        assert_eq!(1, res);
     }
 }
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -63,7 +63,7 @@
 //! }
 //! ```
 //!
-//! # Feature toggles
+//! # Feature Flags
 //!
 //! All middleware are disabled by default and can be enabled using [cargo features].
 //!
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -65,7 +65,7 @@
 //!
 //! # Feature toggles
 //!
-//! All middlewares are disabled by default and can be enabled using [cargo features].
+//! All middleware are disabled by default and can be enabled using [cargo features].
 //!
 //! For example, to enable the [`AddExtension`] middleware, add the "add-extension" feature flag
 //! in your`Cargo.toml`:
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -1,3 +1,5 @@
+//! TODO(david): root module docs
+
 #![doc(html_root_url = "https://docs.rs/tower-http/0.1.0")]
 #![allow(elided_lifetimes_in_paths, clippy::type_complexity)]
 #![warn(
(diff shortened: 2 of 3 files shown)


tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
+//! `async fn(HttpRequest) -> Result<HttpResponse, Error>`
+//!
+//! # Overview
+//!
+//! tower-http is a library that provides HTTP specific middlewares and utilities built on top of
```
hawkw, line 5: nit:
```suggestion
//! `tower-http` is a library that provides HTTP-specific middlewares and utilities built on top of
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//! [tower].
```
hawkw, line 6: nit: how about
```suggestion
//! the [`tower`] and [`http`] crates.
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//!
+//! All middlewares uses the [http] and [http-body] crates as the HTTP abstractions. That means
+//! they're compatible with any library or framework that also uses those crates, such as [hyper].
```
hawkw, line 9:
```suggestion
//! All middlewares uses the [`http`] and [`http-body`] crates as the HTTP abstractions. That means
//! they're compatible with any library or framework that also uses those crates, such as [`hyper`].
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//! Some middlewares might bring in other dependencies for doing various things.
```
hawkw, line 10: this feels a little bit vague...i'm not sure if it's _really_ necessary to have this line?

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//!         .layer(CompressionLayer::new())
+//!         // Wrap a `Service` in our middleware stack
+//!         .service(service_fn(handler));
+//!
+//!     // Run our service using some HTTP server
+//!     run_http_service(service).await;
+//! }
+//! ```
+//!
+//! # Feature toggles
```
hawkw, line 58: nit: generally I think we would call this 
```suggestion
//! # Feature Flags
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//!
+//! All middlewares are disabled by default and can be enabled using [cargo features].
+//!
+//! For example to enable [`AddExtension`] you would add this to your `Cargo.toml`:
```
hawkw, line 62:
```suggestion
//! For example, to enable the [`AddExtension`] middleware, add the "add-extension" feature flag
//! in your`Cargo.toml`:
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
+//! `async fn(HttpRequest) -> Result<HttpResponse, Error>`
+//!
+//! # Overview
+//!
+//! tower-http is a library that provides HTTP specific middlewares and utilities built on top of
+//! [tower].
+//!
+//! All middlewares uses the [http] and [http-body] crates as the HTTP abstractions. That means
```
hawkw, line 8: nit: in the `tower` docs, at least, we use "middleware" as the plural, instead of "middlewares". i think this should be 
```suggestion
//! All middleware uses the [http] and [http-body] crates as the HTTP abstractions. That means
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//! You can use `"full"` to enable everything:
+//!
+//! ```toml
+//! tower-http = { version = "0.1.0", features = ["full"] }
+//! ```
+//!
+//! [tower]: https://crates.io/crates/tower
+//! [http]: https://crates.io/crates/http
+//! [http-body]: https://crates.io/crates/http-body
+//! [hyper]: https://crates.io/crates/hyper
```
hawkw, line 77: if you apply my suggestion from above about adding code formatting to crate names, we should change this to match:
```suggestion
//! [`tower`]: https://crates.io/crates/tower
//! [`http`]: https://crates.io/crates/http
//! [`http-body`]: https://crates.io/crates/http-body
//! [`hyper`]: https://crates.io/crates/hyper
```

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//!         .service(service_fn(handler));
+//!
+//!     // Run our service using some HTTP server
+//!     run_http_service(service).await;
+//! }
+//! ```
+//!
+//! # Feature toggles
+//!
+//! All middlewares are disabled by default and can be enabled using [cargo features].
```
hawkw, line 60: as i mentioned above, i think we tend to use "middleware" as the plural
```suggestion
//! All middleware are disabled by default and can be enabled using [cargo features].
```

tower-http/src/lib.rs
hawkw: it might be nice to have a list of all the feature flags here, as well? although I suppose that if we use the `#[doc(cfg)]` attribute, the RustDoc module list is basically that...

tower-http/src/lib.rs
```diff
@@ -1,3 +1,83 @@
...
+//!     let service = ServiceBuilder::new()
+//!         // Share an `Arc<State>` with all requests
+//!         .layer(AddExtensionLayer::new(Arc::new(state)))
+//!         // Compress responses
+//!         .layer(CompressionLayer::new())
+//!         // Wrap a `Service` in our middleware stack
+//!         .service(service_fn(handler));
+//!
+//!     // Run our service using some HTTP server
+//!     run_http_service(service).await;
```
hawkw, line 54: TIOLI: since the example shows an import of `hyper::Body`, maybe we should show the actual hyper server API here?

hawkw, line 54: i'd kind of like to see a few more comments explaining what this code is doing. for example:
```suggestion
//!
//! // Our request handler. This is where we would implement the application logic
//! // for responding to HTTP requests...
//! async fn handler(request: Request<Body>) -> Result<Response<Body>, Error> {
//!     // ...
//!     # todo!()
//! }
//!
//! /// Shared state across all request handlers --- in this case, a pool of database connections.
//! struct State {
//!     pool: DatabaseConnectionPool,
//! }
//!
//! #[tokio::main]
//! async fn main() {
//!     // Construct the shared state.
//!     let state = State {
//!         pool: DatabaseConnectionPool::new(),
//!     };
//!
//!     // Use `tower`'s `ServiceBuilder` API to build a stack of `tower` middleware
//!     // wrapping our request handler.
//!     let service = ServiceBuilder::new()
//!         // Share an `Arc<State>` with all requests
//!         .layer(AddExtensionLayer::new(Arc::new(state)))
//!         // Compress responses
//!         .layer(CompressionLayer::new())
//!         // Wrap a `Service` in our middleware stack
//!         .service(service_fn(handler));
//!
//!     // Run our service using some HTTP server.
//!     // The HTTP server implementation is provided by a compatible external library,
//!     // such as `hyper`.
//!     run_http_service(service).await;
```

hawkw (commented):
i commented on a bunch of (admittedly, maybe kind of nitpicky) copy-edits. thanks for adding docs!

tower-http/src/lib.rs
davidpdrsn: Yeah I think what rustdoc does is sufficient. I quite like how it looks on tower
image

tower-http/src/lib.rs
davidpdrsn: Yeah I agree. I've removed it.

tower-http/src/lib.rs
davidpdrsn: Yeah thats a good point. At first I wanted to keep the docs hyper-free but then pulling in `hyper::Body` is silly. I've converted the example to use a hyper server and added another note about tower-http not being hyper specific.

tower-http/src/add_extension.rs
```diff
@@ -1,14 +1,22 @@
+//! Middleware for adding some shareable value to [request extensions].
+//!
+//! [request extensions]: https://docs.rs/http/latest/http/struct.Extensions.html
+
 use http::Request;
 use std::task::{Context, Poll};
 use tower_layer::Layer;
 use tower_service::Service;
 
+/// [`Layer`] for adding some shareable value to [request extensions].
```
hawkw, line 10: since `Layer` is defined in another crate, I *think* this needs a link reference.

tower-http/src/add_extension.rs
```diff
@@ -1,14 +1,22 @@
+//! Middleware for adding some shareable value to [request extensions].
```
hawkw, line 1: nit, take it or leave it: what do you think about making a little more explicit that this uses `Clone`? 
```suggestion
//! Middleware that clones a value into each request's [extensions].
```
or something?

tower-http/src/lib.rs
```diff
@@ -1,3 +1,114 @@
...
+//! All middlewares uses the [`http`] and [`http-body`] crates as the HTTP abstractions. That means
+//! they're compatible with any library or framework that also uses those crates, such as
+//! [`hyper`].
+//!
+//! # Example server
+//!
+//! This example shows how to apply middlewares from `tower-http` to a [`Service`] and then run
+//! that service using [`hyper`].
+//!
+//! ```rust,no_run
```
hawkw, line 17: alternatively we could avoid the `no_run` by just sticking the body of the `main` fn inside another function, hiding that function's definition using `#`, and never calling it. then, the docs won't say "this example is not run". :)

hawkw (approved):
lgtm, couple more suggestions, but feel free to take it or leave it!

tower-http/src/lib.rs
davidpdrsn: Is that a docs.rs thing because it doesn't say so when building the docs locally 🤔

tower-http/src/add_extension.rs
davidpdrsn: Nope it seems to be working.

davidpdrsn: Honestly not sure why. There are other times where it doesn't work.

PR #44, opened 2021-01-30
Add `Redirect` service

davidpdrsn:
Little utility service that redirects all requests.

### DIFF
--- a/tower-http/src/redirect.rs
+++ b/tower-http/src/redirect.rs
@@ -3,6 +3,7 @@
 use http::{header, HeaderValue, Response, StatusCode, Uri};
 use std::{
     convert::Infallible,
+    fmt,
     future::Future,
     marker::PhantomData,
     pin::Pin,
@@ -41,7 +42,6 @@ use tower_service::Service;
 /// # Ok(())
 /// # }
 /// ```
-#[derive(Clone, Debug)]
 pub struct Redirect<ResBody> {
     status_code: StatusCode,
     location: HeaderValue,
@@ -103,6 +103,25 @@ where
         ResponseFuture {
             status_code: self.status_code,
             location: Some(self.location.clone()),
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<ResBody> fmt::Debug for Redirect<ResBody> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Redirect")
+            .field("status_code", &self.status_code)
+            .field("location", &self.location)
+            .finish()
+    }
+}
+
+impl<ResBody> Clone for Redirect<ResBody> {
+    fn clone(&self) -> Self {
+        Self {
+            status_code: self.status_code,
+            location: self.location.clone(),
             _marker: PhantomData,
         }
     }
--- a/tower-http/src/redirect.rs
+++ b/tower-http/src/redirect.rs
@@ -1,0 +1,100 @@
+//! Service that redirects all requests.
+
+use http::{header, HeaderValue, Request, Response, StatusCode, Uri};
+use std::{
+    convert::Infallible,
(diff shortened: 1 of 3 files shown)


tower-http/src/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+    _marker: PhantomData<fn() -> ResBody>,
+}
+
+impl<ResBody> Redirect<ResBody> {
+    /// Create a new [`Redirect`] that uses a `307 Temporary Redirect` status code.
+    pub fn temporary(uri: Uri) -> Self {
+        Self::with_status_code(StatusCode::TEMPORARY_REDIRECT, uri)
+    }
+
+    /// Create a new [`Redirect`] that uses a `308 Permanent Redirect` status code.
```
hawkw, line 27: nit, take it or leave it: might be nice to link to docs for these status codes here --- maybe MDN?

tower-http/src/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    fn poll(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut res = Response::default();
+
+        *res.status_mut() = self.status_code;
+
+        res.headers_mut().insert(
+            header::LOCATION,
+            HeaderValue::from_str(&self.uri.take().unwrap().to_string())
+                .expect("URI isn't a valid header value"),
```
hawkw, line 95: since the URI is fixed when the service is built, maybe we should do the `Uri` -> `HeaderValue` conversion in the constructor? that way, we don't have to do this on every request, which might be a bit more efficient --- especially because this allocates twice (once for `to_string` and once for `HeaderValue::from_str`). 

that way, we could also add "If the URI contains characters that aren't a valid header value" to the list of panics given in the constructor...although I _think_ that if a string was a valid URI, it should also be a valid header value, so I don't think we can ever actually hit that panic.

tower-http/src/redirect.rs
```diff
@@ -0,0 +1,100 @@
+//! Service that redirects all requests.
```
hawkw, line 1: nit: an example might be nice...but we can do that later.

tower-http/src/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+    convert::Infallible,
+    future::Future,
+    marker::PhantomData,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_service::Service;
+
+/// Service that redirects all requests.
+#[derive(Clone, Debug)]
```
hawkw, line 14: nit: the derived impls will require `ResBody` to be `Debug` and `Clone`. i think we should probably have manual impls, since this doesn't actually contain a `ResBody`...

hawkw (changes requested):
This is a good start. I had a few suggestions.

tower-http/src/redirect.rs
davidpdrsn: I'm beginning to think about bringing in some proc-macro crate that has more flexible `Clone`, `Debug, etc, derivation 😅

tower-http/src/redirect.rs
davidpdrsn: Good point 👍

tower-http/src/redirect.rs
davidpdrsn: 👍

tower-http/src/redirect.rs
davidpdrsn: 👍

tower-http/src/redirect.rs
hawkw: in general, i think "don't derive `Clone` or `Debug` for types with `PhantomData`" is probably a good rule of thumb

hawkw (approved):

PR #47, opened 2021-01-31
Remove duplication between compression and decompression

davidpdrsn:
While technically this is a net increase in code there are few things that could be upstreamed to http-body which would bring the SLOC down.


tower-http/src/compression_utils.rs
```diff
@@ -0,0 +1,310 @@
...
+    type Output: AsyncRead;
+
+    /// Apply the decorator
+    fn apply(input: Self::Input) -> Self::Output;
+
+    /// Get a pinned mutable reference to the original input.
+    ///
+    /// This is necessary to implement `Body::poll_trailers`.
+    fn get_pin_mut(pinned: Pin<&mut Self::Output>) -> Pin<&mut Self::Input>;
+}
```
davidpdrsn, line 118: @tesaguri what do you think about this (+ `WrapBody` below) as a way of applying an `AsyncRead` decorator to a `Body`?

The need to access the inner body as a `Pin<&mut B>` complicates things a bit but its necessary to implement `poll_trailers`.

tower-http/src/compression_utils.rs
```diff
@@ -0,0 +1,310 @@
...
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Option<http::HeaderMap>, Self::Error>> {
+        let this = self.project();
+        match this.inner.poll_trailers(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Ok(None)) => Poll::Ready(Ok(None)),
+            Poll::Ready(Ok(Some(headers))) => Poll::Ready(Ok(Some(headers))),
+            Poll::Ready(Err(err)) => Poll::Ready(Err((this.f)(err))),
+        }
```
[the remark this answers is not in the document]
davidpdrsn: Applied manually 👍

tower-http/src/compression/body.rs
```diff
@@ -1,320 +1,249 @@
...
+pub(crate) enum BodyInner<B>
+where
+    B: Body,
+{
+    #[cfg(feature = "compression-gzip")]
+    Gzip(
+        #[pin]
+        WrapBody<
+            BodyMapErr<B, fn(<B as Body>::Error) -> io::Error>,
+            GzipEncoder<BodyMapErr<B, fn(<B as Body>::Error) -> io::Error>>,
```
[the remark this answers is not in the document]
davidpdrsn: Good idea. That cleaned it up quite a bit.

tower-http/src/compression_utils.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// This is necessary to implement `Body::poll_trailers`.
+    fn get_pin_mut(pinned: Pin<&mut Self::Output>) -> Pin<&mut Self::Input>;
+}
+
+/// `Body` that has been decorated by an `AsyncRead`
+#[pin_project]
+pub(crate) struct WrapBody<B, M: DecorateAsyncRead> {
+    #[pin]
+    pub(crate) read: M::Output,
+    _marker: PhantomData<B>,
```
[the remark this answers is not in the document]
davidpdrsn: Uhh yeah thats true! Good catch. Simplified things quite a bit 🎊

tower-http/src/compression_utils.rs
```diff
@@ -0,0 +1,310 @@
...
+}
+
+impl<B, R, M> WrapBody<B, M>
+where
+    B: Body,
+    B::Error: Into<io::Error>,
+    M: DecorateAsyncRead<Input = StreamReader<BodyIntoStream<B>, B::Data>, Output = R>,
```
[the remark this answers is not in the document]
davidpdrsn: Applied manually 👍

tower-http/Cargo.toml
```diff
@@ -26,6 +26,7 @@ async-compression = { version = "0.3", optional = true, features = ["deflate", "
 bytes = { version = "1", optional = true }
 hyper = { version = "0.14", optional = true, default_features = false, features = ["stream"] }
 tokio-util = { version = "0.6", optional = true, default_features = false, features = ["io", "codec"] }
```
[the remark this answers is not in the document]
davidpdrsn: Good catch 🙏

tower-http/src/compression_utils.rs
davidpdrsn: Right its not trivial. Hopefully the changes you suggested have made things a bit easier to follow.

tower-http/Cargo.toml
```diff
@@ -25,7 +25,8 @@ tower-service = "0.3"
 async-compression = { version = "0.3", optional = true, features = ["tokio"] }
 bytes = { version = "1", optional = true }
 hyper = { version = "0.14", optional = true, default_features = false, features = ["stream"] }
-tokio-util = { version = "0.6", optional = true, default_features = false, features = ["io", "codec"] }
+tokio-util = { version = "0.6", optional = true, default_features = false, features = ["io"] }
+tokio = { version = "1", optional = true, default_features = false }
```
LucioFranco, line 29: I thought tokio didn't bring in any default features? but I guess this doesn't hurt

tower-http/src/compression/body.rs
```diff
@@ -1,320 +1,233 @@
...
 
 /// Response body of [`Compression`].
 ///
 /// [`Compression`]: super::Compression
 #[pin_project]
-pub struct CompressionBody<B: Body> {
-    #[pin]
-    inner: BodyInner<B, B::Error>,
-}
+pub struct CompressionBody<B>(#[pin] pub(crate) BodyInner<B>)
```
LucioFranco, line 28: Why this change in struct?

tower-http/src/compression/body.rs
```diff
@@ -1,320 +1,233 @@
...
-                BodyInnerProj::Identity(inner, _) => {
-                    return match ready!(inner.poll_data(cx)) {
-                        Some(Ok(mut data)) => {
-                            Poll::Ready(Some(Ok(data.copy_to_bytes(data.remaining()))))
-                        }
-                        Some(Err(e)) => Poll::Ready(Some(Err(Error::Body(e)))),
-                        None => Poll::Ready(None),
-                    };
+        match self.project().0.project() {
+            #[cfg(feature = "compression-gzip")]
```
LucioFranco, line 148: Do we have a way in CI right now to check all these variations of feature flags?

tower-http/src/compression/body.rs
```diff
@@ -1,320 +1,233 @@
...
-struct Adapter<B, E> {
-    #[pin]
-    body: B,
-    error: Option<E>,
-}
+#[cfg(feature = "compression-gzip")]
+impl<B> DecorateAsyncRead for GzipEncoder<B>
+where
+    B: Body,
+    B::Error: Into<io::Error>,
```
LucioFranco, line 185: This bound won't work for hyper? https://docs.rs/hyper/0.14.4/hyper/struct.Body.html#impl-Body

tower-http/src/compression/future.rs
```diff
@@ -21,14 +25,43 @@ pub struct ResponseFuture<F> {
 
 impl<F, B, E> Future for ResponseFuture<F>
 where
-    F: Future<Output = Result<http::Response<B>, E>>,
+    F: Future<Output = Result<Response<B>, E>>,
     B: Body,
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
```
LucioFranco, line 30: might be worth it to just define a `type BoxError`

tower-http/src/compression/body.rs
davidpdrsn: Yes https://github.com/tower-rs/tower-http/blob/master/.github/workflows/CI.yml#L46. Learned about this from tower's GH actions setup.

tower-http/src/compression/body.rs
davidpdrsn: Yes. The error is converted into an `io::Error` using `io::ErrorKind::Other` [here](https://github.com/tower-rs/tower-http/blob/dedup-compression-and-decompression/tower-http/src/compression_utils.rs#L301).

Just to be sure I added a test in 788390426b175.

tower-http/src/compression/body.rs
davidpdrsn: Don't exactly remember. I think I moved some things around and had to make the field `pub(crate)` to create it from another module. And making it a tuple struct made things a bit less verbose in other places.

tower-http/Cargo.toml
davidpdrsn: Ah yeah true. Habit but yeah shouldn't hurt.

tower-http/src/compression/future.rs
davidpdrsn: 36f8084acff29 👍

tower-http/src/compression/body.rs
LucioFranco: I am not sure I like that, why not create a specific error type so that a user can potentially downcast to that type instead of having to string match on an io error?

davidpdrsn: Yeah I agree it’s not great. The fact that AsyncRead requires using io::Error complicates things a bit since Body can have any error type.

The original version of this did actually have a worker around where it would store an error of another type and keep a sentinel io::Error that would be converted later.

I’ll give it another look 👀

davidpdrsn: I found a fix. 726144bae566d.

Added [`StreamErrorIntoIoError`](https://github.com/tower-rs/tower-http/blob/dedup-compression-and-decompression/tower-http/src/compression_utils.rs#L291) which converts errors from a stream into a sentinel `io::Error` and allowing you to get out the actual error later.

So `CompressionBody` and `DecompressionBody` now both use [`BodyOrIoError`](https://github.com/tower-rs/tower-http/blob/dedup-compression-and-decompression/tower-http/src/lib.rs#L208) as their error type and so retains the original errors from the `Body`.

tower-http/src/compression/future.rs
```diff
@@ -21,14 +25,40 @@ pub struct ResponseFuture<F> {
...
+                    CompressionBody(BodyInner::Identity(body)),
+                )))
+            }
+        };
+
+        parts.headers.remove(header::CONTENT_LENGTH);
+
+        parts.headers.insert(
+            header::CONTENT_ENCODING,
+            HeaderValue::from_str(self.encoding.to_str()).unwrap(),
```
LucioFranco, line 58: iirc `HeaderValue` is backed by a `Bytes` so maybe it makes sense to store this header value in the struct and make it a ref count bump?

tower-http/src/compression/mod.rs
```diff
@@ -104,3 +101,61 @@ fn encodings(headers: &HeaderMap, accept: AcceptEncoding) -> Vec<(Encoding, f32)
...
+    async fn works() {
+        let svc = service_fn(handle);
+        let mut svc = Compression::new(svc);
+
+        // call the service
+        let req = Request::builder()
+            .header("accept-encoding", "gzip")
+            .body(Body::empty())
+            .unwrap();
+        let res = svc.ready_and().await.unwrap().call(req).await.unwrap();
```
LucioFranco, line 125: deprecation warnings here!

LucioFranco (approved):
LGTM! Left a suggestion inline but otherwise lets ship it!

tower-http/src/compression/future.rs
davidpdrsn: Good point! The header value is created from the request but we can use `HeaderValue::from_static` as the encodings are an enum.

tower-http/src/compression/mod.rs
davidpdrsn: 👍

PR #49, opened 2021-02-06
Add response classification

davidpdrsn:
The response classification in https://github.com/tower-rs/tower-http/pull/40 didn't work because it didn't properly support gRPC. I have given it some more thought and now have something that does support gRPC. Its inspired by the implementation in [linkerd-proxy](https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/http-classify/src/lib.rs).

This is still a draft since I haven't implemented all of the `Trace` middleware. At this point its just a proof of concept that shows the classification stuff used in a middleware.

@hawkw Please have a look and let me know what you think. If you think the classification could work then I'll go ahead and implement the remaining pieces of the middleware.


tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn classify_response<B>(
+        &mut self,
+        res: &Response<B>,
+    ) -> ClassifiedNowOrLater<Self::Output, Self::ClassifyEos>
+    where
+        B: Body;
+}
+
+/// Trait for classifying end of streams as either success or failure.
+pub trait ClassifyEos {
```
davidbarsky, line 27: This took me a second to figure out, but maybe define "eos" as prior first use? It doesn't seem to be the _most_ common acronym. Could something like this work?

```suggestion
/// Trait for classifying end of streams ("Eos") as either success or failure.
pub trait ClassifyEos {
```

tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// Trait for classifying responses as either success or failure.
+pub trait ClassifyResponse {
+    /// The output of doing the classification.
+    ///
+    /// This could be an HTTP status code or some other kind of status.
+    type Output;
+
+    /// The type used to classify the end of streaming responses.
+    type ClassifyEos: ClassifyEos<Output = Self::Output>;
```
davidbarsky, line 15: See my other comment, but I'd maybe define "Eos" in the documentation:

```suggestion
    /// The type used to classify the end of streaming (Eos) responses.
    type ClassifyEos: ClassifyEos<Output = Self::Output>;
```

tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+/// The default classifier used for normal HTTP responses.
+///
+/// Responses with a `5xx` status code are failures, all others as successes.
+#[derive(Clone, Debug, Default)]
+pub struct ServerErrorsAsFailures {
+    _priv: (),
+}
```
hawkw, line 80: another premade classifier that might be nice to have is one that just takes a list of specific status codes to treat as failures, and treats everything else as a success, or one that takes a list of successes and treats everything else as failures. we can add that later, but i just thought i'd mention.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+use http::{HeaderMap, Response, StatusCode};
+use http_body::Body;
+use std::{convert::Infallible, marker::PhantomData};
+
+/// Trait for classifying responses as either success or failure.
+pub trait ClassifyResponse {
+    /// The output of doing the classification.
+    ///
+    /// This could be an HTTP status code or some other kind of status.
+    type Output;
```
hawkw, line 12: wondering if this name could be a little more descriptive...maybe `Class`? but, that seems kind of ambiguous with the `Classification` type... hmm...

tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Perform the classification.
+    fn classify_eos(&mut self, trailers: Option<&HeaderMap>) -> Classification<Self::Output>;
+}
+
+/// Result of doing a classification.
+#[derive(Debug)]
+// really don't like this name... :(
```
hawkw, line 39: perhaps unsurprisingly i'm not a fan either...we could try to brainstorm something better, but it does get the point across...

hawkw (commented):
Thanks for continuing to work on that --- I think the new classification proposal is definitely a step in the right direction. 

One thing I think could be a potential hole in this design is that we currently classify _responses_ and _successful_ EOSes, based on their status codes/trailers, but we don't classify _Rust errors_. if, for example, a streaming response closes abruptly with an IO error, instead of cleanly with trailers, we might want to classify that failure as well --- obviously, it's a failure, but what _kind_ of failure? this is particularly important if we want to be able to use the response classifier traits for retries as well as logging & metrics --- what IO errors are retryable? what protocol errors are retryable? and so on.

We could add this to the proposed API in a couple of ways. We could do what the Linkerd classifier trait does, and add a separate `error` method to the classifier traits, like
```rust
pub trait ClassifyResponse {
    // ...

    fn classify_response<B>(&mut self, /* ... */) -> ClassifiedNowOrLater</* ... */>
    where
        B: Body,

    fn classify_error<E>(&mut self, &E) -> Classification<Self::Output>
    where 
        E: Error;
}

pub trait ClassifyEos {
    // ...

    fn classify_eos(&mut self,  /* ... */) -> Classification<Self::Output>;

    fn classify_stream_error<E>(&mut self, e: E) -> Classification<Self::Output>
    where 
        E: Error;
}
```

 Alternatively, we could change the classification methods to something like 
```rust
pub trait ClassifyResponse {
    // ...

    fn classify_response<B, E>(
        &mut self,
        res: Result<&Response<B>, E>,
    ) -> ClassifiedNowOrLater<Self::Output, Self::ClassifyEos>
    where
        B: Body,
        E: Error;
}

pub trait ClassifyEos {
    // ...
    
    fn classify_eos<E>(&mut self, trailers: Result<Option<&HeaderMap>, E>) -> Classification<Self::Output>
    where
        E: Error;
}
```
but that seems maybe a little more convoluted.

Alternatively, we might decide to punt on errors here, and if we introduce retries later, we could have a separate interface for determining whether Rust errors are retryable.

Another thing that's worth considering is whether the _request_ should be included in the classification. This, again, is largely an issue if we want to use this trait for determining what responses are retryable. Whether a request should be retried might depend on if its method is  [idempotent](https://developer.mozilla.org/en-US/docs/Glossary/Idempotent) (you don't want to retry a `POST` request that might result in data being changed multiple times, for example), or whether the request had a streaming body (you might not want to retry a request that requires you to buffer the entire body to send it again). 

The Linkerd classification system deals with this by adding an additional trait layer that takes a reference to a `Request` and _returns_ a classifier, where that classifier could be configured based on the request --- e.g. it could store the request method. I think the current response classification system proposed here is compatible with a similar approach. We could add an additional trait for producing response classifiers for individual requests.

We don't need to add that in this PR --- in fact, I think we should probably prefer not to add even more stuff to block this branch :wink:. I don't think the tracing layer's use of response classifiers really needs that, anyway --- it can just use the same classifier everywhere. But, we'll need something similar if we want to use this for retries. I just thought I'd bring it up now as a note for the future :)

davidpdrsn:
Alright. I've made some more changes. I think it would be very nice if we could use the same classification traits for retries as well as long as it doesn't complicate the API unnecessarily.

I've added the `MakeClassifier` thing you mentioned and I don't think it'll have much impact on users who just want the default tracing or metrics layers. Also added `SharedClassifier` which turns a `T: ClassifyResponse + Clone` into a `MakeClassifier`. I'm thinking that is nice for the same reasons `tower::make::Shared` is nice.

I think adding `classify_error` makes sense as well. I think the error bound should be `E: std::error::Error + 'static` though. That way users are able to do `(error as &dyn Error).downcast_ref::<io::Error>()` to check the actual type. I think that non `'static` errors are very rare especially in web apps since there are threads everywhere which basically require `'static`.

For `ServerErrorsAsFailures` and `GrpcErrorsAsFailures`'s implementations of `classify_error` I guess they should return `StatusCode::INTERNAL_SERVER_ERROR` and `13` respectively but feels a bit ad hoc. Do you have other suggestions?

I've also added an implementation of a retry policy based on a classifier like you mentioned. Seems like it could work.

If you squint it is basically the classifier thats in linkerd-proxy 😅 makes sense I guess and not necessarily a bad thing. I think most users will be able to use `Trace::new()` and just move on with their lives.

Still a bunch more docs missing. I'll add those when we're happy with the API.

tower-http/src/classify.rs
davidpdrsn: Yeah that would be cool.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+pub struct GrpcEosErrorsAsFailures {
+    _priv: (),
+}
+
+impl ClassifyEos for GrpcEosErrorsAsFailures {
+    type Output = i32;
+
+    fn classify_eos(&mut self, trailers: Option<&HeaderMap>) -> Classification<Self::Output> {
+        classify_grpc_metadata(trailers.expect("no trailers"))
+            .expect("no or invalid grpc-status in trailers")
```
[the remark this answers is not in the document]
davidpdrsn: Good point. Indeed doesn't make sense to panic. I've changed it to return success.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/// A classification of a response.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Classification<T> {
+    /// The response was a success.
+    Success(T),
+    /// The response was a failure.
+    Failure(T),
+}
```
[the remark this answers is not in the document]
davidpdrsn: I tried changing it to `Result` and actually think its fine 👍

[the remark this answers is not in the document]
davidpdrsn: Hm yeah I think removing the success value is fine. The tracing and metrics layers are still able to access the status if necessary.

I also went ahead and built a retry policy based on your suggestion. Seems like it could work.

tower-http/src/classify.rs
davidpdrsn: In linkerd-proxy you write `Produce a stream classifier for this response`. Do you mean a stream like `Stream<Item = _>` which then gets polled somewhere?

Other ideas:

- `TryClassify`
- `ResponseClassification`
- `ClassifiedResponse`

tower-http/src/classify.rs
davidpdrsn: I went with `FailureClass` in my next iteration because the success case no longer as any data, so this type is only included in the failure case.

davidpdrsn:
I've become unsure about what the bound on `E` should be for `classify_error`. I've seen `E: std::error::Error` lead to issues and have had to use `E: Into<tower::BoxError>` instead. But I'm not sure that fits here...

tower-http/src/classify.rs
hawkw: "stream" here refers to "streaming bodies", not the `Stream` trait, but yes, it produces a classifier for the response body stream. The name `Stream` might not be good to use here because of conflicts with the trait?

hawkw:
> I've become unsure about what the bound on `E` should be for `classify_error`. I've seen `E: std::error::Error` lead to issues and have had to use `E: Into<tower::BoxError>` instead. But I'm not sure that fits here...

I think the `BoxError` bound is probably okay. Alternatively, we make could:
 - Make the _trait_ generic over an arbitrary `E`, so that implementations can introduce their own bounds or concrete types. That would be the most flexible but would introduce some additional API complexity.
 - Just always pass in a `BoxError`, to make the `classify_error` implementation less complex. In practice, when we make the method generic over `E: Into<BoxError>`, we are essentially doing the same thing; we're saying that the only thing the classifier can do with the error is turn it into a `BoxError`. So, if we're going to do that, writing a classifier is a little simpler if the code that _uses_ the classifier just turns the error into a `BoxError` for it.

I think the first approach, making the trait generic over an error, _might_ be the best, but it would mean a lot of stuff has to be generic over an error type param now...I'm not sure if it's worth it or not.

tower-http/src/classify.rs
davidpdrsn: I've gone with `ClassifiedResponse` for now.

davidpdrsn:
I've changed it to be generic over the error type. I think that is the most flexible approach and didn't complicate the API much.

I have also removed the temporary `Trace` middleware. Will rebuild it once this is merged.

And finally I adjusted some docs a bit.

So with all that I think this is ready for review 🙏

tower-http/src/classify.rs
```diff
@@ -0,0 +1,313 @@
...
+    >;
+
+    /// The type of failure classifications.
+    type FailureClass;
+
+    /// The type used to classify the response end of stream (EOS).
+    type ClassifyEos: ClassifyEos<FailureClass = Self::FailureClass>;
+
+    /// Make a response classifier for this request
+    fn make_classify<B>(&self, req: &Request<B>) -> Self::Classifier
```
lpil, line 29:
```suggestion
    fn make_classifier<B>(&self, req: &Request<B>) -> Self::Classifier
```
To match the type name?

tower-http/src/classify.rs
```diff
@@ -0,0 +1,313 @@
...
+        // `NeverClassifyEos` contains an `Infallible` so it can never be constructed
+        unreachable!()
+    }
+}
+
+/// The default classifier used for normal HTTP responses.
+///
+/// Responses with a `5xx` status code are failures, all others as successes.
+#[derive(Clone, Debug, Default)]
+pub struct ServerErrorsAsFailures {
```
lpil, line 140: Would one that considered client errors as well as server errors errors be useful? Similar to what reqwest does

https://docs.rs/reqwest/0.11.1/src/reqwest/async_impl/response.rs.html#341

tower-http/src/classify.rs
```diff
@@ -0,0 +1,313 @@
...
+    {
+        if let Some(classification) = classify_grpc_metadata(res.headers()) {
+            ClassifiedResponse::Ready(classification)
+        } else {
+            ClassifiedResponse::RequiresEos(GrpcEosErrorsAsFailures { _priv: () })
+        }
+    }
+
+    fn classify_error(self, _error: &E) -> Self::FailureClass {
+        13
```
lpil, line 225: I'm not familiar with grpc so much, what does 13 refer to?

lpil (approved):
Looks good!

tower-http/src/classify.rs
davidpdrsn: Its gRPC's version of "internal server error". I added a link that should explain it.

lpil: Thanks!

tower-http/src/classify.rs
davidpdrsn: Yeah totally. I was thinking about adding one that took a list of ranges that would be considered errors. I think I'll do that in a follow up PR though.

Made an issue for tracking that https://github.com/tower-rs/tower-http/issues/73

davidpdrsn:
While implementing a `Trace` middleware I ran into `ClassifyEos` not having a `classify_error` method. I guess it should have one in case a streaming body returns an error. I added it.

davidpdrsn:
Also removed `where B: Body` bound. You can't really do much with the body if you only have a `&Request<B>` or `&Response<B>`.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+        StatusCode::INTERNAL_SERVER_ERROR
+    }
+}
+
+/// Response classifier for gRPC responses.
+///
+/// gRPC doesn't use normal HTTP statuses for indicating success or failure but instead a special
+/// header that might appear in a trailer.
+///
+/// Responses where `grpc-status` is 0 are successes, all others are failures.
```
hawkw, line 186: This implies that this classifies responses where there is no `grpc-status` trailer as failures, but looking at the implementation, it looks like they are considered successes: https://github.com/tower-rs/tower-http/blob/a220f3768e0f0dcc8798068992eb54f8f7c57450/tower-http/src/classify.rs#L237

I do think this is probably the correct behavior, so let's not change that; we should just make this a little clearer in the docs.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct GrpcErrorsAsFailures {
+    _priv: (),
+}
```
hawkw, line 190: afaict this thing's job is just to extract the gRPC status code from the trailers, and handle the cases where its missing or the body errors? it might be worth making this clear in the docs?

hypothetically i can see this getting chained/mapped with another classifier for determining whether or not those status codes are retryable, as part of a retry policy? i think it could be good to add combinators for response classifiers in a future branch so that it's easy to express things like that...

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+
+/// A [`MakeClassifier`] that works by cloning a classifier.
+///
+/// If you have a [`ClassifyResponse`] that doesn't depend on the request you can use
+/// [`SharedClassifier`] to get a [`MakeClassifier`] for your classifier.
+#[derive(Debug, Clone)]
+pub struct SharedClassifier<C> {
+    classifier: C,
+}
```
hawkw, line 38: it seems like this _could_ be a blanket impl for `C: ClassifyResponse<E> + Clone`. I'm not sure whether that's a good idea, since it might conflict with other blanket impls, and this is more explicit, so that's just a thought. We could always add that later if it turns out to be a common use case.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+    >;
+
+    /// The type of failure classifications.
+    type FailureClass;
+
+    /// The type used to classify the response end of stream (EOS).
+    type ClassifyEos: ClassifyEos<E, FailureClass = Self::FailureClass>;
+
+    /// Make a response classifier for this request
+    fn make_classifier<B>(&self, req: &Request<B>) -> Self::Classifier;
```
hawkw, line 28: there should probably either be a blanket impl for `F: Fn(Request<B>) -> ClassifyResponse<E>`, or an adapter for making a function into a `MakeClassifier`?

can add that separately of course.

hawkw (approved):
I think this looks really good. I had some last-minute API thoughts, but I don't think they're really blocking.

Also, I kind of rewrote most of the docs in this branch...sorry about that :)

tower-http/src/classify.rs
```diff
@@ -0,0 +1,330 @@
...
+        E,
+        FailureClass = Self::FailureClass,
+        ClassifyEos = Self::ClassifyEos,
+    >;
+
+    /// The type of failure classifications.
+    type FailureClass;
+
+    /// The type used to classify the response end of stream (EOS).
+    type ClassifyEos: ClassifyEos<E, FailureClass = Self::FailureClass>;
```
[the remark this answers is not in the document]
davidpdrsn: I think having the extra associated types are nice. I would be fine writing `<M::Classifier as ClassifyResponse<E>>::FailureClass: SomeTrait` but I imagine `M::FailureClass: SomeTrait` is easier to understand if someone comes across it in the docs. It also matches `IntoIterator`, `IntoFuture`, and `MakeService`.

tower-http/src/classify.rs
davidpdrsn: Good point 👍 Made an issue for it https://github.com/tower-rs/tower-http/issues/74

tower-http/src/classify.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// Make a response classifier for this request
+    fn make_classifier<B>(&self, req: &Request<B>) -> Self::Classifier
+    where
+        B: Body;
+}
+
+/// A [`MakeClassifier`] that works by cloning a classifier.
+///
+/// If you have a [`ClassifyResponse`] that doesn't depend on the request you can use
+/// [`SharedClassifier`] to get a [`MakeClassifier`] for your classifier.
```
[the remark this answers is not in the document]
davidpdrsn: https://github.com/tower-rs/tower-http/issues/75 👍

tower-http/src/classify.rs
davidpdrsn: Right. I think it would be more useful to keep the blanket impl for closures but I'm not quite sure if they would overlap.

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+impl<E> ClassifyEos<E> for GrpcEosErrorsAsFailures {
+    type FailureClass = i32;
+
+    fn classify_eos(self, trailers: Option<&HeaderMap>) -> Result<(), i32> {
+        trailers.and_then(classify_grpc_metadata).unwrap_or(Ok(()))
+    }
+
+    fn classify_error(self, _error: &E) -> Self::FailureClass {
+        // https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+        13
```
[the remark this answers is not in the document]
davidpdrsn: Yeah I agree it feels very "ad-hoc". I also think `ServerErrorsAsFailures` returning "internal server" error for `classify_error` is weird for the same reason. Perhaps returning an enum thats either a code or "error ¯\_(ツ)_/¯".

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+/// The [`ClassifyEos`] for [`GrpcErrorsAsFailures`].
+#[derive(Debug, Clone)]
+pub struct GrpcEosErrorsAsFailures {
+    _priv: (),
+}
+
+impl<E> ClassifyEos<E> for GrpcEosErrorsAsFailures {
+    type FailureClass = i32;
```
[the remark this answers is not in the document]
davidpdrsn: It would be nice if there was a gRPC equivalent of `http`.

tower-http/src/classify.rs
davidpdrsn: Good point. I made it a bit more clear in 4e9f328046de1

tower-http/src/classify.rs
```diff
@@ -0,0 +1,316 @@
...
+    if status == 0 {
+        Some(Ok(()))
+    } else {
+        Some(Err(status))
+    }
+}
+
+// Just verify that we can actually use this response classifier to determine retries as well
+#[cfg(test)]
+mod usable_for_retries {
```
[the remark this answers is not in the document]
davidpdrsn: Yeah that would make sense.

PR #50, opened 2021-02-09
Add `get_ref`, `get_mut`, and `into_inner` to all middlewares

davidpdrsn:
Defines a little macro for hiding the boilerplate. It currently assumes the inner service type is named `S` but thats easy to fix if necessary.

### DIFF
--- a/tower-http/src/macros.rs
+++ b/tower-http/src/macros.rs
@@ -16,4 +16,3 @@ macro_rules! define_inner_service_accessors {
         }
     };
 }
-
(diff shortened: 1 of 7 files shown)


hawkw (approved):
cool!

tower-http/src/macros.rs
```diff
@@ -0,0 +1,19 @@
...
+            &mut self.inner
+        }
+
+        /// Consumes `self`, returning the underlying service.
+        pub fn into_inner(self) -> S {
+            self.inner
+        }
+    };
+}
+
```
hawkw, line 19: rustfmt is angry about the trailing newline
```suggestion
```

PR #52, opened 2021-02-10
Fix specification of algorithm featues of `async-compression`

tesaguri:
The algorithm features are already enabled by `{de,}compression-*` features individually and should not be enabled unconditionally, as doing that spoils the whole point of the feature set.


davidpdrsn (approved):
Good catch. Cargo features have so many rough edges.

PR #53, opened 2021-02-10
Move "leaf services" into their own module

davidpdrsn:
As discussed on Discord we think it makes sense to group leaf services together in their own module as to not mix them up with middlewares. We don't think the middlewares belong in their own module since they're likely to be the most interesting part of `tower-http` and so we want them visible in the root module docs.

Suggested names of this module:

- `leaf_services` - Technically correct but not a term I expect users to be familiar with.
- `services` - Probably clearer than `leaf_services` but one could be pedantic and argue that middlewares are also services.

Out of these two I personally prefer `services`.

@tesaguri @LucioFranco @olix0r @hawkw any opinions on the module name?

### DIFF
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -173,8 +173,6 @@ pub mod sensitive_header;
 #[cfg_attr(docsrs, doc(cfg(feature = "decompression")))]
 pub mod decompression;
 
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):
I guess you can say you _redirected_ this service 😆

Issue #66, opened 2021-02-17
SetRequestHeader middleware


davidpdrsn:
Once https://github.com/tower-rs/tower-http/pull/39 is done we should add a `SetRequestHeader` middleware. The API should be the same except inserting headings to requests rather than responses.

PR #67, opened 2021-02-19
Add `SetRequestHeader`

davidpdrsn:
Exactly the same as `SetResponseHeader` but modifies the request instead. Useful on clients for example for setting `User-Agent`.

As it shares `MakeHeaderValue` and `InsertHeaderMode` with `SetResponseHeader` I moved these two middlewares into a `set_header` module.

Fixes https://github.com/tower-rs/tower-http/issues/66

### DIFF
--- a/tower-http/src/set_header/request.rs
+++ b/tower-http/src/set_header/request.rs
@@ -16,7 +16,7 @@
 //!
 //! # #[tokio::main]
 //! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
-//! # let render_html = tower::service_fn(|_: Request<Body>| async move {
+//! # let http_client = tower::service_fn(|_: Request<Body>| async move {
 //! #     Ok::<_, std::convert::Infallible>(Response::new(Body::empty()))
 //! # });
 //! #
@@ -34,7 +34,7 @@
 //!             HeaderValue::from_static("my very cool app"),
 //!         )
 //!     )
-//!     .service(render_html);
+//!     .service(http_client);
 //!
 //! let request = Request::new(Body::empty());
 //!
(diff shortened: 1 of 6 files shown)


tower-http/src/set_header/request.rs
```diff
@@ -0,0 +1,255 @@
...
+//! use http::{Request, Response, header::{self, HeaderValue}};
+//! use tower::{Service, ServiceExt, ServiceBuilder};
+//! use tower_http::set_header::SetRequestHeaderLayer;
+//! use hyper::Body;
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! # let render_html = tower::service_fn(|request: Request<Body>| async move {
+//! #     Ok::<_, std::convert::Infallible>(Response::new(Body::from("1234567890")))
+//! # });
```
tesaguri, line 59:
```suggestion
//! # let render_html = tower::service_fn(|_: Request<Body>| async move {
//! #     Ok::<_, std::convert::Infallible>(Response::new(Body::empty()))
//! # });
```

tower-http/src/set_header/request.rs
```diff
@@ -0,0 +1,255 @@
...
+//!         // we don't use a closure to produce the header value.
+//!         //
+//!         // `if_not_present` will only insert the header if it does not already
+//!         // have a value.
+//!         SetRequestHeaderLayer::<_, Body>::if_not_present(
+//!             header::USER_AGENT,
+//!             HeaderValue::from_static("my very cool app"),
+//!         )
+//!     )
+//!     .service(render_html);
```
tesaguri, line 37: I believe this example is an HTTP client, but `render_html` sounds rather like a server. Maybe the name should be something like `client` (or `http_client` to be explicit)? Or just passing `hyper::Client::new()` might be fine since the example already uses `hyper` for `hyper::Body`.

davidpdrsn: Good point! I went with `http_client` 937f3a7679058

hawkw (approved):

Issue #62, opened 2021-02-12
Example: HTTP API built with warp


davidpdrsn:
I think a complete example of a simple HTTP API built with warp and enhanced with tower-http would be awesome. Should help users put everything together.

I think a super basic key/value store would be a good project.

PR #69, opened 2021-02-19
Warp example

davidpdrsn:
Adds a little key/value store API built with warp and extended with tower + tower-http.

Fixes https://github.com/tower-rs/tower-http/issues/62

### DIFF
--- a/examples/warp-key-value-store/README.md
+++ b/examples/warp-key-value-store/README.md
@@ -1,0 +1,14 @@
+# warp-key-value-store
+
+This examples contains a simple key/value store with an HTTP API built using warp.
+
+## Endpoints
+
+- `GET /:key` - Look up a key. If the key doesn't exist it returns `400 Not Found`
+- `POST /:key` - Insert a key. The value is the request body.
+
+## Running the example
+
+```
+cargo run --bin warp-key-value-store
+```
--- a/examples/warp-key-value-store/README.md
+++ b/examples/warp-key-value-store/README.md
@@ -4,7 +4,7 @@ This examples contains a simple key/value store with an HTTP API built using war
 
 ## Endpoints
 
-- `GET /:key` - Look up a key. If the key doesn't exist it returns `400 Not Found`
+- `GET /:key` - Look up a key. If the key doesn't exist it returns `404 Not Found`
 - `POST /:key` - Insert a key. The value is the request body.
 
 ## Running the example
--- a/examples/warp-key-value-store/src/main.rs
+++ b/examples/warp-key-value-store/src/main.rs
@@ -26,7 +26,10 @@ struct Config {
     port: u16,
 }
 
-type Database = Arc<RwLock<HashMap<String, Bytes>>>;
+#[derive(Clone, Debug)]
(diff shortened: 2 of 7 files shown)


examples/warp-key-value-store/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+#[tokio::main]
+async fn main() {
+    // Setup tracing
+    tracing_subscriber::fmt::init();
+
+    // Parse command line arguments
+    let config = Config::from_args();
+
+    // Build our service.
+    let service = make_service();
```
LucioFranco, line 42: I think we can have a better name for this function, that describes better what its doing.

examples/warp-key-value-store/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    tracing::info!("Listening on {}", addr);
+
+    Server::bind(&addr)
+        .serve(Shared::new(service))
+        .await
+        .unwrap();
+}
+
+// Function to construct our service.
```
LucioFranco, line 55: Add a comment about the return type here its quite complex so some docs here might make this easier to grok

examples/warp-key-value-store/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+fn make_service() -> impl Service<
+    Request<Body>,
+    Response = Response<CompressionBody<Body>>,
+    Error = BoxError,
+    Future = impl Future<Output = Result<Response<CompressionBody<Body>>, BoxError>> + Send,
+> + Clone {
+    // Build our database for holding the key/value pairs
+    let db: Database = Arc::new(RwLock::new(HashMap::new()));
```
LucioFranco, line 63: I would avoid the type alias and actually just make this a struct.

LucioFranco (commented):
This is looking good!

examples/warp-key-value-store/src/main.rs
davidpdrsn: I found a way to avoid having the service returned by a function. Should be simpler :)

examples/warp-key-value-store/src/main.rs
davidpdrsn: Found a way to avoid having this function which got rid of the gnarly type.

examples/warp-key-value-store/src/main.rs
davidpdrsn: Good point 👍 Fixed

examples/warp-key-value-store/README.md
```diff
@@ -0,0 +1,14 @@
+# warp-key-value-store
+
+This examples contains a simple key/value store with an HTTP API built using warp.
+
+## Endpoints
+
+- `GET /:key` - Look up a key. If the key doesn't exist it returns `400 Not Found`
```
tesaguri, line 7:
```suggestion
- `GET /:key` - Look up a key. If the key doesn't exist it returns `404 Not Found`
```

LucioFranco (approved):

PR #70, opened 2021-02-22
Add `RequireAuthorization` middleware

### LINKED ISSUE #58: Authorization middleware
A basic middleware for checking the `Authorization` header would be useful I think.

Use cases I'm thinking:

- Validating a bearer token.
- Checking username+password when doing basic auth.

davidpdrsn:
Fixes #58


LucioFranco (approved):
Do we have tests for these?

Issue #74, opened 2021-03-08
Support treating closures returning classifiers as `MakeClassifier`s


davidpdrsn:
Something like `F: Fn(Request<B>) -> ClassifyResponse<E>` should be considered a `MakeClassifier`. Could either be with a blanket impl or something like `make_classifier_fn`.

davidpdrsn:
@hawkw I gave this a shot but running into some issues.

Just trying to implement `MakeClassifier` for a closure doesn't quite work. I run into two problems:

```rust
impl<E, F, B, C> MakeClassifier<E> for F
//            ^ unconstrained type
where
    F: Fn(&Request<B>) -> C,
    C: ClassifyResponse<E>,
{
    type Classifier = C;
    type FailureClass = C::FailureClass;
    type ClassifyEos = C::ClassifyEos;

    fn make_classifier<B2>(&self, req: &Request<B2>) -> Self::Classifier
    where
        B == B2, // just some made up syntax. F takes a Request<B> not Request<B2>
    {
        self(req)
    }
}
```

It _might_ be possible to fix by making the whole trait generic over `B` instead of just the method. But I didn't explore that.

davidpdrsn:
I'm going to close this. Don't think there is a good solution.

PR #76, opened 2021-03-10
Classification combinators

davidpdrsn:
I have started integrating the response classifiers into Embark's code base. Found these combinators and utilities helpful.

New things:

- `MakeClassifier`, `ClassifyResponse`, and `ClassifyEos` implemented for `tower::util::Either`.
- `make_classifier_fn` for creating `MakeClassifier` from a function. The function is given method, uri, version, and headers. It cannot take the request as the `MakeClassifier::make_classifier` method is generic over the body type rather than the `MakeClassifier` trait itself.
- `ClassifyResponse::map_failure_class` and `ClassifyEos::map_failure_class` for changing the output type of classifiers.
- `ClassifiedResponse::map_failure_class` also for changing classification output type.
- `ClassifiedResponse::map_classify_eos` for changing the end of stream classifier.

Fixes https://github.com/tower-rs/tower-http/issues/74


davidpdrsn:
Once https://github.com/tower-rs/tower-http/pull/90 is merged I'll rewrite this. I think that will simplify things quite a bit.

PR #68, opened 2021-02-19
Make sure everything has examples

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/55

PR #79, opened 2021-03-18
Add `FollowRedirect` middleware

### LINKED ISSUE #78: Client-side redirection middleware
A middleware that handles client-side HTTP redirections would be nice, I think.

The middleware needs to rewrite the URI and request method of the request based on the response, and since `tower::retry::Policy` does not allow you to modify the request (IIUIC), we have to implement it as an independent middleware.

It would also be nice to have a mechanism akin to [`reqwest::redirect::Policy`] to control the behavior of redirection e.g. to limit the number of successive redirections and to restrict the origin of the destination.

[`reqwest::redirect::Policy`]: https://docs.rs/reqwest/0.11.2/reqwest/redirect/struct.Policy.html

If this idea looks fine to you, I'd be happy to work on a PR.

tesaguri:
This is a middleware that retries requests to follow HTTP redirections.

This includes `Policy` trait to customize behavior of the middleware, which is heavily inspired by [`reqwest::redirect::Policy`].

`Policy` also has an optional method `clone_body` that tries to clone a request body, like [`tower::retry::Policy::clone_request`]. But even if the body cannot be cloned, `FollowRedirect` uses `<ReqBody as Default>::default` when `<ReqBody as http_body::Body>::size_hint` returns zero, so that it can handle common `GET` requests without `clone_body`.

The API proposed here is a proof of concept to some extent, and it is fine if this is not fully merged at once (~2fe3fea1e3c2829cbf0474cb2b36194fd289f287...22212ba78c1f6585289abaeaa54b3bbaf43c01bd~ ~3f8deda5638a2e28c2570855e8d1d3db0d65fcb5...22212ba78c1f6585289abaeaa54b3bbaf43c01bd~ 3f8deda5638a2e28c2570855e8d1d3db0d65fcb5...0e6a949f3793c8cfd5ba5b4dafb7dbec508ee3dc is optional in particular).

Fixes #78.

[`reqwest::redirect::Policy`]: https://docs.rs/reqwest/0.11.2/reqwest/redirect/struct.Policy.html
[`tower::retry::Policy::clone_request`]: https://docs.rs/tower/0.4.6/tower/retry/trait.Policy.html#method.clone_request

### DIFF
--- a/tower-http/src/follow_redirect/mod.rs
+++ b/tower-http/src/follow_redirect/mod.rs
@@ -2,8 +2,9 @@
 
 pub mod policy;
 
-use self::policy::{ActionKind, Attempt, Policy, Standard};
+use self::policy::{Action, Attempt, Policy, Standard};
 use futures_core::ready;
+use futures_util::future::Either;
 use http::{
     header::LOCATION, HeaderMap, HeaderValue, Method, Request, Response, StatusCode, Uri, Version,
 };
@@ -21,6 +22,7 @@ use std::{
     str,
     task::{Context, Poll},
 };
+use tower::util::Oneshot;
 use tower_layer::Layer;
 use tower_service::Service;
 
@@ -68,6 +70,8 @@ impl<S> FollowRedirect<S> {
     pub fn standard(inner: S) -> Self {
         Self::new(inner, Standard::default())
     }
+
+    define_inner_service_accessors!();
 }
 
 impl<S, P> FollowRedirect<S, P>
@@ -95,14 +99,15 @@ where
 {
     type Response = Response<ResBody>;
     type Error = S::Error;
-    type Future = ResponseFuture<S::Future, S, ReqBody, P>;
+    type Future = ResponseFuture<S, ReqBody, P>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         self.inner.poll_ready(cx)
     }
 
     fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
-        let mut service = self.inner.clone();
+        let service = self.inner.clone();
+        let mut service = mem::replace(&mut self.inner, service);
         let mut policy = self.policy.clone();
         let mut body = BodyRepr::None;
         body.try_clone_from(req.body(), &policy);
@@ -113,7 +118,7 @@ where
             version: req.version(),
             headers: req.headers().clone(),
             body,
-            future: service.call(req),
+            future: Either::Left(service.call(req)),
             service,
             policy,
         }
@@ -123,9 +128,12 @@ where
 /// Response future for [`FollowRedirect`].
 #[pin_project]
 #[derive(Debug)]
-pub struct ResponseFuture<F, S, B, P> {
+pub struct ResponseFuture<S, B, P>
+where
+    S: Service<Request<B>>,
+{
     #[pin]
-    future: F,
+    future: Either<S::Future, Oneshot<S, Request<B>>>,
     service: S,
     policy: P,
     method: Method,
@@ -135,10 +143,9 @@ pub struct ResponseFuture<F, S, B, P> {
     body: BodyRepr<B>,
 }
 
-impl<F, S, ReqBody, ResBody, P> Future for ResponseFuture<F, S, ReqBody, P>
+impl<S, ReqBody, ResBody, P> Future for ResponseFuture<S, ReqBody, P>
 where
-    F: Future<Output = Result<Response<ResBody>, S::Error>>,
-    S: Service<Request<ReqBody>, Future = F> + Clone,
+    S: Service<Request<ReqBody>, Response = Response<ResBody>> + Clone,
     ReqBody: Body + Default,
     P: Policy<ReqBody, S::Error>,
 {
@@ -177,12 +184,7 @@ where
         let location = res
             .headers()
             .get(&LOCATION)
-            .and_then(|loc| str::from_utf8(loc.as_bytes()).ok())
-            .and_then(|loc| RiReferenceStr::<UriSpec>::new(loc).ok())
-            .and_then(|loc| {
-                Some(loc.resolve_against(&RiAbsoluteString::try_from(this.uri.to_string()).ok()?))
-            })
-            .and_then(|loc| Uri::try_from(loc.as_str()).ok());
+            .and_then(|loc| resolve_uri(str::from_utf8(loc.as_bytes()).ok()?, this.uri));
         let location = if let Some(loc) = location {
             loc
         } else {
@@ -194,8 +196,8 @@ where
             location: &location,
             previous: this.uri,
         };
-        match this.policy.redirect(&attempt).kind {
-            ActionKind::Follow => {
+        match this.policy.redirect(&attempt)? {
+            Action::Follow => {
                 this.body.try_clone_from(&body, &this.policy);
 
                 let mut req = Request::new(body);
@@ -204,13 +206,13 @@ where
                 *req.version_mut() = *this.version;
                 *req.headers_mut() = this.headers.clone();
                 this.policy.on_request(&mut req);
-                this.future.set(this.service.call(req));
+                this.future
+                    .set(Either::Right(Oneshot::new(this.service.clone(), req)));
 
                 cx.waker().wake_by_ref();
                 Poll::Pending
             }
-            ActionKind::Stop => Poll::Ready(Ok(res)),
-            ActionKind::Error(e) => Poll::Ready(Err(e)),
+            Action::Stop => Poll::Ready(Ok(res)),
         }
     }
 }
@@ -264,6 +266,14 @@ where
     }
 }
 
+/// Try to resolve a URI reference `relative` against a base URI `base`.
+fn resolve_uri(relative: &str, base: &Uri) -> Option<Uri> {
+    let relative = RiReferenceStr::<UriSpec>::new(relative).ok()?;
+    let base = RiAbsoluteString::try_from(base.to_string()).ok()?;
+    let uri = relative.resolve_against(&base);
+    Uri::try_from(uri.as_str()).ok()
+}
+
 #[cfg(test)]
 mod tests {
     use super::{policy::*, *};
@@ -274,7 +284,7 @@ mod tests {
     #[tokio::test]
     async fn follows() {
         let svc = ServiceBuilder::new()
-            .layer(FollowRedirectLayer::new(Action::follow()))
+            .layer(FollowRedirectLayer::new(Action::Follow))
             .service_fn(handle);
         let req = Request::builder()
             .uri("http://example.com/42")
@@ -287,7 +297,7 @@ mod tests {
     #[tokio::test]
     async fn stops() {
         let svc = ServiceBuilder::new()
-            .layer(FollowRedirectLayer::new(Action::stop()))
+            .layer(FollowRedirectLayer::new(Action::Stop))
             .service_fn(handle);
         let req = Request::builder()
             .uri("http://example.com/42")
--- a/tower-http/src/follow_redirect/mod.rs
+++ b/tower-http/src/follow_redirect/mod.rs
@@ -1,4 +1,89 @@
 //! Middleware for following redirections.
+//!
+//! # Overview
+//!
+//! The [`FollowRedirect`] middleware retries requests with the inner [`Service`] to follow HTTP
+//! redirections.
+//!
+//! The middleware tries to clone the original [`Request`] when making a redirected request.
+//! However, since [`Extensions`][http::Extensions] are `!Clone`, any extensions set by outer
+//! middlewares will be discarded. Also, the request body cannot always be cloned. When the
+//! original body is known to be empty by [`Body::size_hint`], the middleware uses `Default`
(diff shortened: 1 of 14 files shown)


tower-http/src/follow_redirect/mod.rs
```diff
@@ -0,0 +1,325 @@
+//! Middleware for following redirections.
```
davidpdrsn, line 1: A few high level examples here would be nice. I think there should be one showing the default use case using `FollowRedirectLayer::standard` and one that customizes the policy in some way.

Something like what I've done here https://github.com/tower-rs/tower-http/pull/68.

tower-http/src/follow_redirect/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    pub fn new(inner: S, policy: P) -> Self {
+        FollowRedirect { inner, policy }
+    }
+
+    /// Returns a new [`Layer`] that wraps services with a `FollowRedirect` middleware.
+    ///
+    /// [`Layer`]: tower_layer::Layer
+    pub fn layer(policy: P) -> FollowRedirectLayer<P> {
+        FollowRedirectLayer::new(policy)
+    }
```
davidpdrsn, line 87: We should have `get_ref`, `get_mut`, and `into_inner` accessors here. You can generate them with `define_inner_service_accessors!()`.

tower-http/src/follow_redirect/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    type Response = Response<ResBody>;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future, S, ReqBody, P>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
+        let mut service = self.inner.clone();
```
davidpdrsn, line 105: The clone of a ready service might not be ready. This is a workaround:

```suggestion
        let clone = self.inner.clone();
        let mut service = std::mem::replace(&mut self.inner, clone);
```

More details [here](https://github.com/tower-rs/tower/issues/547#issuecomment-767629149).

tower-http/src/follow_redirect/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        let location = res
+            .headers()
+            .get(&LOCATION)
+            .and_then(|loc| str::from_utf8(loc.as_bytes()).ok())
+            .and_then(|loc| RiReferenceStr::<UriSpec>::new(loc).ok())
+            .and_then(|loc| {
+                Some(loc.resolve_against(&RiAbsoluteString::try_from(this.uri.to_string()).ok()?))
+            })
+            .and_then(|loc| Uri::try_from(loc.as_str()).ok());
```
davidpdrsn, line 185: What does this actually do? I'm not familiar with [iri_string](https://docs.rs/iri-string/0.3.0/iri_string/index.html).

What do you think about extracting this bit into a function that returns `Option<_>` so we could use `?` without chaining `and_then` so many times?

tower-http/src/follow_redirect/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            ActionKind::Follow => {
+                this.body.try_clone_from(&body, &this.policy);
+
+                let mut req = Request::new(body);
+                *req.uri_mut() = location;
+                *req.method_mut() = this.method.clone();
+                *req.version_mut() = *this.version;
+                *req.headers_mut() = this.headers.clone();
+                this.policy.on_request(&mut req);
+                this.future.set(this.service.call(req));
```
davidpdrsn, line 207: As `call` is called on the inner service in `FollowRedirect::call` we have to wait for the service to become ready again.

One could do `this.service.clone().oneshot(req)` but that returns a different future type than what `F` is here. Maybe `futures::future::Either` could be used to combine the two types.

tower-http/src/follow_redirect/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+        let svc = ServiceBuilder::new()
+            .layer(FollowRedirectLayer::new(Limited::new(10)))
+            .service_fn(handle);
+        let req = Request::builder()
+            .uri("http://example.com/42")
+            .body(Body::empty())
+            .unwrap();
+        let res = svc.oneshot(req).await.unwrap();
+        assert_eq!(res.into_body(), 42 - 10);
+    }
```
davidpdrsn, line 311: I'm thinking a test where the inner service is wrapper in a `tower::buffer::Buffer` and does a couple of redirects would be cool. `Buffer` should highlight any mistakes in how calls to `poll_ready`/`call` are managed.

tower-http/src/follow_redirect/policy/filter_credentials.rs
```diff
@@ -0,0 +1,156 @@
...
+        self.block_any = false;
+        self.block_cross_origin(false)
+    }
+
+    /// Configure `self` to remove blocklisted headers in "blocked" redirections.
+    ///
+    /// The blocklist includes the following headers:
+    ///
+    /// - `Authorization`
+    /// - `Cookie`
```
davidpdrsn, line 59: Should this list mention `Proxy-Authorization` as well?

tower-http/src/follow_redirect/policy/filter_credentials.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn remove_blocklisted(mut self, enable: bool) -> Self {
+        self.remove_blocklisted = enable;
+        self
+    }
+
+    /// Configure `self` to remove all headers in "blocked" redirections.
+    pub fn remove_any(mut self) -> Self {
```
davidpdrsn, line 66: What do you think about naming this `remove_all`? Thinking thats more clear.

tower-http/src/follow_redirect/policy/join.rs
```diff
@@ -0,0 +1,139 @@
...
+///     }),
+/// );
+/// ```
+pub fn join<A, B, Bd, E>(a: A, b: B) -> Join<A, B>
+where
+    A: Policy<Bd, E>,
+    B: Policy<Bd, E>,
+{
+    Join { a, b }
+}
```
davidpdrsn, line 72: I think this is cool but wondering if it should be method directly on `Policy` and perhaps be called `and`? That way you can do `policy_one.and(policy_two)`. I got confused initially by the name `join` as I was thinking of `futures::join!`.

tower-http/src/follow_redirect/policy/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+/// ```
+pub trait Policy<B, E> {
+    /// Invoked when the service received a response with a redirection status code (`3xx`).
+    ///
+    /// This method returns an [`Action`] which indicates whether the service should follow
+    /// the redirection.
+    fn redirect(&mut self, attempt: &Attempt<'_>) -> Action<E>;
+
+    /// Invoked right before the service makes a request, regardless of whether it is redirected
+    /// or not.
```
davidpdrsn, line 59: Think we should have a little note about why this method is useful:

```suggestion
    /// or not.
    ///
    /// This can for example be used to remove sensitive headers from the request
    /// or prepare the request in other ways.
    ///
    /// The default implementation returns does nothing.
```

tower-http/src/follow_redirect/policy/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+pub struct Action<E> {
+    pub(crate) kind: ActionKind<E>,
+}
+
+#[derive(Clone)]
+pub(crate) enum ActionKind<E> {
+    Follow,
+    Stop,
+    Error(E),
+}
```
davidpdrsn, line 151: I not sure, but wondering if the error case is necessary to have on `Action` or if using something like `Result<Action, E>` would work just as well.

If we do this it might also be worth making the `ActionKind` enum into `Action` and expose that:

```rust
pub enum Action {
    Follow,
    Stop,
}
```

We could still have `is_follow` and `is_stop` convenience methods.

It would make adding variants a breaking change but not sure what we might want to add. Either you follow a redirect or you don't 🤷

tower-http/src/follow_redirect/policy/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        let mut debug = f.debug_tuple("Action");
+        match self.kind {
+            ActionKind::Follow => debug.field(&Follow),
+            ActionKind::Stop => debug.field(&Stop),
+            ActionKind::Error(ref e) => debug.field(&Error(e)),
+        };
+        debug.finish()
+    }
+}
```
davidpdrsn, line 219: I think having `#[derive(Debug)]` on `Action` and `ActionKind` is fine.

tower-http/src/follow_redirect/policy/select.rs
```diff
@@ -0,0 +1,130 @@
...
+/// #[derive(Clone)]
+/// enum MyError {
+///     TooManyRedirects,
+///     // ...
+/// }
+///
+/// let policy =
+///     policy::select::<_, _, (), _>(Limited::default(), Action::error(MyError::TooManyRedirects));
+/// ```
+pub fn select<A, B, Bd, E>(a: A, b: B) -> Select<A, B>
```
davidpdrsn, line 57: Similarly to `join` I'm thinking if this should be a method on `Policy` and perhaps renamed to `or`.

It wasn't first clear to me what "select" meant. I was kinda thinking of `futures::select!`.

davidpdrsn (changes requested):
Really great work on this!

I left some comments but overall I think this setup is nice.

I would also like to hear what @hawkw thinks. Adding her as reviewer.

tower-http/src/follow_redirect/policy/join.rs
tesaguri: > wondering if it should be method directly on Policy and perhaps be called and?

That was exactly the same design I came up with initially in commit 2fe3fea1e3c2829cbf0474cb2b36194fd289f287, but after I added the error type parameter to `Policy` in commit 04cc691e034342be605242b33f0ff976e8316d2a, the `and` method began to break the type inference:

```
  --> src/follow_redirect/policy/mod.rs:90:33
   |
13 | let policy = Limited::default().and(clone_body_fn(|body| {
   |                                 ^^^ cannot infer type for type parameter `E` declared on the trait `Policy`
```

To annotate the type parameter, the caller needs to use the UFCS like `Policy::<_, ()>::and(Limited::default(), clone_body_fn(/* ... */))`, and I thought that a freestanding function would be more ergonomic than UFCS. But I agree that this is still an awkward interface. I will explore other possibilities (maybe an extension trait would work?).

tower-http/src/follow_redirect/policy/mod.rs
tesaguri: Just for the record, I have considered of encoding more information in `Action` or `Attempt`, like whether `FollowRedirect` should remove credentials. But now that `FilterCredentials` exists, there is no need to hide the `ActionKind` variants, indeed.

We could make `Action` `#[non_exhaustive]`, but I neither think that we would benefit from it.

tower-http/src/follow_redirect/mod.rs
tesaguri: This is resolving the URI reference of the `Location` header against the request URI ([RFC 7231 section 7.1.2](https://tools.ietf.org/html/rfc7231#section-7.1.2)). I added an explanation to the extracted function.

tower-http/src/follow_redirect/policy/join.rs
tesaguri: I have just added an extension trait `PolicyExt` with `and` and `or` methods and this seems to work fine.

However, I had to implement it for arbitrary`T` without `T: Policy<B, E>` bound to avoid having type parameters in the extension trait, and the methods conflict with methods from other traits named `and` or `or` (I confirmed that they do not conflict with inherent methods like `Option::<_>::and`). I have no idea whether this would cause any issues in practice.

tower-http/src/follow_redirect/mod.rs
tesaguri: Added in 35e66afc40078576fe27072e3f58742b3588889a. It seems to be able to catch mistakes in both `FollowRedirect::call` and `ResponseFuture::poll`.

tower-http/src/follow_redirect/mod.rs
tesaguri: Added in ac8818b02904d0055f1fd2a8c0013c7e5c7e7f80.

davidpdrsn: Awesome!

davidpdrsn (commented):
I think it looks good! I'm gonna try and integrate it with our code at embark as the final test.

davidpdrsn (approved):
Its working quite well! I ended up with something like this:

```rust
// automatically follow redirects
.option_layer(self.max_redirects.map(|limit| {
    FollowRedirectLayer::with_policy(
        Limited::new(limit).and::<_, _, HttpError>(clone_body_fn(Body::try_clone)),
    )
}))
```

lpil:
Hey really cool!

davidpdrsn:
Think I'm just going to merge this. We can always improve things later.

Issue #81, opened 2021-03-21
Compression and Decompression breaks tonic streams


davidpdrsn:
I've noticed using the compression and decompression middlewares makes tonic streaming messages not get delivered for some reason.

Noticed while adding a stream example to the tonic example.

davidpdrsn:
I have spent all day trying to fix this but I cannot find anything 😞 

I have made a [minimal reproducible example](https://github.com/davidpdrsn/minimal-compression-example) in case someone wants to have a go.

I first thought it was due to `is_end_stream` not getting delegated correctly, but that doesn't seem to be the case.

Also, if you compress/decompress things directly in `poll_data` with [flate2](https://docs.rs/flate2/1.0.20/flate2) it does work. Could async-compression be splitting messages in a way that isn't compatible with grpc?

@tesaguri do you have any idea what could be causing this?

seanmonstar:
Doesn't protobuf place compression info in the protobuf binary format? I think there's a flag when decoding that signals that a message is compressed.

davidpdrsn:
Yeah I think it does. Seen mentions of it in the tonic code.

Tower-http's middleware work at the http_body::Body layer so in theory compression should be applied and unapplied transparently before responses reaches the grpc client service. But seems not 😞

LucioFranco:
@davidpdrsn do you think we can implement a `Body` that does it DATA frame by DATA frame instead of over the whole body?

We could probably read the headers than when we create the Body middleware to decompress we can teach it how to read grpc? but maybe this is getting to the point where it makes sense to add this into tonic itself.

davidpdrsn:
> do you think we can implement a Body that does it DATA frame by DATA frame instead of over the whole body?

@LucioFranco Yep totally. I actually did that during my testing and that does work. You can even do it with something like [flate2](https://crates.io/crates/flate2) as you don't need async compression.

That is a direction I want to explore in the future but probably not necessary for the first version of this. Might also require some additional like for things like "should users be able to choose between whole body or single frame compression regardless of gRPC or not" kinda questions.

PR #82, opened 2021-03-21
Disable compression for gRPC and add tonic streaming example

davidpdrsn:
This adds a server streaming example to the tonic example. Adding that made me aware of https://github.com/tower-rs/tower-http/issues/81.

After much debugging I most admit I don't fully understand why compression breaks gRPC streams. My best idea is that async-compression changes the body chunks in ways that are incompatible with gRPC. This is also hinted at in the [gRPC core document on compression](https://grpc.github.io/grpc/core/md_doc_compression.html) where it says:

> The compression supported by gRPC acts at the individual message level

If you don't compress the whole body stream but instead compress (and decompress) each individual chunk yielded by `poll_data` then everything works.

The solution I have implemented is to simply detect gRPC responses and don't compress them, even if the client claims to accept compression. As compression part of the gRPC protocol that is probably a better place to implement it.

Fixes https://github.com/tower-rs/tower-http/issues/81

### DIFF
--- a/examples/tonic-key-value-store/build.rs
+++ b/examples/tonic-key-value-store/build.rs
@@ -1,3 +1,6 @@
 fn main() {
-    tonic_build::compile_protos("proto/key_value_store.proto").unwrap();
+    tonic_build::configure()
+        .format(false)
+        .compile(&["key_value_store.proto"], &["proto"])
+        .unwrap();
 }
--- a/examples/tonic-key-value-store/src/main.rs
+++ b/examples/tonic-key-value-store/src/main.rs
@@ -19,7 +19,7 @@ use std::{
 };
 use structopt::StructOpt;
 use tokio::io::AsyncReadExt;
-use tokio::sync::broadcast::{channel, Sender};
+use tokio::sync::broadcast::{self, Sender};
 use tokio_stream::{wrappers::BroadcastStream, Stream};
 use tonic::{async_trait, body::BoxBody, transport::Channel, Code, Request, Response, Status};
 use tower::{make::Shared, ServiceBuilder};
@@ -153,7 +153,7 @@ async fn serve_forever(listener: TcpListener) -> Result<(), Box<dyn std::error::
     // Build our database for holding the key/value pairs
     let db = Arc::new(RwLock::new(HashMap::new()));
 
-    let (tx, _rx) = channel(1024);
+    let (tx, _rx) = broadcast::channel(1024);
 
     // Build our tonic `Service`
     let service = key_value_store_server::KeyValueStoreServer::new(ServerImpl { db, tx });
--- a/examples/tonic-key-value-store/src/main.rs
+++ b/examples/tonic-key-value-store/src/main.rs
@@ -241,7 +241,7 @@ impl key_value_store_server::KeyValueStore for ServerImpl {
                 item.ok()
             })
             .map(Ok);
-        let stream: Self::SubscribeStream = Box::pin(stream);
+        let stream = Box::pin(stream) as Self::SubscribeStream;
         let res = Response::new(stream);
 
         Ok(res)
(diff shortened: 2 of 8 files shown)


examples/tonic-key-value-store/build.rs
```diff
@@ -1,3 +1,6 @@
 fn main() {
-    tonic_build::compile_protos("proto/key_value_store.proto").unwrap();
+    tonic_build::configure()
+        .format(false)
```
LucioFranco, line 3: why this?

examples/tonic-key-value-store/src/main.rs
```diff
@@ -128,8 +153,10 @@ async fn serve_forever(listener: TcpListener) -> Result<(), Box<dyn std::error::
     // Build our database for holding the key/value pairs
     let db = Arc::new(RwLock::new(HashMap::new()));
 
+    let (tx, _rx) = channel(1024);
```
LucioFranco, line 156: minor nit I would prob keep this as `broadcast::channel` or `mpsc::channel` to make it clear which flavor it is.

examples/tonic-key-value-store/src/main.rs
```diff
@@ -188,10 +216,36 @@ impl key_value_store_server::KeyValueStore for ServerImpl {
...
     }
+
+    type SubscribeStream =
+        Pin<Box<dyn Stream<Item = Result<SubscribeReply, Status>> + Send + Sync + 'static>>;
+
+    async fn subscribe(
+        &self,
+        request: Request<SubscribeRequest>,
+    ) -> Result<Response<Self::SubscribeStream>, Status> {
+        let SubscribeRequest {} = request.into_inner();
```
LucioFranco, line 235: What is this syntax? Are you matching on an empty struct huh

examples/tonic-key-value-store/src/main.rs
```diff
@@ -188,10 +216,36 @@ impl key_value_store_server::KeyValueStore for ServerImpl {
...
+
+        let rx = self.tx.subscribe();
+        let stream = BroadcastStream::new(rx)
+            .filter_map(|item| async move {
+                // ignore receive errors
+                item.ok()
+            })
+            .map(Ok);
+        let stream: Self::SubscribeStream = Box::pin(stream);
```
LucioFranco, line 244:
```suggestion
        let stream = Box::pin(stream) as Self::SubscribeStream;
```

nit: I tend to prefer this style.

tower-http/src/compression_utils.rs
```diff
@@ -308,3 +308,27 @@ where
...
+
+#[allow(clippy::clippy::needless_bool)]
+pub(crate) fn supports_transparent_compression(headers: &HeaderMap) -> bool {
+    let content_type = if let Some(content_type) = content_type(headers) {
+        content_type
+    } else {
+        return true;
+    };
+
+    if content_type == "application/grpc" {
```
LucioFranco, line 320: probably want to do `starts_with` since sometimes grpc impls will tack on `+proto`/`+json`

examples/tonic-key-value-store/build.rs
davidpdrsn: We were getting some CI errors about rustfmt not being installed when publishing the docs. This is an easy fix for that. Don't publish docs for examples regardless.

examples/tonic-key-value-store/src/main.rs
davidpdrsn: Mostly to get a compile error if/when we add more fields. Not strictly necessary for this example but might help users who copy+paste this.

tower-http/src/compression_utils.rs
davidpdrsn: Good catch!

LucioFranco (approved):

Issue #65, opened 2021-02-13
Tracing middleware


davidpdrsn:
We should provide a middleware that adds high level tracing to services. Things like attaching spans to futures produced by services, and events when a request is received and completed.

This currently blocked by the response classification stuff we're working on in https://github.com/tower-rs/tower-http/pull/49 as the middleware needs to know if a response is a success or failure.

Some initial work was done in https://github.com/tower-rs/tower-http/pull/40 but that was scrapped because finishing both response classification and tracing in one PR was too much.

PR #84, opened 2021-04-03
Use better failure class for HTTP and gRPC classifiers

davidpdrsn:
Previously `<ServerErrorsAsFailures as ClassifyResponse<E>>::classify_error` would return `StatusCode::INTERNAL_SERVER_ERROR`.

That is because `classify_error` receives an opaque `&E` so it couldn't do much else without adding more requirements or returning some kind of "an error happened, but I know nothing about it" value.

However creating a `StatusCode::INTERNAL_SERVER_ERROR` out of the blue is very ad-hoc especially since a response was never produced so there is no status code to speak of.

When I implemented the trace middleware in #65 I gave this some more thought. I'm currently thinking requiring `E: fmt::Display` for `ServerErrorsAsFailures` and `GrpcErrorsAsFailures` and then returning something that contains the error's string representation. Its not ideal because we loose the error chain but I'm thinking users who can to customize things further can make their own classifier. We could perhaps require `E: std::error::Error` instead but I'm not sure that wouldn't break some setups. In tower we mainly use `Into<BoxError>` instead of `std::error::Error`.

Note that this only changes the built-in classifiers. It classification traits themselves are not changed.

### DIFF
--- a/examples/tonic-key-value-store/src/main.rs
+++ b/examples/tonic-key-value-store/src/main.rs
@@ -27,6 +27,7 @@ use tower::{BoxError, Service};
 use tower_http::{
     compression::CompressionLayer,
     decompression::DecompressionLayer,
+    map_response_body::MapResponseBodyLayer,
     sensitive_header::SetSensitiveHeaderLayer,
     set_header::SetRequestHeaderLayer,
     trace::{DefaultMakeSpan, TraceLayer},
@@ -287,11 +288,23 @@ async fn make_client(
         .layer(
             TraceLayer::new_for_grpc().make_span_with(DefaultMakeSpan::new().include_headers(true)),
         )
+        .layer(MapResponseBodyLayer::new(|body: hyper::Body| {
+            body.map_err(TonicOrHyperError::from)
+        }))
+        .map_err(TonicOrHyperError::from)
         // Build our final `Service`
         .service(channel);
 
     // Construct our tonic client
     Ok(KeyValueStoreClient::new(channel))
+}
+
+#[derive(Debug, thiserror::Error)]
+enum TonicOrHyperError {
+    #[error(transparent)]
+    Tonic(#[from] tonic::transport::Error),
+    #[error(transparent)]
+    Hyper(#[from] hyper::Error),
 }
 
 #[cfg(test)]
--- a/examples/tonic-key-value-store/Cargo.toml
+++ b/examples/tonic-key-value-store/Cargo.toml
@@ -8,14 +8,15 @@ license = "MIT"
 
 [dependencies]
 bytes = "1"
+futures = "0.3"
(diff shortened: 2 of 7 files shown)


davidpdrsn:
I think I've found a nicer solution. Instead of always requiring the error type to be `fmt::Display` you can take a closure that classifies the error. Thus defining `ServerErrorsAsFailures` as

```rust
pub struct ServerErrorsAsFailures<E, F = fn(&E) -> String> {
    map_error: F,
    _error: PhantomData<fn() -> E>,
}
```

I'm writing the code now. Will push soon.

davidpdrsn:
Gah so that _almost_ works. A problem you run into is that Tonic clients (`Channel`) uses `tonic::transport::Error` as the `Service::Error` but `hyper::Body` as the response body which means the response body's error type doesn't match the service error type.

I've pushed the changes and will look into it later.

davidpdrsn:
Its possible to fix for the client service by mapping the service error and body errors into the same type. Had to make a separate enum because `BoxError` leads to odd compile errors.

```rust
#[derive(Debug, thiserror::Error)]
enum TonicOrHyperError {
    #[error(transparent)]
    Tonic(#[from] tonic::transport::Error),
    #[error(transparent)]
    Hyper(#[from] hyper::Error),
}

let channel = Channel::builder(uri).connect().await?;

let channel = ServiceBuilder::new()
    .layer(
        TraceLayer::new_for_grpc().make_span_with(DefaultMakeSpan::new().include_headers(true)),
    )
    .layer(MapResponseBodyLayer::new(|body: hyper::Body| {
        body.map_err(TonicOrHyperError::from)
    }))
    .map_err(TonicOrHyperError::from)
    .service(channel);
```

But of course you run into the same problem with the server service. There the service error type is `tonic::codegen::Never` and the response body type is `Status`.

Having to add boilerplate just to merge some types is really quite unfortunate. I guess one could change it in tonic but that would be a breaking change and what do we next time another lib comes around? Having to do this in the first place is unfortunate...

I guess either we require `E: fmt::Display` for `ServerErrorsAsFailures` and `GrpcErrorsAsFailures` and always convert the error into a `String` (and loose type information) or ignore the error all together 🤷

examples/tonic-key-value-store/src/main.rs
```diff
@@ -287,13 +288,25 @@ async fn make_client(
...
     Ok(KeyValueStoreClient::new(channel))
 }
 
+#[derive(Debug, thiserror::Error)]
+enum TonicOrHyperError {
+    #[error(transparent)]
+    Tonic(#[from] tonic::transport::Error),
+    #[error(transparent)]
+    Hyper(#[from] hyper::Error),
+}
```
davidpdrsn, line 308: Not having to do this would be really great 🤔

PR #85, opened 2021-04-07
More easily mark multiple headers as sensitive

davidpdrsn:
Today I had to mark a list of headers as sensitive. That was quite tedious and required applying the middleware multiple times with different headers. I think this case is quite common so I think it makes more sense to construct these middlewares with an `IntoIterator<Item = HeaderName>`.

Users who just want to mark one header can use `std::iter::once(header)`.


tower-http/src/sensitive_header.rs
```diff
@@ -155,78 +177,93 @@ where
     }
 
     fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
-        if let Some(value) = req.headers_mut().get_mut(&self.header) {
-            value.set_sensitive(true);
+        for header in self.headers.clone() {
```
hawkw, line 180: I think this clone is unnecessary and could be avoided --- we only need an `&HeaderName` to call `headers_mut().get_mut()`. It would be nice to not clone header names, which may be strings, on every call.

What if we changed the
```rust
where
    I: IntoIterator<Item = HeaderName> + Clone
```
bound to 
```rust
where
    for<'a> &'a I: IntoIterator<Item = &'a HeaderName>
```

That way, we could just write
```rust
for header in &self.headers {
```
here, and avoid having to clone all the headers (and the data structure they're stored in, which might be a `Vec` or similar) in every `call`.

tower-http/src/sensitive_header.rs
```diff
@@ -47,103 +50,122 @@ use std::{
...
 ///
 /// Produces [`SetSensitiveHeader`] services.
 ///
 /// See the [module docs](crate::sensitive_header) for more details.
 ///
 /// [sensitive]: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.set_sensitive
 #[derive(Clone, Debug)]
-pub struct SetSensitiveHeaderLayer {
-    header: HeaderName,
+pub struct SetSensitiveHeaderLayer<I> {
```
hawkw, line 61: take it or leave it: maybe worth renaming this to `SetSensitiveHeadersLayer` (and change all the other types correspondingly) now that it can set multiple headers?

tower-http/src/sensitive_header.rs
```diff
@@ -236,32 +273,36 @@ where
     fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
         SetSensitiveResponseHeaderResponseFuture {
             future: self.inner.call(req),
-            header: self.header.clone(),
+            headers: Some(self.headers.clone()),
```
hawkw, line 276: For response headers, I am not sure if there is as good a way to avoid cloning the headers, like we can for the request headers, since they have to be passed into the response future in any case.

However, if we used the 
```rust
    for<'a> &'a I: IntoIterator<Item = &'a HeaderName>
```
as I discussed above, it would at least permit using an `Arc<Vec<HeaderName>>` in this case, which would make this clone much more lightweight than cloning an entire `Vec` into every response future. In fact, I think we _might_ want to remove the `Clone` bound on `I` and just always `Arc` it, for the response headers case?

That might add an unnecessary `Arc` in the case where `I` is `Option<HeaderName>` and the `HeaderName` was constructed with `HeaderName::from_static` or similar...but that might be worthwhile if it makes it impossible to accidentally write code that clones a large `Vec<HeaderName>` into every response future...

hawkw (dismissed):
It would be really nice to avoid cloning the list of header names here. I had some suggestions on how we might do this.

tower-http/src/sensitive_header.rs
davidpdrsn: I can't quite get this to work. After updating the bounds and applying the layer like so

```rust
.layer(SetSensitiveHeaderLayer::new(once(header::AUTHORIZATION)))
```

I get this error

```
 1  error[E0277]: `&'a std::iter::Once<hyper::header::HeaderName>` is not an iterator
    --> examples/tonic-key-value-store/src/main.rs:139:16
     |
 139 |         .layer(SetSensitiveHeaderLayer::new(once(header::AUTHORIZATION)))
     |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `&'a std::iter::Once<hyper::header::HeaderName>` is not an iterator
     |
    ::: /Users/david.pedersen/dev/major/tower-http/tower-http/src/sensitive_header.rs:69:37
     |
 69  |         for<'a> &'a I: IntoIterator<Item = &'a HeaderName>,
     |                                     --------------------- required by this bound in `tower_http::sensitive_header::SetSensitiveHeaderLayer::<I>::new`
     |
     = help: the trait `std::iter::Iterator` is not implemented for `&'a std::iter::Once<hyper::header::HeaderName>`
     = note: `std::iter::Iterator` is implemented for `&'a mut std::iter::Once<hyper::header::HeaderName>`, but not for `&'a std::iter::Once<hyper::header::HeaderName>`
     = note: required because of the requirements on the impl of `for<'a> std::iter::IntoIterator` for `&'a std::iter::Once<hyper::header::HeaderName>`
```

Not sure if thats fixable. Do you know? I've pushed the code in case you wanna play around.

Also, looking at the `HeaderName` implementation it seems to be backed by either a `Copy` constant or a `ByteStr` which is cheap to clone (thanks to `Bytes`). If the iterator is `vec![...].into_iter()` that isn't cheap but I've been using the new `std::array::IntoIter::new([...])` which should avoid allocations from a `Vec`.

tower-http/src/sensitive_header.rs
davidpdrsn: Good idea! I'll fix that when the clones have been dealt with 😊

tower-http/src/sensitive_header.rs
hawkw: > Not sure if thats fixable. Do you know? I've pushed the code in case you wanna play around.

Hmm, I think it's specifically `std::iter::once` that's the issue here. An array with one element ought to work, [as should `Option`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#impl-IntoIterator-1).

> Also, looking at the `HeaderName` implementation it seems to be backed by either a `Copy` constant or a `ByteStr` which is cheap to clone (thanks to `Bytes`). If the iterator is `vec![...].into_iter()` that isn't cheap but I've been using the new `std::array::IntoIter::new([...])` which should avoid allocations from a `Vec`

Yeah, the issue here IMO is not cloning the header names themselves, but the type they're stored in. My concern specifically is that because this function is generic and you can pass in anything, users will easily construct `SetSensitiveHeaderLayer`s with a `Vec<HeaderName>`, which results in the `Vec` itself getting cloned a bunch.

Another option is to change the `IntoIterator` bound to `AsRef<[&HeaderName]>`. Then, we could just call `as_ref()` rather than cloning, and iterate over the slice. This should work fine.

The downside of _that_ approach is that it doesn't allow you to pass in a `HashSet` or something (though I'm not immediately sure why you would do that instead of just using a slice/`Vec`?) and it wouldn't work with `std::iter::once` _or_ with an `Option` --- a single header would _have_ to be a one-element slice...

davidpdrsn:
@hawkw You suggestion of using `Arc<Vec<HeaderName>>` works quite well. I also updated the docs with an example of how to mark both request and response headers as sensitive when logging is involved.

Ready for another review.

tower-http/src/sensitive_headers.rs
```diff
@@ -0,0 +1,376 @@
+//! Middlewares that mark headers as [sensitive].
```
LucioFranco, line 1: Would it maybe make sense to break out the request/response ones into their own module? Might make it a bit easier to read.

LucioFranco (approved):
LGTM, left a few suggestions

PR #77, opened 2021-03-10
`Trace` middleware

davidpdrsn:
This adds a quite comprehensive `Trace` middleware. We have used it for some time at Embark for both servers and clients and its been working well.

I've spent put a lot of effort into making it easy to get started with while also providing deep customization.

I'm sorry for the size of this PR but most of it is plumbing like builder methods. The actual service/future/body implementation isn't that big.

Basic usage:

```rust
let service = ServiceBuilder::new()
    .layer(TraceLayer::new_for_http())
    .service_fn(handle);
```

Output with default tracing subscriber:

```
Mar 05 20:50:28.523 DEBUG request{method=GET path="/foo"}: tower_http::trace::on_request: started processing request
Mar 05 20:50:28.524 DEBUG request{method=GET path="/foo"}: tower_http::trace::on_response: finished processing request latency=1 ms status=200
```

Using a gRPC response classifier:

```rust
let service = ServiceBuilder::new()
    .layer(TraceLayer::new_for_grpc())
    .service_fn(handle);
```

Custom classifiers are of course also supported.

Every part of the output can be customized:

```rust
let service = ServiceBuilder::new()
    .layer(
        TraceLayer::new_for_http()
            .make_span_with(|request: &Request<Body>| {
                tracing::debug_span!("http-request")
            })
            .on_request(|request: &Request<Body>| {
                tracing::debug!("started {} {}", request.method(), request.uri().path())
            })
            .on_response(|response: &Response<Body>, latency: Duration| {
                tracing::debug!("response generated in {:?}", latency)
            })
            .on_body_chunk(|chunk: &Bytes, latency: Duration| {
                tracing::debug!("sending {} bytes", chunk.len())
            })
            .on_eos(|trailers: Option<&HeaderMap>, stream_duration: Duration| {
                tracing::debug!("stream closed after {:?}", stream_duration)
            })
            .on_failure(|error: StatusCode, latency: Duration| {
                tracing::debug!("something went wrong")
            })
    )
    .service_fn(handle);
```

You can probably imagine this leads to `Trace` having quite a few generic types 😕

If you only want to customize a few things the default behaviours have builder methods:

```rust
let service = ServiceBuilder::new()
    .layer(
        TraceLayer::new_for_http()
            .make_span_with(
                DefaultMakeSpan::new().include_headers(true)
            )
            .on_response(
                DefaultOnResponse::new()
                    .level(Level::INFO)
                    .latency_unit(LatencyUnit::Micros)
            )
            // ...
    )
[...]

Issue #57, opened 2021-02-12
Metrics middleware


davidpdrsn:
A middleware that provides high level metrics would be really cool.

Should at least cover:
- Traffic, requests per second/minute
- Number of in flight requests
- Error rate (this requires the response classification stuff thats being worked [here](https://github.com/tower-rs/tower-http/pull/49))
- Throughput, bytes per second/minute

At Embark we use prometheus and the [metrics](https://crates.io/crates/metrics) crate. But tower-http should probably allow you to use whatever metrics setup you want.

PR #86, opened 2021-05-04
Add `InFlightRequests` - The first metrics middleware

davidpdrsn:
Part of https://github.com/tower-rs/tower-http/issues/57

I think the two killer features of tower-http is going to be easy setup of tracing/logging and metrics. Those are non-trivial pieces of functionality that pretty much everyone needs. Now that the [tracing middleware is in](https://github.com/tower-rs/tower-http/pull/77) I would like to start thinking about metrics.

My initial goal is to cover these metrics:

- Traffic, requests per second/minute
- Number of in flight requests
- Error rate
- Throughput, bytes per second/minute

Each of those will be a separate middleware, starting with in-flight requests as I think its the simplest.

The module structure I'm thinking is:

- `tower_http::metrics` all middlewares related to metrics are here
- The `metrics` module exposes the most important pieces, so `tower_http::metrics::{InFlightRequests, InFlightRequestsLayer}` and so on when we have more middlewares
- Response futures, bodies and other lower level types are in `tower_http::metrics::in_flight_requests`.

This is similar to how `std::collections` do it.

Example usage of `InFlightRequests`:

```rust
use tower::{Service, ServiceExt, ServiceBuilder};
use tower_http::metrics::InFlightRequestsLayer;
use http::{Request, Response};
use hyper::Body;
use std::{time::Duration, convert::Infallible};

async fn handle(req: Request<Body>) -> Result<Response<Body>, Infallible> {
    // ...
}

async fn update_in_flight_requests_metric(count: usize) {
    // ...
}

// Create a `Layer` with an associated counter.
let (in_flight_requests_layer, counter) = InFlightRequestsLayer::pair();

// Spawn a task that will receive the number of in-flight requests every 10 seconds.
tokio::spawn(
    counter.run_emitter(Duration::from_secs(10), |count| async move {
        update_in_flight_requests_metric(count).await;
    }),
);

let mut service = ServiceBuilder::new()
    // Keep track of the number of in-flight requests. This will increment and decrement
    // `counter` automatically.
    .layer(in_flight_requests_layer)
    .service_fn(handle);

// Call the service.
let response = service
    .ready()
    .await?
    .call(Request::new(Body::empty()))
    .await?;
```

I'm also considering making a `tower-http-metrics` crate that provides integration between `tower-http` and the excellent [`metrics`](https://crates.io/crates/metrics) which we use at Embark.

### DIFF
--- a/tower-http/src/metrics/in_flight_requests.rs
+++ b/tower-http/src/metrics/in_flight_requests.rs
@@ -2,8 +2,7 @@
 //!
 //! In-flight requests is the number of requests a service is currently processing. The processing
 //! of a request starts when it is received by the service (`tower::Service::call` is called) and
-//! is considered complete when the entire response body as been generated
-//! (`http_body::Body::poll_{data,trailers}` completes) or when an error occurs along the way.
+//! is considered complete when the response body is consumed, dropped, or an error happens.
 //!
 //! # Example
 //!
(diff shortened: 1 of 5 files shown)


tower-http/src/metrics/in_flight_requests.rs
```diff
@@ -0,0 +1,283 @@
...
+where
+    F: Future<Output = Result<Response<B>, E>>,
+{
+    type Output = Result<Response<ResponseBody<B>>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let response = ready!(this.inner.poll(cx))?;
+        let guard = this.guard.take().unwrap();
+        let response = response.map(move |body| ResponseBody { inner: body, guard });
```
tobz, line 247: The way I'm reading it is that while the "in flight" aspect of a request only starts when it hits this layer -- which is fine, and obviously, you can't know it any earlier -- we actually extend the notion of how long a request is in-flight to the point where the response itself is dropped/consumed.

Personally, this feels wrong to me.  If this was middleware on a client calling a remote service, is the "request" really in flight still if we've received a response but haven't done something with it yet?

davidpdrsn: Good question. I would say yes, if the connecting still open and the body is being read then the request is still in flight. If we've only received the headers but no body then I suppose the request isn't complete. That also applies to streams which might run for a long time.

tobz: Chatting in Discord, we discussed that based on how a body will typically be used, this approach should be representative of capturing the full lifecycle of a request as it pertains to sending _and_ fully receiving the response, but no further than that point.

tobz (approved):

PR #90, opened 2021-05-11
Properly classify errors for built-in classifiers

davidpdrsn:
This supersedes #84 which didn't work and was way too complicated.

The problem that this addresses is that previously `<ServerErrorsAsFailures as ClassifyResponse>::classify_error` would return `StatusCode::INTERNAL_SERVER_ERROR`. Instead the error is now formatted using `fmt::Display` so as little info is lost as possible.

As this is a breaking change its something we have to merge before shipping 0.1.

## Why did it return `StatusCode::INTERNAL_SERVER_ERROR` in the first place?

On `master` the `ClassifyResponse` trait is defined as:

```rust
pub trait ClassifyResponse<E> {
    type FailureClass;
    type ClassifyEos: ClassifyEos<E, FailureClass = Self::FailureClass>;

    fn classify_response<B>(
        self,
        res: &Response<B>,
    ) -> ClassifiedResponse<Self::FailureClass, Self::ClassifyEos>;

    fn classify_error(self, error: &E) -> Self::FailureClass;
}
```

Since `E` was a generic type, `ServerErrorsAsFailures` couldn't do anything with the `error: &E` it got in `classify_error`. It has to either add more trait bounds to `E` or return some kind of sentinel value.

Adding bounds doesn't work because Tonic doesn't use the same error type across its `Service` and `Body` implementations. So `E` would be bound to `Service::Error` is and then type error when it got to `Body::Error`. There were no issues with the warp example which I assume is because [it uses `Infallible` as the error type](https://github.com/seanmonstar/warp/blob/b6d1fc0719604ef1010aec00544408e6af1289a5/src/filter/service.rs#L93) which got coerced to `hyper::Error` in the `Body`, I suppose 🤷 

Having `E` bound on the trait also meant that anything that contained a `ClassifyResponse` would have to be generic over `E` as well. That lead to a lot boilerplate in the tracing middleware.

So therefore `ServerErrorsAsFailures` wouldn't put any bounds on `E` and instead return a sentinel value.

## The fix

This changes the `ClassifyResponse` to instead bind `E` on the `classify_error` method with a `fmt::Display + 'static` bound. So the definition is now:

```rust
pub trait ClassifyResponse {
    type FailureClass;
    type ClassifyEos: ClassifyEos<FailureClass = Self::FailureClass>;

    fn classify_response<B>(
        self,
        res: &Response<B>,
    ) -> ClassifiedResponse<Self::FailureClass, Self::ClassifyEos>;

    fn classify_error<E>(self, error: &E) -> Self::FailureClass
    where
        E: fmt::Display + 'static;
}
```

We cannot use `E: std::fmt::Error` since `Box<dyn Error>` doesn't implement `Error` so `fmt::Display` is the second best option. `Box<dyn Error>` does implement `fmt::Display`. It does mean we loose the ability to format the full error chain but I don't see a way to resolve that.

`'static` is added such that `E` implicitly will implement `Any` to allow for downcasting. That can be done with `(error as &dyn Any).downcast_ref::<SomeOtherError>()`.

The failure class of `ServerErrorsAsFailures` now becomes either a status code or an `fmt::Display` formatted error:

```rust
/// Result of doing a classification.
#[derive(Debug)]
pub enum ClassifiedResponse<FailureClass, ClassifyEos> {
    /// The response was able to be classified immediately.
    Ready(Result<(), FailureClass>),
    /// We have to wait until the end of a streaming response to classify it.
    RequiresEos(ClassifyEos),
}
```

`GrpcErrorsAsFailures` was changed in a very similar way.

## Additional wins

Since `ClassifyResponse` is no longer generic over the error type `Trace` could remove that from its list of generic type params. That meant we could derive `Clone` and `Debug` impls which is much cleaner.

### DIFF
--- a/tower-http/src/classify.rs
+++ b/tower-http/src/classify.rs
@@ -1,7 +1,7 @@
 //! Tools for classifying responses as either success or failure.
 
 use http::{HeaderMap, Request, Response, StatusCode};
-use std::{convert::Infallible, marker::PhantomData, num::NonZeroI32};
+use std::{convert::Infallible, fmt, marker::PhantomData, num::NonZeroI32};
 
 /// Trait for producing response classifiers from a request.
 ///
@@ -12,10 +12,9 @@ use std::{convert::Infallible, marker::PhantomData, num::NonZeroI32};
 /// This is necessary for [`ClassifyResponse::classify_error`].
 ///
 /// [`Error` type]: https://docs.rs/tower/latest/tower/trait.Service.html#associatedtype.Error
-pub trait MakeClassifier<E> {
+pub trait MakeClassifier {
     /// The response classifier produced.
     type Classifier: ClassifyResponse<
-        E,
         FailureClass = Self::FailureClass,
         ClassifyEos = Self::ClassifyEos,
     >;
@@ -28,7 +27,7 @@ pub trait MakeClassifier<E> {
     type FailureClass;
 
     /// The type used to classify the response end of stream (EOS).
-    type ClassifyEos: ClassifyEos<E, FailureClass = Self::FailureClass>;
+    type ClassifyEos: ClassifyEos<FailureClass = Self::FailureClass>;
 
     /// Returns a response classifier for this request
     fn make_classifier<B>(&self, req: &Request<B>) -> Self::Classifier;
@@ -46,17 +45,17 @@ pub struct SharedClassifier<C> {
 
 impl<C> SharedClassifier<C> {
     /// Create a new `SharedClassifier` from the given classifier.
-    pub fn new<E>(classifier: C) -> Self
+    pub fn new(classifier: C) -> Self
     where
-        C: ClassifyResponse<E> + Clone,
+        C: ClassifyResponse + Clone,
     {
         Self { classifier }
     }
 }
 
-impl<C, E> MakeClassifier<E> for SharedClassifier<C>
+impl<C> MakeClassifier for SharedClassifier<C>
 where
-    C: ClassifyResponse<E> + Clone,
+    C: ClassifyResponse + Clone,
 {
     type FailureClass = C::FailureClass;
     type ClassifyEos = C::ClassifyEos;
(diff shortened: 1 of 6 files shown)


tower-http/src/classify.rs
```diff
@@ -351,7 +402,8 @@ mod usable_for_retries {
 
     impl<ReqB, ResB, E, C> Policy<Request<ReqB>, Response<ResB>, E> for RetryBasedOnClassification<C>
     where
-        C: ClassifyResponse<E> + Clone,
+        C: ClassifyResponse + Clone,
+        E: fmt::Display + 'static,
```
LucioFranco, line 406: is Error a better bound here? since it includes display?

tower-http/src/trace/layer.rs
```diff
@@ -74,13 +47,12 @@ impl<M, E> TraceLayer<M, E> {
...
             on_body_chunk: DefaultOnBodyChunk::default(),
             on_response: DefaultOnResponse::default(),
-            _error: PhantomData,
         }
     }
 }
 
-impl<M, E, MakeSpan, OnRequest, OnResponse, OnBodyChunk, OnEos, OnFailure>
-    TraceLayer<M, E, MakeSpan, OnRequest, OnResponse, OnBodyChunk, OnEos, OnFailure>
+impl<M, MakeSpan, OnRequest, OnResponse, OnBodyChunk, OnEos, OnFailure>
```
LucioFranco, line 54: image

LucioFranco (approved):
I just have a slight concern about the display vs error when composing and producing crazy error message but its a non blocking question.

tower-http/src/classify.rs
davidpdrsn: I wish we could require `std::error::Error` but since `Box<dyn std::error::Error>` doesn't implement `std::error::Error` it wouldn't work 😞

tower-http/src/trace/layer.rs
davidpdrsn: 😅

tower-http/src/classify.rs
LucioFranco: lovely...hahah..sounds good if it compiles etc I am happy with it

PR #95, opened 2021-05-21
Add `AddAuthorization` middleware

davidpdrsn:
This is the client counterpart to `RequireAuthorization`. It sets the
`Authorization` with basic or bearer token auth. Its meant to be
straight forward to use and if someone has a more complex use case
they can use `SetRequestHeader` instead which is more flexible.

TODO:
- [x] Update changelog


tower-http/src/auth/add_authorization.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// Authorize requests using a username and password pair.
+    ///
+    /// The `Authorization` header will be set to `Basic {credentials}` where `credentials` is
+    /// `base64_encode("{username}:{password}")`.
+    ///
+    /// Since the username and password is sent in clear text it is recommended to use HTTPS/TLS
+    /// with this method. However use of HTTPS/TLS is not enforced by this middleware.
+    pub fn basic(username: &str, password: &str) -> Self {
+        let encoded = base64::encode(format!("{}:{}", username, password));
+        let value = HeaderValue::from_str(&format!("Basic {}", encoded)).unwrap();
```
kazk, line 70:
```suggestion
        let mut value = HeaderValue::from_str(&format!("Basic {}", encoded)).unwrap();
        value.set_sensitive(true);
```

davidpdrsn: I think adding a doc comment about the value not being marked as sensitive and linking to https://docs.rs/tower-http/0.1.0/tower_http/sensitive_headers/struct.SetSensitiveRequestHeaders.html is a bit nicer.

kazk: Yeah, I thought that might be the reason. Is there a reason not to set sensitive?

In `kube`, we provide a default stack based on config, and it has an optional authorization layer (`Option<Either<AddAuthorizationLayer, SomeOtherAuthLayer>>`) applied.

```rust
let stack = ServiceBuilder::new()
    .option_layer(maybe_auth) // depends on the config
    // ... layers
    .layer(DecompressionLayer::new())
    .layer(TraceLayer::new_for_http()) // BTW, thanks for these :)
    .build(hyper_client);
```
I can add `.layer(SetSensitiveHeadersLayer::new(once(AUTHORIZATION)))`, but feels awkward since the authorization header is optional. It's not a huge deal, so I'll switch to whatever `tower-http` provides, but it was a bit unexpected.

davidpdrsn: Whoa so cool that you're using tower-http in kube ❤️

Having thought a bit more about this I think a builder style config method would probably nice. So something like:

```rust
.layer(AddAuthorizationLayer::basic("username", "password").as_sensitive(true))
```

Not sure that `as_sensitive` is the best name so suggestions welcome.

kazk: Yeah, that would be nice :+1: 
`as_sensitive` looks fine to me. But just `sensitive()` (no bool) works too.

davidpdrsn: I think the method taking a bool is nice because you can more easily set it according to some env var or something. Like `.as_sensitive(env::var("LOCAL").is_ok())`.

I've added it in 94a7cb012d0be.

LucioFranco (approved):

tesaguri (commented):
A couple of minor nits.

PR #96, opened 2021-05-26
Add `LifeCycleHooks` for adding callbacks to requests

davidpdrsn:
Edit: Things have changed a bit since this description was written. See https://github.com/tower-rs/tower-http/pull/96#discussion_r652613094

----

This adds `metrics::Traffic` which is a middleware for measuring requests per second/minute and error rates. Doesn't actually aggregate anything. Just facilitates calling some callbacks that do the actual work such as sending events to prometheus.

Example usage:

```rust
#[tokio::test]
async fn unary_request() {
    let mut svc = ServiceBuilder::new()
        .layer(TrafficLayer::new(
            ServerErrorsAsFailures::make_classifier(),
            MyCallbacks,
        ))
        .service_fn(echo);

    svc.ready()
        .await
        .unwrap()
        .call(Request::new(Body::from("foobar")))
        .await
        .unwrap();
}

async fn echo(req: Request<Body>) -> Result<Response<Body>, hyper::Error> {
    Ok(Response::new(req.into_body()))
}

#[derive(Clone)]
struct MyCallbacks;

struct Data {
    uri: Uri,
    method: Method,
    version: Version,
    request_received_at: Instant,
}

impl Callbacks<ServerErrorsFailureClass> for MyCallbacks {
    type Data = Data;

    fn prepare<B>(&mut self, request: &Request<B>) -> Self::Data {
        Data {
            uri: request.uri().clone(),
            method: request.method().clone(),
            version: request.version(),
            request_received_at: Instant::now(),
        }
    }

    fn on_response<B>(
        &mut self,
        response: &Response<B>,
        classification: ClassifiedResponse<ServerErrorsFailureClass, ()>,
        data: &mut Data,
    ) {
        // ...
    }

    fn on_eos(
        self,
        trailers: Option<&HeaderMap>,
        classification: Result<(), ServerErrorsFailureClass>,
        data: Data,
    ) {
        // ...
    }

    fn on_failure(
        self,
        failed_at: FailedAt,
        failure_classification: ServerErrorsFailureClass,
        data: Data,
    ) {
        // ...
    }
}
```

There are details about exactly when everything is called in the docs.

Part of https://github.com/tower-rs/tower-http/issues/57


tower-http/src/metrics/traffic/metrics_sink.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// - [`Body::poll_trailers`] returns an error.
+    ///
+    /// That means this method is _not_ called if a response is classified as a failure (then
+    /// [`on_response`] is called) or an end-of-stream is classified as a failure (then [`on_eos`]
+    /// is called).
+    ///
+    /// `failed_at` specifies where the error happened.
+    ///
+    /// The default implementation does nothing and returns immediately.
+    ///
```
davidpdrsn, line 97: I'm a bit unsure if this design is actually correct. Specifically I'm wondering if `Body::poll_data` can return multiple failures. If so that'll require changes to the classification. I've asked on Discord. Until thats been cleared we cannot merge this.

davidpdrsn:
@tobz would you maybe be up for reviewing this?

tower-http/CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-None.
+- Add `metrics::Traffic` for measuring how many responses or errors a service is
+  producing.
```
tobz, line 11: This is potentially a more philosophical thought but `Traffic` feels like it should almost do all of this, _plus_ track response latency, _plus_ track the in-flight count.  The last major thing to track, IMO, is response latency... which would bring the middleware count up to three.

davidpdrsn: Yeah I've had similar thoughts. I've considered calling it something like `RequestLifecycleCallbacks` 😜 but better.

The intended use case is to use this together with some metrics backend chosen by the user. At Embark we use the `metrics` crate and prometheus meaning all the aggregation is taken care of.

As mentioned I also do want to provide another crate that provides ready to use integration between tower-http and the metrics crate. However I don't think baking the metrics crate directly in to tower-http is appropriate.

tower-http/src/metrics/traffic/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+            );
+        }
+
+        fn on_eos(
+            self,
+            _trailers: Option<&HeaderMap>,
+            classification: Result<(), ServerErrorsFailureClass>,
+            data: Data,
+        ) {
+            let stream_duration = data.stream_start.unwrap().elapsed().as_millis() as f64;
```
tobz, line 224: Same comment as above WRT `as_secs_f64()`.

tower-http/src/metrics/traffic/mod.rs
```diff
@@ -0,0 +1,272 @@
+//! Middleware for adding high level traffic metrics to a [`Service`].
+//!
+//! The primary focus of this middleware is to enable adding request per second/minute and error
+//! rate metrics.
+//!
+//! The middleware doesn't do any kind of aggregate but instead uses a [`Callbacks`] which
+//! contains callbacks that the middleware will call. These methods can call into your actual
+//! metrics system as appropriate. See [`Callbacks`] for details on when each callback is called.
+//!
+//! Additionally it uses a [classifier] to determine if responses are success or failure.
```
tobz, line 10:
```suggestion
//! Additionally, it uses a [classifier] to determine if responses are success or failure.

File: tower-http/src/metrics/traffic/future.rs [hunk 5]
```diff
@@ -0,0 +1,95 @@
...
+        let this = self.project();
+        let result = ready!(this.inner.poll(cx));
+
+        let classifier = this.classifier.take().unwrap();
+
+        match result {
+            Ok(res) => {
+                let classification = classifier.classify_response(&res);
+                let mut callbacks: CallbacksT = this.callbacks.take().unwrap();
+                let mut callbacks_data = this.callbacks_data.take().unwrap();
```
tobz, line 50: Might as well hoist these up to line 44? You have to `take()` them regardless of `Ok(..)` vs `Err(..)`.

tobz: Also, is the type hint actually required for `callbacks`?

tower-http/src/metrics/traffic/service.rs
```diff
@@ -0,0 +1,76 @@
...
+/// [`Layer`]: tower_layer::Layer
+/// [`Service`]: tower_service::Service
+#[derive(Debug, Clone)]
+pub struct Traffic<S, M, Callbacks> {
+    pub(super) inner: S,
+    pub(super) make_classifier: M,
+    pub(super) callbacks: Callbacks,
+}
+
+impl<S, M, Callbacks> Traffic<S, M, Callbacks> {
```
tobz, line 25: `Callbacks` should be `CallbacksT` here, seemingly?  Just to follow the same pattern as the actual service impl.

tobz: Another question, too: maybe just use `C` for all generic type parameters where we're taking a `Callbacks` implementation? It feels like a code smell to have so many spots where we need to differentiate with `CallbacksT` in order to avoid confusion.

tower-http/src/metrics/traffic/future.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Response future for [`Traffic`].
+///
+/// [`Traffic`]: crate::metrics::Traffic
+#[pin_project]
+pub struct ResponseFuture<F, C, Callbacks, CallbacksData> {
+    #[pin]
+    pub(super) inner: F,
+    pub(super) classifier: Option<C>,
+    pub(super) request_received_at: Instant,
```
tobz, line 23: I'm not seeing a spot where this ever gets used?  Do we need it?  Seems like callers could capture this in their callback data.

tower-http/src/metrics/traffic/future.rs
```diff
@@ -0,0 +1,95 @@
...
+    E: fmt::Display + 'static,
+{
+    type Output =
+        Result<Response<ResponseBody<ResBody, C::ClassifyEos, CallbacksT, CallbacksT::Data>>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let result = ready!(this.inner.poll(cx));
+
+        let classifier = this.classifier.take().unwrap();
```
tobz, line 44: Do you think it might make more sense if we replace `unwrap()` with `expect("polled ResponseFuture after completion")`?

This would be more consistent with existing messages provided by the various futures types in the `futures` crate, etc.

tobz (changes requested):
Solid foundation so far.

tower-http/src/metrics/traffic/service.rs
davidpdrsn: The pattern I've been following elsewhere in tower-http is to fully spell out the name of the generic type if there is another type close by with a similar name. For example `C` might be `Classifier` or `Callbacks`.

And then since you cannot do `Callbacks: Callbacks` in bounds I would use `CallbacksT` for the type in those places.

Not a perfect pattern but thats at least the reasoning.

tower-http/src/metrics/traffic/future.rs
davidpdrsn: Nah the type hint isn't needed. Was just to help rust-analyzer while developing it. I'll just remove it.

tower-http/src/metrics/traffic/future.rs
davidpdrsn: Good catch! Can indeed be removed.

tower-http/CHANGELOG.md
davidpdrsn: @tobz Do you think think `RequestResponseCallbacks` would be a better name?

tower-http/src/metrics/traffic/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+                    is_error = class.is_err();
+                    is_stream = false;
+                }
+                ClassifiedResponse::RequiresEos(_) => {
+                    is_error = false;
+                    is_stream = true;
+                }
+            }
+
+            let duration_ms = data.request_received_at.elapsed().as_millis() as f64;
```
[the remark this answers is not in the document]
tobz: The problem is that you're only capturing _whole_ milliseconds with `as_millis()`.  You're potentially crushing resolution significantly for services with fast response times.

tower-http/CHANGELOG.md
tobz: Hmmm, something with lifecycle in the name definitely sounds better.  Just spitballing, but maybe like... lifecycle hooks? Something like that.

tower-http/src/metrics/traffic/service.rs
tobz: Yeah, I mean, it makes sense.  I personally find the fully-spelled out versions a little harder to remember vs the acronyms, but I know it's an established pattern in Tower so there's no eliminating it altogether. :)

My gut feeling is that it should just be consistently applied, and you're using `CallbacksT` in some places, so that seems like the pattern that should be followed, IMO.

tower-http/CHANGELOG.md
davidpdrsn: I've renamed and moved some things around. This middleware is now called `LifeCycleHooks` and lives in `tower_http::life_cycle`, so doesn't live in `metrics` anymore. I agree I think this makes more sense.

@tobz what do you think?

davidpdrsn:
I think I'm going to close this for now. I'm not confident in the design. Specifically I think it would be cool to have something that was powerful enough to implement the `Trace` middleware on top of. That isn't possible with the current design.

Issue #73, opened 2021-03-03
Add more common response classifiers


davidpdrsn:
Once https://github.com/tower-rs/tower-http/pull/49 is merged we should consider adding more classifiers similar to `ServerErrorsAsFailures` for common cases. One that takes a list of ranges might be useful. Then you could easily consider anything within `400..=599` errors.

Blocked by https://github.com/tower-rs/tower-http/pull/90

PR #101, opened 2021-05-29
Add `StatusInRangeAsFailures` response classifier

davidpdrsn:
`StatusInRangeAsFailures` is a response classifier that considers
responses with status code in a certain range as failures.

This is useful for HTTP clients where both server errors (5xx) and
client errors (4xx) are considered failures.

Fixes https://github.com/tower-rs/tower-http/issues/73


tobz (approved):
Seems fine, tbh.

PR #103, opened 2021-05-29
Clarify exactly when each `Trace` callback is called

davidpdrsn:


tower-http/src/trace/body.rs
```diff
@@ -77,26 +77,30 @@ where
...
     ) -> Poll<Result<Option<HeaderMap>, Self::Error>> {
-        let this = self.project();
+        let mut this = self.project();
         let _guard = this.span.enter();
         let result = ready!(this.inner.poll_trailers(cx));
 
         let latency = this.start.elapsed();
 
-        if let Some((classify_eos, mut on_failure)) =
-            this.classify_eos.take().zip(this.on_failure.take())
```
davidpdrsn, line 87: `this.on_failure.take()` meant we would only call `on_failure` for the first failure. That was a bug. It should be called for each failure.

davidpdrsn: Looking at this again it still might not be correct since we're still doing `this.classify_eos.take()`. Meaning we can only classify one error and so the callback will still only be called once.

I've asked for clarification on Discord about what the contract of `http_body::Body` is.

davidpdrsn: Hyper will not call `poll_data` again if it returns an error. So this is fine.

tower-http/CHANGELOG.md
```diff
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-None.
+- Clarify exactly when each `Trace` callback is called.
+- Call `Trace`'s `on_failure` callback each time the response body produces an
+  error (in `http_body::Body::poll_data`). It would previously only be called
+  the first time.
```
davidpdrsn, line 13: Note to self: Remove this line. It's wrong. It is only called the first time poll_data fails.

tower-http/src/trace/body.rs
```diff
@@ -77,26 +77,30 @@ where
...
-                    }
+                if let Some((on_eos, stream_start)) = this.on_eos.take() {
+                    on_eos.on_eos(trailers.as_ref(), stream_start.elapsed(), this.span);
                 }
-                Err(err) => {
+            }
+            Err(err) => {
+                if let (Some(classify_eos), Some(on_failure)) =
+                    (this.classify_eos.take(), &mut this.on_failure)
+                {
```
davidpdrsn, line 103: Note to self: Should probably revert the changes to this file as they don't actually change anything.

tobz (approved):

PR #105, opened 2021-06-02
Fix documentation for `DefaultOnFailure::level`

kazk:
## Motivation

https://github.com/tower-rs/tower-http/blob/27db4f2b3a4fb936986189f7311ff14cf7ff5f63/tower-http/src/trace/on_failure.rs#L47-L54


davidpdrsn (approved):
Good catch. Thanks!

PR #106, opened 2021-06-04
Add `ServiceBuilderExt`

davidpdrsn:
Adds `ServiceBuilderExt` which adds methods to `tower::ServiceBuilder`
for adding middleware from tower-http. Figured this might be a nice
convenience to have.

Not all middleware are available since they require some specific
configuration. For those I think adding the layer explicitly is better.


seanmonstar (approved):

Nehliin:
Nice!

PR #107, opened 2021-06-04
Change compression body error type to `BoxError`

davidpdrsn:
The compression related middleware would previously use `BodyOrIoError`
as the body error type. That only implemented `std::error::Error` if the
inner error did as well. Since `Box<dyn std::error::Error + Send + Sync>`
does not implement `std::error::Error` `Compression` and
`Decompression` wouldn't be usable with hyper if the body they wrapped
had that error type.

This changes the middleware to use `BoxError` as the error type which
resolves the issue. Its also consistent with other middleware.

Note: This is a breaking change so should be released in 0.2 when we decide to do that.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -11,7 +11,12 @@ None.
 
 ## Breaking changes
 
-None.
+- Change the response body error type of `Compression` and `Decompression` to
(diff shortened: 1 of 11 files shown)


davidpdrsn:
Superseded by https://github.com/tower-rs/tower-http/pull/166

PR #109, opened 2021-06-16
Add `RequestBodyTimeout` and `ResponseBodyTimeout`

davidpdrsn:
This adds two new middleware:

- `RequestBodyTimeout`
- `ResponseBodyTimeout`

They both do pretty much what it says on the tin.

### DIFF
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -335,3 +335,6 @@ pub enum LatencyUnit {
     /// Use nanoseconds.
     Nanos,
 }
+
+/// Alias for a type-erased error type.
+pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -210,12 +210,15 @@
     clippy::match_like_matches_macro,
     clippy::type_complexity
 )]
-#![forbid(unsafe_code)]
+#![cfg_attr(not(test), forbid(unsafe_code))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(test, allow(clippy::float_cmp))]
 
 #[macro_use]
 pub(crate) mod macros;
+
+#[cfg(test)]
+pub(crate) mod test_utils;
 
 #[cfg(feature = "auth")]
 #[cfg_attr(docsrs, doc(cfg(feature = "auth")))]
@@ -267,6 +270,10 @@ pub mod follow_redirect;
 #[cfg(feature = "metrics")]
 #[cfg_attr(docsrs, doc(cfg(feature = "metrics")))]
 pub mod metrics;
+
+#[cfg(feature = "timeout")]
+#[cfg_attr(docsrs, doc(cfg(feature = "timeout")))]
+pub mod timeout;
 
 pub mod classify;
 pub mod services;
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -94,10 +94,12 @@
 //! use tower_http::{
 //!     decompression::DecompressionLayer,
 //!     set_header::SetRequestHeaderLayer,
+//!     timeout::RequestBodyTimeoutLayer,
 //! };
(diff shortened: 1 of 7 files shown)


tower-http/src/timeout/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Only start the timeout after first poll of the body. This enum manages that.
+#[allow(clippy::large_enum_variant)]
+#[pin_project(project = StateProj)]
+#[derive(Debug)]
+enum State {
+    NotPolled(Duration),
+    SleepPending(#[pin] Sleep),
+}
```
davidpdrsn, line 52: Not sure how useful this would be. I'm just wondering if users might receive a request and not buffer the body immediately and hit the timeout because they were slow to call the first poll.

davidpdrsn:
Requires merging https://github.com/tower-rs/tower-http/pull/110 to avoid the duplicate nom dependency.

tower-http/src/timeout/mod.rs
LucioFranco: would it make sense to have a Lazy and non lazy version?

tower-http/src/timeout/response_body.rs
```diff
@@ -0,0 +1,231 @@
...
+        .await
+        .expect("body was empty")
+        .unwrap_err();
+        assert!(poll_data_err.is::<tower::timeout::error::Elapsed>());
+
+        let poll_trailers_err = futures::future::poll_fn(|cx| {
+            unsafe { Pin::new_unchecked(&mut body) }.poll_trailers(cx)
+        })
+        .await
+        .unwrap_err();
```
tesaguri, line 184:
```suggestion
        let body = res.into_body();
        tokio::pin!(body);

        let poll_data_err = futures::future::poll_fn(|cx| body.as_mut().poll_data(cx))
            .await
            .expect("body was empty")
            .unwrap_err();
        assert!(poll_data_err.is::<tower::timeout::error::Elapsed>());

        let poll_trailers_err = futures::future::poll_fn(|cx| body.as_mut().poll_trailers(cx))
            .await
            .unwrap_err();
```

tower-http/src/timeout/mod.rs
```diff
@@ -0,0 +1,132 @@
+//! Middleware that adds timeouts to request or response bodies.
+//!
+//! Note these middleware differ from [`tower::timeout::Timeout`] which only
+//! adds a timeout to the response future and doesn't consider request bodies.
```
jplatte, line 4: What does this mean for `ResponseBodyTimeout`? Is it equivalent to the tower middleware?

tower-http/src/timeout/mod.rs
jplatte: It's common to first hit the DB for an auth check with a value from an authorization header before starting to look at the http body. This seems like a good idea to me.

(based on my understanding that the body would usually be polled immediately, except for use cases where the body is streamed)

tower-http/src/timeout/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn is_end_stream(&self) -> bool {
+        self.inner.is_end_stream()
+    }
+
+    fn size_hint(&self) -> http_body::SizeHint {
+        self.inner.size_hint()
+    }
+}
+
+impl<B> Default for TimeoutBody<B>
```
jplatte, line 123: What is this impl useful for?

tower-http/src/lib.rs
```diff
@@ -217,13 +218,16 @@
     clippy::match_like_matches_macro,
     clippy::type_complexity
 )]
-#![forbid(unsafe_code)]
+#![cfg_attr(not(test), forbid(unsafe_code))]
```
jplatte, line 221: Seems like all unsafe usage went away in later commits, so this can be reverted?

LucioFranco:
@davidpdrsn we should add this stuff into tonic and let it mature there maybe? Its quite an important feature for running code in prod.

davidpdrsn:
Sure we could do that!

PR #112, opened 2021-07-16
Add `Cors` for setting CORS headers

davidpdrsn:
Example usage:

```rust
use http::{Request, Response, Method, header};
use hyper::Body;
use tower::{ServiceBuilder, ServiceExt, Service};
use tower_http::cors::{CorsLayer, Any};
use std::convert::Infallible;

async fn handle(request: Request<Body>) -> Result<Response<Body>, Infallible> {
    Ok(Response::new(Body::empty()))
}

let cors = CorsLayer::new()
    .allow_methods(vec![Method::GET, Method::POST, Method::OPTIONS])
    .allow_origin(Any)
    .allow_credentials(false);

let mut service = ServiceBuilder::new()
    .layer(cors)
    .service_fn(handle);

let request = Request::builder()
    .header(header::ORIGIN, "https://example.com")
    .body(Body::empty())
    .unwrap();

let response = service
    .ready()
    .await?
    .call(request)
    .await?;

assert_eq!(
    response.headers().get(header::ACCESS_CONTROL_ALLOW_ORIGIN).unwrap(),
    "*",
);
```

Implementation is based on [`CorsMiddleware`](https://docs.rs/tide/0.16.0/tide/security/struct.CorsMiddleware.html#method.allow_credentials) from tide.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -10,10 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix the bug that `ServeDir` can not handle percent encoded uri path correctly.
 - Enhancement: avoid using permanent redirection when redirect directory path without a trailing slash to the one has
 - Fix a [bug](https://github.com/tower-rs/tower-http/issues/121) which happens when `append_index_html_on_directories` is set to `false` in `ServeDir`.
+- Add `Cors` for setting [CORS] headers.
 
 ## Breaking changes
 
-- Add `Cors` for setting [CORS] headers.
+None.
 
 # 0.1.1 (July 2, 2021)
 
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Breaking changes
 
-None.
+- Add `Cors` for setting [CORS] headers.
 
 # 0.1.1 (July 2, 2021)
 
@@ -37,3 +37,5 @@ None.
 # 0.1.0 (May 27, 2021)
 
 - Initial release.
+
+[CORS]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
--- a/tower-http/src/cors.rs
+++ b/tower-http/src/cors.rs
@@ -76,10 +76,9 @@ pub struct CorsLayer {
     allow_methods: HeaderValue,
     allow_origin: AnyOr<Origin>,
     expose_headers: Option<HeaderValue>,
-    max_age: HeaderValue,
+    max_age: Option<HeaderValue>,
 }
 
-const DEFAULT_MAX_AGE: &str = "86400"; // 24 hours
 const DEFAULT_METHODS: &str = "GET, POST, OPTIONS";
 const WILDCARD: &str = "*";
 
@@ -92,7 +91,7 @@ impl CorsLayer {
             allow_methods: DEFAULT_METHODS.parse().unwrap(),
             allow_origin: AnyOr(AnyOrInner::Any),
             expose_headers: None,
-            max_age: DEFAULT_MAX_AGE.parse().unwrap(),
+            max_age: None,
         }
     }
 
@@ -147,11 +146,15 @@ impl CorsLayer {
     /// let layer = CorsLayer::new().max_age(Duration::from_secs(60) * 10);
     /// ```
     ///
-    /// By default the header will be set to 24 hours.
+    /// By default the header will not be set will disables caching and will require a preflight
+    /// call for all requests.
+    ///
+    /// Note that each browser has a maximum internal value that takes precedence when the
+    /// Access-Control-Max-Age is greater. For more details see [mdn].
     ///
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
     pub fn max_age(mut self, max_age: Duration) -> Self {
-        self.max_age = max_age.as_secs().into();
+        self.max_age = Some(max_age.as_secs().into());
         self
     }
 
@@ -482,9 +485,11 @@ impl<S> Cors<S> {
             self.layer.allow_headers.clone(),
         );
 
-        response
-            .headers_mut()
-            .insert(header::ACCESS_CONTROL_MAX_AGE, self.layer.max_age.clone());
+        if let Some(max_age) = self.layer.max_age.clone() {
+            response
+                .headers_mut()
+                .insert(header::ACCESS_CONTROL_MAX_AGE, max_age);
+        }
 
         if let Some(allow_credentials) = self.layer.allow_credentials.clone() {
             response
--- a/tower-http/src/cors.rs
+++ b/tower-http/src/cors.rs
@@ -1,0 +1,627 @@
+//! Middleware which adds headers for [CORS][mdn].
+//!
+//! # Example
+//!
+//! ```
+//! use http::{Request, Response, Method, header};
+//! use hyper::Body;
+//! use tower::{ServiceBuilder, ServiceExt, Service};
+//! use tower_http::cors::{CorsLayer, Any};
+//! use std::convert::Infallible;
+//!
+//! async fn handle(request: Request<Body>) -> Result<Response<Body>, Infallible> {
+//!     Ok(Response::new(Body::empty()))
+//! }
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! let cors = CorsLayer::new()
+//!     .allow_methods(vec![Method::GET, Method::POST, Method::OPTIONS])
+//!     .allow_origin(Any)
+//!     .allow_credentials(false);
+//!
+//! let mut service = ServiceBuilder::new()
+//!     .layer(cors)
+//!     .service_fn(handle);
+//!
+//! let request = Request::builder()
+//!     .header(header::ORIGIN, "https://example.com")
+//!     .body(Body::empty())
+//!     .unwrap();
+//!
+//! let response = service
+//!     .ready()
+//!     .await?
+//!     .call(request)
+//!     .await?;
+//!
+//! assert_eq!(
+//!     response.headers().get(header::ACCESS_CONTROL_ALLOW_ORIGIN).unwrap(),
+//!     "*",
+//! );
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+
+use futures_core::ready;
+use http::{
+    header::{self, HeaderName, HeaderValue},
+    Method, Request, Response, StatusCode,
+};
+use pin_project::pin_project;
+use std::{
+    future::Future,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+    time::Duration,
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Layer that applies [`Cors`] which adds headers for [CORS][mdn].
+///
+/// See the [module docs](crate::cors) for an example.
+///
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+#[derive(Debug, Clone)]
+pub struct CorsLayer {
+    allow_credentials: Option<HeaderValue>,
+    allow_headers: HeaderValue,
+    allow_methods: HeaderValue,
+    allow_origin: AnyOr<Origin>,
+    expose_headers: Option<HeaderValue>,
+    max_age: HeaderValue,
+}
+
+const DEFAULT_MAX_AGE: &str = "86400"; // 24 hours
+const DEFAULT_METHODS: &str = "GET, POST, OPTIONS";
+const WILDCARD: &str = "*";
+
+impl CorsLayer {
+    /// Create a new `CorsLayer`.
+    pub fn new() -> Self {
+        Self {
+            allow_credentials: None,
+            allow_headers: WILDCARD.parse().unwrap(),
+            allow_methods: DEFAULT_METHODS.parse().unwrap(),
+            allow_origin: AnyOr(AnyOrInner::Any),
+            expose_headers: None,
+            max_age: DEFAULT_MAX_AGE.parse().unwrap(),
+        }
+    }
+
+    /// Set the value of the [`Access-Control-Allow-Credentials`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    ///
+    /// let layer = CorsLayer::new().allow_credentials(true);
+    /// ```
+    ///
+    /// By default the header will not be set.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials
+    pub fn allow_credentials(mut self, allow_credentials: bool) -> Self {
+        self.allow_credentials = Some(allow_credentials.to_string().parse().unwrap());
+        self
+    }
+
+    /// Set the value of the [`Access-Control-Allow-Headers`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    /// use http::header::HeaderValue;
+    ///
+    /// let layer = CorsLayer::new().allow_headers(vec![
+    ///     "*".parse().unwrap(),
+    /// ]);
+    /// ```
+    ///
+    /// By default the header will be set to `*`.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers
+    pub fn allow_headers<I>(mut self, headers: I) -> Self
+    where
+        I: IntoIterator<Item = HeaderValue>,
+    {
+        self.allow_headers = separated_by_commas(
+            headers
+                .into_iter()
+                .map(|value| value.to_str().unwrap().to_string()),
+        );
+        self
+    }
+
+    /// Set the value of the [`Access-Control-Max-Age`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    /// use std::time::Duration;
+    ///
+    /// let layer = CorsLayer::new().max_age(Duration::from_secs(60) * 10);
+    /// ```
+    ///
+    /// By default the header will be set to 24 hours.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
+    pub fn max_age(mut self, max_age: Duration) -> Self {
+        self.max_age = max_age.as_secs().into();
+        self
+    }
+
+    /// Set the value of the [`Access-Control-Allow-Methods`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    /// use http::Method;
+    ///
+    /// let layer = CorsLayer::new().allow_methods(vec![Method::GET, Method::POST]);
+    /// ```
+    ///
+    /// All methods can be allowed with
+    ///
+    /// ```
+    /// use tower_http::cors::{CorsLayer, Any};
+    ///
+    /// let layer = CorsLayer::new().allow_methods(Any);
+    /// ```
+    ///
+    /// By default the header will be set to `GET, POST, OPTIONS`.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods
+    pub fn allow_methods<T>(mut self, methods: T) -> Self
+    where
+        T: Into<AnyOr<Vec<Method>>>,
+    {
+        self.allow_methods = match methods.into().0 {
+            AnyOrInner::Any => "*".parse().unwrap(),
+            AnyOrInner::Value(methods) => separated_by_commas(methods),
+        };
+        self
+    }
+
+    /// Set the value of the [`Access-Control-Allow-Origin`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::{CorsLayer, Origin};
+    ///
+    /// let layer = CorsLayer::new().allow_origin(Origin::exact(
+    ///     "http://example.com".parse().unwrap(),
+    /// ));
+    /// ```
+    ///
+    /// All origins can be allowed with
+    ///
+    /// ```
+    /// use tower_http::cors::{CorsLayer, Any};
+    ///
+    /// let layer = CorsLayer::new().allow_origin(Any);
+    /// ```
+    ///
+    /// By default the header will be set to `*`.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
+    pub fn allow_origin<T>(mut self, origin: T) -> Self
+    where
+        T: Into<AnyOr<Origin>>,
+    {
+        self.allow_origin = origin.into();
+        self
+    }
+
+    /// Set the value of the [`Access-Control-Expose-Headers`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    /// use http::header::CONTENT_ENCODING;
+    ///
+    /// let layer = CorsLayer::new().expose_headers(vec![CONTENT_ENCODING]);
+    /// ```
+    ///
+    /// All headers can be allowed with
+    ///
+    /// ```
+    /// use tower_http::cors::{CorsLayer, Any};
+    ///
+    /// let layer = CorsLayer::new().expose_headers(Any);
+    /// ```
+    ///
+    /// By default the header will not be set.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+    pub fn expose_headers<I>(mut self, headers: I) -> Self
+    where
+        I: Into<AnyOr<Vec<HeaderName>>>,
+    {
+        self.expose_headers = Some(match headers.into().0 {
+            AnyOrInner::Any => "*".parse().unwrap(),
+            AnyOrInner::Value(headers) => separated_by_commas(headers),
+        });
+        self
+    }
+}
+
+/// Represents a wildcard value (`*`) used with some CORS headers such as
+/// [`CorsLayer::allow_methods`].
+#[derive(Debug, Clone, Copy)]
+pub struct Any;
+
+/// Used to make methods like [`CorsLayer::allow_methods`] more convenient to call.
+///
+/// You shouldn't have to use this type directly.
+#[derive(Debug, Clone, Copy)]
+pub struct AnyOr<T>(AnyOrInner<T>);
+
+#[derive(Debug, Clone, Copy)]
+enum AnyOrInner<T> {
+    Any,
+    Value(T),
+}
+
+impl From<Origin> for AnyOr<Origin> {
+    fn from(origin: Origin) -> Self {
+        AnyOr(AnyOrInner::Value(origin))
+    }
+}
+
+impl<T> From<Any> for AnyOr<T> {
+    fn from(_: Any) -> Self {
+        AnyOr(AnyOrInner::Any)
+    }
+}
+
+impl<I> From<I> for AnyOr<Vec<Method>>
+where
+    I: IntoIterator<Item = Method>,
+{
+    fn from(methods: I) -> Self {
+        AnyOr(AnyOrInner::Value(methods.into_iter().collect()))
+    }
+}
+
+impl<I> From<I> for AnyOr<Vec<HeaderName>>
+where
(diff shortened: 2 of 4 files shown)


jplatte:
I've had a look and one thing I'm missing is a little more flexibility for origins. `actix_cors` allows you to simply provide a closure to check origins which aren't in the allow list. In the crate docs this is part of their example:

```rust
.allowed_origin_fn(|origin, _req_head| {
  origin.as_bytes().ends_with(b".rust-lang.org")
})
```

and I've previously done a very similar thing with that crate, with an additional `origin.starts_with(b"https://")` check.

examples/tonic-key-value-store/Cargo.toml
```diff
@@ -9,16 +9,16 @@ license = "MIT"
...
 hyper = { version = "0.14.4", features = ["full"] }
-prost = "0.7"
+prost = "0.8"
 structopt = "0.3.21"
 tokio = { version = "1.2.0", features = ["full"] }
 futures = "0.3"
-tokio-stream = { version = "0.1", features = ["sync"] }
-tonic = "0.4"
+tokio-stream = { version = "0.1", features = ["sync", "net"] }
+tonic = "0.5"
```
hawkw, line 17: i think the prost and tonic updates are breaking changes, right? would have preferred to bump them in a separate PR (but it's not terribly important)

tower-http/CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ None.
 
 ## Breaking changes
 
-None.
+- Add `Cors` for setting [CORS] headers.
```
hawkw, line 14: i don't think adding the `Cors` middleware is a _breaking_ change?

tower-http/src/cors.rs
```diff
@@ -0,0 +1,627 @@
...
+//! async fn handle(request: Request<Body>) -> Result<Response<Body>, Infallible> {
+//!     Ok(Response::new(Body::empty()))
+//! }
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! let cors = CorsLayer::new()
+//!     .allow_methods(vec![Method::GET, Method::POST, Method::OPTIONS])
+//!     .allow_origin(Any)
+//!     .allow_credentials(false);
```
hawkw, line 21: it could be nice to have comments explaining what the different configuration methods are doing here?

hawkw (commented):
> I've had a look and one thing I'm missing is a little more flexibility for origins. `actix_cors` allows you to simply provide a closure to check origins which aren't in the allow list. In the crate docs this is part of their example:
> 
> ```rust
> .allowed_origin_fn(|origin, _req_head| {
>   origin.as_bytes().ends_with(b".rust-lang.org")
> })
> ```
> 
> and I've previously done a very similar thing with that crate, with an additional `origin.starts_with(b"https://")` check.

Hmm, yeah, I wonder if what we want is a trait for allowed origin predicates that can be implemented by closures _and_ by `T: Into<AnyOr<Origin>>`?

hawkw:
I wonder if it would make sense to implement the CORS logic as an [`AsyncPredicate`](https://docs.rs/tower/0.4.8/tower/filter/trait.AsyncPredicate.html) for use with [`tower::filter`](https://docs.rs/tower/0.4.8/tower/filter/trait.AsyncPredicate.html). The CORS layer could then build filter services with that predicate.

OTTOH, I'm not sure how much value there is from this — most of the meat is in the `Future` impl, which would stay the same either way.

examples/tonic-key-value-store/Cargo.toml
davidpdrsn: Yeah I probably should have. CI complained about the RUSTSEC in prost so just went ahead and did it here.

tower-http/CHANGELOG.md
davidpdrsn: 🤦

davidpdrsn:
> Hmm, yeah, I wonder if what we want is a trait for allowed origin predicates that can be implemented by closures and by T: Into<AnyOr<Origin>>?

I like that idea! I'll experiment.

davidpdrsn:
Another thing to consider is that since parts of this code is copied from tide, we have to include their license somewhere.

tower-http/src/cors.rs
```diff
@@ -0,0 +1,627 @@
...
+                OriginInner::List(list) => list.contains(origin),
+            },
+        }
+    }
+
+    fn build_preflight_response<B>(&self, origin: HeaderValue) -> Response<B>
+    where
+        B: Default,
+    {
+        let mut response = Response::new(B::default());
```
nkconnor, line 438: I think preflight should have no body which I don't think is guaranteed by `::default()`. I doubt any of the browser implementations depend on that, so, it's probably more of an ideal than anything. 

Still is `impl Default` for `Body` typical in `tower` and elsewhere? would it be more clear to have a different trait `EmptyBody` or requiring this from the builder..

nkconnor (commented):
Adding my comments from Discord so I can add follow-ups given the time

fourbytes:
Using Axum for a project at the moment and in need of a CORS layer, any idea when this will be ready?

tower-http/src/cors.rs
```diff
@@ -0,0 +1,627 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<S> Layer<S> for CorsLayer {
+    type Service = Cors<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        Cors::new(inner)
```
fourbytes, line 317: I've tried using the `cors` branch but it appears the CorsLayer options don't actually get used. Could be wrong, but shouldn't this pass the settings down to the Cors struct?

davidpdrsn:
> Using Axum for a project at the moment and in need of a CORS layer, any idea when this will be ready?

I'm gonna look into this once axum 0.2 is released, which will happen this week. I can't give an ETA on merging this however. Until then you can apply the CORS headers manually and add a catch-all route for OPTIONS requests.

fourbytes:
Ended up quickly [forking](https://github.com/fourbytes/tower-http/commit/e68bf19705f8a89a517990094824c1315dc6d987) this branch as I needed the ability to allow multiple origins which was difficult to do with manually injected headers.

With these changes it seems to be working nicely.

examples/tonic-key-value-store/Cargo.toml
davidpdrsn: Made [a new PR](https://github.com/tower-rs/tower-http/pull/133) and reverted the update here.

tower-http/src/cors.rs
davidpdrsn: Good catch! Fixed in df1d6e001ee6c

tower-http/src/cors.rs
davidpdrsn: All bodies in http-body, axum, and tower-http are empty in their `Default` impl. While I see your point I think its safe for now. Can fix if someone reports issues with it.

tower-http/src/cors.rs
```diff
@@ -0,0 +1,627 @@
...
+                    Response::builder()
+                        .status(StatusCode::UNAUTHORIZED)
+                        .body(ResBody::default())
+                        .unwrap(),
+                )),
+            };
+        }
+
+        // Return results immediately upon preflight request
+        if req.method() == Method::OPTIONS {
```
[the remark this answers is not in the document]
davidpdrsn: Thanks! Fixed in 1782deade72e7

tower-http/src/cors.rs
```diff
@@ -0,0 +1,627 @@
...
+    /// Set the value of the [`Access-Control-Max-Age`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    /// use std::time::Duration;
+    ///
+    /// let layer = CorsLayer::new().max_age(Duration::from_secs(60) * 10);
+    /// ```
+    ///
+    /// By default the header will be set to 24 hours.
```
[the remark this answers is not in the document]
davidpdrsn: I've changed the default to `None`. Its also what actix does.

davidpdrsn:
Allow-Origin can be now determined via a closure, similar to actix-cors:

```rust
use tower_http::cors::{CorsLayer, Any};
use http::{HeaderValue, request::Parts};

let layer = CorsLayer::new().allow_origin(|origin: &HeaderValue, _request_head: &Parts| {
    origin.as_bytes().ends_with(b".rust-lang.org")
});
```

@jplatte @hawkw @fourbytes @nkconnor Wanna give this another look?

nkconnor:
lg2m

tower-http/src/cors.rs
```diff
@@ -0,0 +1,718 @@
...
+            AnyOrInner::Value(allow_origin) => match &allow_origin.0 {
+                OriginInner::Exact(s) => s == origin,
+                OriginInner::List(list) => list.contains(origin),
+                OriginInner::Closure(f) => f(origin, parts),
+            },
+        }
+    }
+
+    fn is_valid_request_method(&self, method: &HeaderValue) -> bool {
+        self.layer.allow_methods == method
```
fourbytes, line 479: Looks like there's a bug here. This results in a 405 method not allowed for a valid CORS request because it's comparing the list of the allowed methods directly with the requested method (`GET,POST` != `GET`).

fourbytes (changes requested):
Doesn't seem to be working for me yet due to an issue with the valid method check. Other than that though, API looks great.

tower-http/src/cors.rs
davidpdrsn: Ah yeah of course! Thanks for noticing. Fixed in a4f515229f447

tower-http/src/cors.rs
```diff
@@ -0,0 +1,749 @@
...
+const WILDCARD: &str = "*";
+
+impl CorsLayer {
+    /// Create a new `CorsLayer`.
+    pub fn new() -> Self {
+        Self {
+            allow_credentials: None,
+            allow_headers: WILDCARD.parse().unwrap(),
+            allow_methods: DEFAULT_METHODS.parse().unwrap(),
+            allow_origin: AnyOr(AnyOrInner::Any),
```
jplatte, line 92: I'm not a fan of these defaults, they seem rather arbitrary. What are they based on?

tower-http/src/cors.rs
```diff
@@ -0,0 +1,749 @@
...
+    ///
+    /// ```
+    /// use tower_http::cors::{CorsLayer, Origin};
+    ///
+    /// let layer = CorsLayer::new().allow_origin(Origin::exact(
+    ///     "http://example.com".parse().unwrap(),
+    /// ));
+    /// ```
+    ///
+    /// Multiple origins can be allowed with
```
jplatte, line 202: Probably good to note that calling `.allow_origin` multiple times will not do the same thing.

jplatte (commented):
One thing this still doesn't allow which actix-cors allows is

```rust
Cors::ctor()
    .allowed_origin("http://127.0.0.1")
    .allowed_origin("foo.bar.org")
    .allowed_origin_fn(|origin, _| {
        let origin = origin.as_bytes();
        origin.starts_with(b"https://") && origin.ends_with(b".rust-lang.org")
    })
```

which also means that it's impossible with the current code structure for clients to discover some of the origins that would be allowed to use a route, at least when using a closure (it will look like no origins are allowed when doing the request from a disallowed origin). I *think* what actix-cors does is send all origins that are specified explicitly regardless of whether they match, and *additionally* send the request origin back in the header if the function marks it as allowed.

I don't think it would be a problem for me, but I thought I'd still bring it up.

tower-http/src/cors.rs
davidpdrsn: They based on tide https://github.com/http-rs/tide/blob/main/src/security/cors.rs#L37-L46.

I don't really have any preference and these worked in my testing. What would you suggest?

jplatte: For my own usage, I would prefer no headers to be sent by default, and me configuring all of them explicitly. An "allow-anything" constructor might additionally be useful for quick prototyping (actix-cors has [`Cors::permissive`](https://docs.rs/actix-cors/0.5.4/actix_cors/struct.Cors.html#method.permissive)).

davidpdrsn:
> which also means that it's impossible with the current code structure for clients to discover some of the origins that would be allowed to use a route, at least when using a closure (it will look like no origins are allowed when doing the request from a disallowed origin). I think what actix-cors does is send all origins that are specified explicitly regardless of whether they match, and additionally send the request origin back in the header if the function marks it as allowed.

Hm thats a good point. I was wondering why actix that it setup the way they did. Although I do think its unfortunate to have to store both a list a allowed origins, and a closure to call if nothing else matches 🤔 

> I don't think it would be a problem for me, but I thought I'd still bring it up.

Would be an issue for me either. I'm leaning towards keeping it as is 🤷

tower-http/src/cors.rs
davidpdrsn: Actually the docs in general could be more explicit about a lot of things. I'll do another pass over them later.

tower-http/src/cors.rs
davidpdrsn: I like that!

tower-http/src/cors.rs
```diff
@@ -0,0 +1,749 @@
...
+        // Return results immediately upon preflight request
+        if req.method() == Method::OPTIONS {
+            // the method the real request will be made with
+            match req.headers().get(header::ACCESS_CONTROL_REQUEST_METHOD) {
+                Some(request_method) => {
+                    if !self.is_valid_request_method(request_method) {
+                        return ResponseFuture {
+                            inner: Kind::Error(Some(
+                                Response::builder()
+                                    .status(StatusCode::METHOD_NOT_ALLOWED)
```
mathroc, line 619: I’m not sure returning a `METHOD_NOT_ALLOWED` (or `UNAUTHORIZED` below) is what’s supposed to happen here. Because from the client POV it would look like the `OPTIONS` method was not allowed / authorized

if `Access-Control-Allow-Methods` is present and does not include the `ACCESS_CONTROL_REQUEST_METHOD` it looks like it’s enough to stop the CORS pre flight request : https://fetch.spec.whatwg.org/#ref-for-concept-request-method%E2%91%A0%E2%91%A6

davidpdrsn: I see that `METHOD_NOT_ALLOWED` might be confusing. Actix returns `bad request` but their internal error is called `MethodNotAllowed` 🤷 Which status do you think would be more appropriate? I'm not a fan of just terminating the connection without returning a response.

mathroc: For a pre-fligh `OPTIONS` request, I would expect a `No Content` response, that includes  a (possibly empty) `Access-Control-Allow-Methods` header. 

In fact, if the `Access-Control-Allow-Methods` is always present (I did not check) and the list of authorized methods is static, I would not manually check for the `ACCESS_CONTROL_REQUEST_METHOD` header. 

So unless you plan to add a way to provide a callback to determine dynamically if a request method would be accepted, like with `allow_origin(|origin ...| ...)`

---

additional remarks:

* the current behavior seems appropriate when it’s not a pre-flight request but I’m not sure if this header is included in a non pre-flight request
* the `is_valid_origin` check looks okay for a non pre-flight request but should also return a 204 in pre-flight mode (with the `Access-Control-Allow-Origin` header containing or not the origin depending if it’s a valid origin

tower-http/src/cors.rs
```diff
@@ -0,0 +1,749 @@
...
+
+const DEFAULT_METHODS: &str = "GET,POST,OPTIONS";
+const WILDCARD: &str = "*";
+
+impl CorsLayer {
+    /// Create a new `CorsLayer`.
+    pub fn new() -> Self {
+        Self {
+            allow_credentials: None,
+            allow_headers: WILDCARD.parse().unwrap(),
```
DoumanAsh, line 90: Why not `HeaderValue::from_static(WILDCARD)` and below too?
Both should be always valid and you can leave unwrap inside `from_static`

tower-http/src/cors.rs
```diff
@@ -0,0 +1,749 @@
...
+    ///
+    /// By default the header will not be set.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
+    pub fn expose_headers<I>(mut self, headers: I) -> Self
+    where
+        I: Into<AnyOr<Vec<HeaderName>>>,
+    {
+        self.expose_headers = Some(match headers.into().0 {
+            AnyOrInner::Any => "*".parse().unwrap(),
```
DoumanAsh, line 270: `"*"` -> `WILDCARD

tower-http/src/cors.rs
DoumanAsh: I'm not sure if 405 is good for CORS.
I think generally Bad Request would be better.

Semantically 405 indicates that resource doesn't support method at all.
In general I think following is ok:
400 (as more generic might better) or 403 as CORS is kinda auth issue?

DoumanAsh (commented):
We're interesting in having cors for our ongoing migration to axum.
I've put some minor comments

tower-http/src/cors.rs
```diff
@@ -0,0 +1,749 @@
...
+    ///     "*".parse().unwrap(),
+    /// ]);
+    /// ```
+    ///
+    /// By default the header will be set to `*`.
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers
+    pub fn allow_headers<I>(mut self, headers: I) -> Self
+    where
+        I: IntoIterator<Item = HeaderValue>,
```
RabidFire, line 130: This might need to be `IntoIterator<Item = HeaderName>`. I was trying to do something like this, which resulted in an error (`expected struct HeaderName, found struct HeaderValue`):

```
use axum::http::header;

CorsLayer::new().allow_headers(vec![
  header::ORIGIN,
  header::CONTENT_LENGTH,
  header::CONTENT_TYPE,
  "X-Token".parse().unwrap()
])
```

 We previously used actix-cors, and [it seems like they do `Item = TryInto<HeaderName>`](https://docs.rs/actix-cors/0.5.4/src/actix_cors/builder.rs.html#288-289), which feels nicer since I don't have to do `.parse().unwrap()` for custom headers.

davidpdrsn: Oh yeah! Thanks for noticing this! Its just a scary typo. I guess we need to add some more doc tests to make sure things actually work.

tower-http/src/cors.rs
mathroc: I stll don’t think it should be a 4xx at all, in a preflight request, you’re asking "can I do X ?", this request is OK even if the answer is no. and to express "no", you _just_ have to send back the correct CORS headers

RabidFire: Agreed with @mathroc here. Did some digging and it seems like 200 might be the better status code to return.

- There's a [neat summary on Stackoverflow](https://stackoverflow.com/a/46028619) about the Fetch spec requiring that an "ok status" be sent back even in invalid cases. The "ok status" is defined as 2xx in the spec, but I'm guess some browsers interpreted it narrowly as just 200. Therefore, 204 did cause problems with legacy browsers (found an [old Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=845881)) and apparently with Smart TVs as well.
- Express.js allows you to customize the response code for CORS through `[optionsSuccessStatus](https://expressjs.com/en/resources/middleware/cors.html#configuration-options)`. They also [set Content-Length to 0](https://github.com/expressjs/cors/blob/master/lib/index.js#L176-L180) and [default to 204 status](https://github.com/expressjs/cors/blob/c49ca10e92ac07f98a3b06783d3e6ba0ea5b70c7/lib/index.js#L12).
- I ran a test with actix-cors on my current server. It _is_ returning 400, but says "Requested method is not allowed" just as @davidpdrsn confirmed. This seems incorrect to me.

DoumanAsh: There is nothing wrong with using 400.
From server perspective it is bad request.
I suppose you can make exception for options since it doesn't suppose to actually perform something on server so for `OPTIONS` it should be 204, not 200.
For a general case (i.e. any other method) it should be 400

mathroc: @dDoumanAsh, yes we're talking about preflight ( = `OPTIONS`) requests here

davidpdrsn: Haha I'm glad you reached an agreement. I've changed the status to `OK`.

tower-http/src/cors.rs
```diff
@@ -678,7 +678,7 @@ where
                         return ResponseFuture {
                             inner: Kind::Error(Some(
                                 Response::builder()
-                                    .status(StatusCode::METHOD_NOT_ALLOWED)
+                                    .status(StatusCode::OK)
```
jplatte, line 681: The two branches here acn now be combined by changing the above `if` to a match guard.

tower-http/src/cors.rs
```diff
@@ -0,0 +1,798 @@
+//! Middleware which adds headers for [CORS][mdn].
```
DoumanAsh, line 1: It is minor, but would be good to document behavior of middleware (i.e. how it behavers in case of non-`OPTIONS` and `OPTIONS` etc

davidpdrsn: For sure! I wanna do a pass over all the docs when the implementation is correct.

tower-http/src/cors.rs
```diff
@@ -0,0 +1,798 @@
...
+    /// let layer = CorsLayer::new().allow_credentials(true);
+    /// ```
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials
+    pub fn allow_credentials(mut self, allow_credentials: bool) -> Self {
+        self.allow_credentials = Some(allow_credentials.to_string().parse().unwrap());
+        self
+    }
+
+    /// Set the value of the [`Access-Control-Allow-Headers`][mdn] header.
```
DoumanAsh, line 134: By the way, this header is required when client sends `Access-Control-Request-Headers` (normally browser)
It would be good to add note for that to encourage users to set it

tower-http/src/cors.rs
```diff
@@ -0,0 +1,798 @@
...
+    /// Set the value of the [`Access-Control-Allow-Credentials`][mdn] header.
+    ///
+    /// ```
+    /// use tower_http::cors::CorsLayer;
+    ///
+    /// let layer = CorsLayer::new().allow_credentials(true);
+    /// ```
+    ///
+    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials
+    pub fn allow_credentials(mut self, allow_credentials: bool) -> Self {
```
DoumanAsh, line 129: MDN recommends instead of setting it to false, just omit header.
I think better API would be `pub fn allow_credentials(mut self) -> Self` unconditionally setting it to `true`

tower-http/src/cors.rs
```diff
@@ -0,0 +1,798 @@
...
+use tower_service::Service;
+
+/// Layer that applies [`Cors`] which adds headers for [CORS][mdn].
+///
+/// See the [module docs](crate::cors) for an example.
+///
+/// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+#[derive(Debug, Clone)]
+pub struct CorsLayer {
+    allow_credentials: Option<HeaderValue>,
```
DoumanAsh, line 74: I believe more efficient layout would be `HeaderMap`  with `insert` method used to guarantee only single header.
Just clone it around when needed?

tower-http/src/cors.rs
```diff
@@ -0,0 +1,798 @@
...
+        AnyOr(AnyOrInner::Value(headers.into_iter().collect()))
+    }
+}
+
+fn separated_by_commas<I>(into_iter: I) -> HeaderValue
+where
+    I: IntoIterator,
+    I::Item: ToString,
+{
+    let methods = into_iter
```
DoumanAsh, line 361: Suggestion to improve: Create mutable bytes buffer and write it directly, then pass it to `HeaderValue::from_maybe_shared` (simplest way is to crete abstraction that implements `fmt::Write` by extending `BytesMut`)

If it is too much of work I can do PR myself later

DoumanAsh (commented):
Small suggestions to improve efficiency and performance

tower-http/src/cors.rs
davidpdrsn: Interesting! Wanna elaborate why you think that? I'm not very familiar with the internals of `HeaderMap` but since `HeaderValue` is ref counted I imagined this would be faster.

tower-http/src/cors.rs
davidpdrsn: You're right be since these methods are only called during initialization (and not on every `call`) I figured it would be fine.

You're very welcome to submit a follow up PR once this is merge 😊

tower-http/src/cors.rs
davidpdrsn: I generally find methods that take bools more flexible. Makes it easier to set it based on some other logic.

How about this e3c6ac81b2496 instead?

tower-http/src/cors.rs
davidpdrsn: Note added in 9f09fc55bf74b

tower-http/src/cors.rs
DoumanAsh: So let me break down what is `HeaderValue`:
- `Bytes`;
- `bool`

Let's see what is `Bytes`:

```
pub struct Bytes {
    ptr: *const u8,
    len: usize,
    // inlined "trait object"
    data: AtomicPtr<()>,
    vtable: &'static Vtable,
}
```

Basically 4 pointers hence `size_of::<Bytes>() == 32`
Coupled with the fact that each HeaderValue is wrapped inside Option it means 40 bytes for each field (thankfully Option space optimization helps to avoid 48 bytes layout)

Well `HeaderMap` itself is bloated as hell too, but it is only `96` bytes instead of 240 bytes (6 fields by 40 bytes)

davidpdrsn: I see. That makes sense.

I think I'm gonna leave things as they are now. We can always optimize things later. Currently I'm focused on getting the functionality in.

tower-http/src/cors.rs
DoumanAsh: Well it is up to you. MDN only _recommends_ to omit on `false` hence simplifying API is reasonable.
But making it conditional is ok too
It is just from user point of view `allow_credentials(false)` makes no sense because it is default behavior for when you do not send header 😄 
Hence flexibility actually is useless here

tower-http/src/cors.rs
DoumanAsh: Ok, once it is merged I'll make PR.
Since you have to clone layer a lot, it would be beneficial to optimize its layout

tower-http/src/cors.rs
DoumanAsh: Ah yeah, you're right, it is minor thing, but it would serve as good reference in future as creating `Bytes` directly via formatting is far more efficient than creating temporarily allocations. I'll be sure to make PR

tower-http/src/cors.rs
davidpdrsn: Its mostly to allow things like `allow_credentials(std::env::var("SOME_ENV_VAR").is_ok())`. That also matches things like [`DefaultMakeSpan::include_headers`](https://docs.rs/tower-http/0.1.1/tower_http/trace/struct.DefaultMakeSpan.html#method.include_headers).

DoumanAsh: Ah I see, makes sense

davidpdrsn: > The only valid value for this header is true (case-sensitive). If you don't need credentials, omit this header entirely (rather than setting its value to false).

That also feels like stronger than just a recommendation 😅 Definitely something we should follow.

DoumanAsh: Not really, RFC itself doesn't have strong words about it.

From https://datatracker.ietf.org/doc/html/rfc7480#section-5.6
```
By default, browsers do not send cookies when cross origin requests
   are allowed.  Setting the Access-Control-Allow-Credentials header
   field to "true" will send cookies.  Use of the
   Access-Control-Allow-Credentials header field is NOT RECOMMENDED.
```

@davidpdrsn note that it actually says nothing about `false` at all, and only recommends to not send it.

davidpdrsn: Seems the current implementation is the way to go then.

silvioprog:
Thanks a lot for the progress in this PR!

davidpdrsn:
Sorry its taken this long to get merged 😞 Hoping to get it merged and shipped in version 0.1.2 within a week.

tower-http/src/cors.rs
```diff
@@ -0,0 +1,809 @@
...
+            AnyOrInner::Value(headers) => separated_by_commas(headers),
+        });
+        self
+    }
+}
+
+/// Represents a wildcard value (`*`) used with some CORS headers such as
+/// [`CorsLayer::allow_methods`].
+#[derive(Debug, Clone, Copy)]
+pub struct Any;
```
LucioFranco, line 315:
```suggestion
#[derive(Debug, Default, Clone, Copy)]
pub struct Any(());
```

davidpdrsn: I went with a `fn any() -> Any { Any }` instead which is a little easier to type 👉 2f9b5985f49f9

LucioFranco (approved):

PR #117, opened 2021-08-09
compression: Remove `ResBody::Error: Into<BoxError>` bounds

tesaguri:
## Motivation

The `Into<BoxError>` trait bounds on the {de,}compression middlewares are unnecessarily strong requirements since the middlewares return the errors as-is without boxing.

## Solution

Remove the trait bounds.

### DIFF
--- a/tower-http/src/compression/future.rs
+++ b/tower-http/src/compression/future.rs
@@ -1,7 +1,7 @@
 #![allow(unused_imports)]
 
 use super::{body::BodyInner, CompressionBody, Encoding};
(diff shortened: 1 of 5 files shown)


tesaguri:
The CI failure (`cargo-deny check`) seems to be unrelevant to this change.

davidpdrsn (approved):
This seems to have been a copy/paste error. Is indeed not necessary. Thanks for noticing!

The cargo-deny issue will be fixed by updating prost. I've done that in https://github.com/tower-rs/tower-http/pull/112 but think I should probably do it in a separate PR so we can get it merged quickly without blocking this. I'll into that this week.

Issue #116, opened 2021-08-07
Make `AuthorizeRequest::authorize` return a Future


hamza1311:
## Feature Request

Make [`AuthorizeRequest::authorize`](https://docs.rs/tower-http/0.1.1/tower_http/auth/require_authorization/trait.AuthorizeRequest.html#tymethod.authorize) return a Future.

### Motivation

 In it's current state, it is impossible to make async (database) calls in `AuthorizeRequest::authorize`. Such calls are are very important in a real-world application where user data is stored in a database.

### Proposal

Make the given function `async`. This does, however, comes with the downsides of making a trait function `async`.

### Alternatives

Add an async version of the function/trait, similar to [`AsyncFilter`](https://docs.rs/tower/0.4.8/tower/filter/struct.AsyncFilter.html).

hawkw:
> ### Proposal
>
> Make the given function `async`. This does, however, comes with the downsides of making a trait function `async`.
>
> ### Alternatives
> 
> Add an async version of the function/trait, similar to [`AsyncFilter`](https://docs.rs/tower/0.4.8/tower/filter/struct.AsyncFilter.html).

I think the alternative suggestion is the correct thing to do here, for two reasons:
1. Changing the existing `AuthorizeRequest` to be async is a breaking change, but adding a new trait is not breaking
2. In some cases, it's not necessary to do something asynchronous to authorize a request; when that's the case, it would be nice to avoid the overhead of having to return a future. We probably want to have a new `AuthorizeAsync` trait (or something similar).

hamza1311:
How will the `RequireAuthorization` struct accept that? I imagine it would using an enum (PS: I didn't look at implementation of `AsyncFilter`)

I'd be happy to PR this in a couple of days

PR #118, opened 2021-08-09
Add `RequireAuthorizationAsync`

hamza1311:
Fixes https://github.com/tower-rs/tower-http/issues/116

## Motivation

> In it's current state, it is impossible to make async (database) calls in AuthorizeRequest::authorize. Such calls are are very important in a real-world application where user data is stored in a database.

From #116

## Solution

Add a new struct `RequireAuthorizationAsync` which uses `AuthorizeRequestAsync`. `AuthorizeRequestAsync` allows the user to return a `Future`.

### DIFF
--- a/tower-http/src/auth/async_require_authorization.rs
+++ b/tower-http/src/auth/async_require_authorization.rs
@@ -5,7 +5,7 @@
 //! # Example
 //!
 //! ```
-//! use tower_http::auth::{RequireAuthorizationAsyncLayer, AuthorizeRequestAsync};
+//! use tower_http::auth::{AsyncRequireAuthorizationLayer, AsyncAuthorizeRequest};
 //! use hyper::{Request, Response, Body, Error};
 //! use http::{StatusCode, header::AUTHORIZATION};
 //! use tower::{Service, ServiceExt, ServiceBuilder, service_fn};
@@ -14,7 +14,7 @@
 //! #[derive(Clone, Copy)]
 //! struct MyAuth;
 //!
-//! impl AuthorizeRequestAsync for MyAuth {
+//! impl AsyncAuthorizeRequest for MyAuth {
 //!     type Output = UserId;
 //!     type Future = BoxFuture<'static, Option<UserId>>;
 //!     type ResponseBody = Body;
@@ -60,7 +60,7 @@
 //! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
 //! let service = ServiceBuilder::new()
 //!     // Authorize requests using `MyAuth`
-//!     .layer(RequireAuthorizationAsyncLayer::custom(MyAuth))
+//!     .layer(AsyncRequireAuthorizationLayer::new(MyAuth))
 //!     .service_fn(handle);
 //! # Ok(())
 //! # }
@@ -78,35 +78,35 @@ use std::{
 use tower_layer::Layer;
 use tower_service::Service;
 
-/// Layer that applies [`RequireAuthorizationAsync`] which authorizes all requests using the
+/// Layer that applies [`AsyncRequireAuthorization`] which authorizes all requests using the
 /// [`Authorization`] header.
 ///
-/// See the [module docs](crate::auth::require_authorization_async) for an example.
+/// See the [module docs](crate::auth::async_require_authorization) for an example.
 ///
 /// [`Authorization`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
 #[derive(Debug, Clone)]
-pub struct RequireAuthorizationAsyncLayer<T> {
+pub struct AsyncRequireAuthorizationLayer<T> {
     auth: T,
 }
 
-impl<T> RequireAuthorizationAsyncLayer<T>
+impl<T> AsyncRequireAuthorizationLayer<T>
 where
-    T: AuthorizeRequestAsync,
+    T: AsyncAuthorizeRequest,
 {
     /// Authorize requests using a custom scheme.
-    pub fn custom(auth: T) -> RequireAuthorizationAsyncLayer<T> {
+    pub fn new(auth: T) -> AsyncRequireAuthorizationLayer<T> {
         Self { auth }
     }
 }
 
-impl<S, T> Layer<S> for RequireAuthorizationAsyncLayer<T>
+impl<S, T> Layer<S> for AsyncRequireAuthorizationLayer<T>
 where
-    T: Clone,
+    T: Clone + AsyncAuthorizeRequest,
 {
-    type Service = RequireAuthorizationAsync<S, T>;
+    type Service = AsyncRequireAuthorization<S, T>;
 
     fn layer(&self, inner: S) -> Self::Service {
-        RequireAuthorizationAsync::new(inner, self.auth.clone())
+        AsyncRequireAuthorization::new(inner, self.auth.clone())
     }
 }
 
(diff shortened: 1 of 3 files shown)


tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+    ///
+    /// See the [module docs](crate::auth::require_authorization_async) for an example.
+    ///
+    /// [request extension]: https://docs.rs/http/latest/http/struct.Extensions.html
+    /// [JWT]: https://jwt.io
+    #[inline]
+    fn on_authorized<B>(&mut self, _request: &mut Request<B>, _output: Self::Output) {}
+
+    /// Create the response for an unauthorized request.
+    fn unauthorized_response<B>(&mut self, request: &Request<B>) -> Response<Self::ResponseBody>;
```
sidred, line 269: Should there be a way to pass the error from authorize function to this? The use case for that is, if there is a db or redis call in authorize function, then on network error we should be able to send a 5xx error instead of a 401 error.

hamza1311: I agree with you but IMO that's outside the scope of this PR. This PR adds async version of `RequireAuthorization`. Such a change would also need to made to `RequireAuthorization` in addition to `RequireAuthorizationAsync`.

davidpdrsn:
Just a quick heads up. I haven't forgotten about this. Things have just been busy with work and axum.

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+//! impl AuthorizeRequestAsync for MyAuth {
+//!     type Output = UserId;
+//!     type Future = BoxFuture<'static, Option<UserId>>;
+//!     type ResponseBody = Body;
+//!
+//!     fn authorize<B>(&mut self, request: &Request<B>) -> Self::Future {
+//!         # Box::pin(async {
+//!             // ...
+//!             # None
+//!         })
```
davidpdrsn, line 26:
```suggestion
//!         Box::pin(async {
//!             // ...
//!             # None
//!         })
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+use tower_service::Service;
+
+/// Layer that applies [`RequireAuthorizationAsync`] which authorizes all requests using the
+/// [`Authorization`] header.
+///
+/// See the [module docs](crate::auth::require_authorization_async) for an example.
+///
+/// [`Authorization`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
+#[derive(Debug, Clone)]
+pub struct RequireAuthorizationAsyncLayer<T> {
```
davidpdrsn, line 88:
```suggestion
pub struct AsyncRequireAuthorizationLayer<T> {
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Layer that applies [`RequireAuthorizationAsync`] which authorizes all requests using the
+/// [`Authorization`] header.
+///
+/// See the [module docs](crate::auth::require_authorization_async) for an example.
```
davidpdrsn, line 84: I think we should `Async` a prefix. That way it matches `Filter` and `AsyncFilter` in tower.

```suggestion
/// See the [module docs](crate::auth::async_require_authorization) for an example.
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+pub struct RequireAuthorizationAsyncLayer<T> {
+    auth: T,
+}
+
+impl<T> RequireAuthorizationAsyncLayer<T>
+where
+    T: AuthorizeRequestAsync,
+{
+    /// Authorize requests using a custom scheme.
+    pub fn custom(auth: T) -> RequireAuthorizationAsyncLayer<T> {
```
davidpdrsn, line 97: Since other schemes don't make sense (as they're aren't async) I think calling this `new` is better.

```suggestion
    /// Authorize requests using the given scheme.
    pub fn new(auth: T) -> RequireAuthorizationAsyncLayer<T> {
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+}
+
+/// Middleware that authorizes all requests using the [`Authorization`] header.
+///
+/// See the [module docs](crate::auth::require_authorization_async) for an example.
+///
+/// [`Authorization`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
+#[derive(Clone, Debug)]
+pub struct RequireAuthorizationAsync<S, T> {
```
davidpdrsn, line 119:
```suggestion
pub struct AsyncRequireAuthorization<S, T> {
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+}
+
+impl<S, T> RequireAuthorizationAsync<S, T>
+where
+    T: AuthorizeRequestAsync,
+{
+    /// Authorize requests using a custom scheme.
+    ///
+    /// The `Authorization` header is required to have the value provided.
+    pub fn custom(inner: S, auth: T) -> RequireAuthorizationAsync<S, T> {
```
davidpdrsn, line 139:
```suggestion
    /// Authorize requests using the given scheme.
    pub fn new(inner: S, auth: T) -> RequireAuthorizationAsync<S, T> {
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+                StateProj::Authorized { fut } => {
+                    return fut.poll(cx);
+                }
+            }
+        }
+    }
+}
+
+/// Trait for authorizing requests.
+pub trait AuthorizeRequestAsync {
```
davidpdrsn, line 236:
```suggestion
pub trait AsyncAuthorizeRequest {
```

tower-http/src/auth/require_authorization_async.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// The Future type returned by `authorize`
+    type Future: Future<Output = Option<Self::Output>>;
+
+    /// The body type used for responses to unauthorized requests.
+    type ResponseBody: Body;
+
+    /// Authorize the request.
+    ///
+    /// If `Some(_)` is returned then the request is allowed through, otherwise not.
```
davidpdrsn, line 250:
```suggestion
    /// If the future resolves to `Some(_)` then the request is allowed through, otherwise not.
```

tower-http/src/auth/require_authorization_async.rs
davidpdrsn: I think mapping errors to responses should be handled by a separate middleware. So I think this is fine. We can always consider adding error handling to these middleware for 0.2, if someone requests it.

davidpdrsn (approved):
Sorry the long delay 😞

I think this look great! Just a couple of small comments.

davidpdrsn (changes requested):
Sorry didn't mean to approve quite yet.

tower-http/src/auth/require_authorization_async.rs
hamza1311: ...and that person will be me. I was using it and ran into a case where it's impossible to return 401/403 depending upon the request. We need a better way here.

I was gonna make an issue but life got in the way...

davidpdrsn: Hehe alright. Do you wanna address it here or should we move forward with this as is, and fix it for 0.2? I'm fine with either actually.

hamza1311:
Fixed the CI errors, should be good to go.

tower-http/src/auth/async_require_authorization.rs
```diff
@@ -0,0 +1,353 @@
...
+    type Service = AsyncRequireAuthorization<S, T>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        AsyncRequireAuthorization::new(inner, self.auth.clone())
+    }
+}
+
+/// Middleware that authorizes all requests using the [`Authorization`] header.
+///
+/// See the [module docs](crate::auth::require_authorization_async) for an example.
```
davidpdrsn, line 115:
```suggestion
/// See the [module docs](crate::auth::async_require_authorization) for an example.
```

tower-http/src/auth/async_require_authorization.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Callback for when a request has been successfully authorized.
+    ///
+    /// For example this allows you to save `Self::Output` in a [request extension][] to make it
+    /// available to services further down the stack. This could for example be the "claims" for a
+    /// valid [JWT].
+    ///
+    /// Defaults to doing nothing.
+    ///
+    /// See the [module docs](crate::auth::require_authorization_async) for an example.
```
davidpdrsn, line 257:
```suggestion
    /// See the [module docs](crate::auth::async_require_authorization) for an example.
```

tower-http/src/auth/require_authorization_async.rs
hamza1311: Let's fix it in some other PR. It's a breaking change. I don't have the time for it right now anyway.

davidpdrsn:
Hm it seems CI doesn't actually check the docs for broken links 🤔 I'll fix that separately.

hamza1311:
> Hm it seems CI doesn't actually check the docs for broken links thinking I'll fix that separately.

Fixed the links

davidpdrsn (approved):

davidpdrsn:
Thanks for working on this!

PR #122, opened 2021-08-19
Fix ServeDir Bug

### LINKED ISSUE #121: ServeDir Bug
### Version

0.1.1

### Platform

`Linux 4.19.0-17-amd64 #1 SMP Debian 4.19.194-1 (2021-06-10) x86_64 GNU/Linux`

### Description

When `append_index_html_on_directories` is set to `false`, requests to a directory endpoint will return `200` response with a `Body` that will definitely fail because directories can't be read. Requests to a directory endpoint should always return `404` if `append_index_html_on_directories` is set to `false`.

programatik29:
## Motivation

Fixes #121.

## Solution

With this change requests to a directory endpoint will always return `404` if `append_index_html_on_directories` is set to `false`.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-None.
+- Fix a bug which happens when `append_index_html_on_directories` is set to `false` in `ServeDir`.
 
 ## Breaking changes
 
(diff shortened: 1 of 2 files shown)


davidpdrsn:
I think we should have a test for it as well.

programatik29:
> I think we should have a test for it as well.

Done.

programatik29:
Couldn't find a good name for the test. Any suggestions?

davidpdrsn:
I think its fine 😊

The fix looks good to me. @tesaguri do you wanna give it a quick look as well?

tesaguri:
Looks good to me too.

davidpdrsn (approved):

Issue #123, opened 2021-08-19
`TraceLayer` only logs request method and path, if debug logging is enabled


hannobraun:
### Version

v0.1.1

### Platform

Linux 5.13.10-arch1-1 #1 SMP PREEMPT Thu, 12 Aug 2021 21:59:14 +0000 x86_64 GNU/Linux

### Description

I'm configuring a `TraceLayer` like this:

``` rust
TraceLayer::new_for_http()
    .on_request(trace::DefaultOnRequest::new().level(Level::INFO))
    .on_response(trace::DefaultOnResponse::new().level(Level::INFO))
```

When executing my program with `RUST_LOG=info cargo run`, I get the following log output:

```
INFO tower_http::trace::on_request: started processing request
INFO tower_http::trace::on_response: finished processing request latency=0 ms status=200
```

The requests are logged, but the request method and path are missing.

If I run `RUST_LOG=debug cargo run`, both are there:

```
INFO request{method=GET uri=/ version=HTTP/1.1}: tower_http::trace::on_request: started processing request
INFO request{method=GET uri=/ version=HTTP/1.1}: tower_http::trace::on_response: finished processing request latency=1 ms status=200
```

I believe that the reason for that is that the `make_span` method is using `tracing::debug_span`, regardless of any other configuration: https://github.com/tower-rs/tower-http/blob/e1252c2b39d2f8333339e917d33b1b37915b113b/tower-http/src/trace/make_span.rs#L58-L75

I'm not sure what is supposed to be done here. The straight-forward fix is to add a `Level` field to `DefaultMakeSpan` (along with a method to set it), but that makes no sense to me, API-wise. If I configure my requests to be logged with INFO, I don't expect to need extra steps to get basic information with my log messages.

hannobraun:
I've submitted #124 to address this.

hannobraun:
As @davidbarsky pointed out in #124, the existing `TraceLayer::make_span_with` is enough to work around this issue, as you can pass your own `MakeSpan` implementation there and aren't beholden to what `DefaultMakeSpan` does.

This addresses my problem, but maybe this issue should stay open to track the weirdness of the API and resulting potential for confusion? I don't know. I'm leaving it open and letting the maintainers decide.

davidpdrsn:
I think you're right that the API is inconsistent since DefaultMakeSpan doesn't have a builder method for setting the span level. I'll review your PR tomorrow 😊

PR #124, opened 2021-08-19
Allow users to control level of Tracing spans

hannobraun:
## Motivation

I want to see the request method and path in my log message. See #123 for more information.

## Solution

This is a minimally invasive solution for #123: It allows the user to configure the level for request spans. While this is far from ideal API-wise, it's a quick fix. I don't think I know enough about how Tracing is supposed to be used to come up with something better right now (that spans have associated levels in the first place seems wrong to me, but I'm sure I'm missing something).


davidbarsky:
(I'm putting on my "tracing contibutor" hat on.)

davidbarsky:
(I'm putting on my "tracing contributor" hat on.)

Anyways! I noticed that in #123, you mentioned that you're creating the `TraceLayer`  thusly:

```rust
TraceLayer::new_for_http()
    .on_request(trace::DefaultOnRequest::new().level(Level::INFO))
    .on_response(trace::DefaultOnResponse::new().level(Level::INFO))
```

It seems to me that [`TraceLayer::make_span_with`](https://docs.rs/tower-http/0.1.1/tower_http/trace/struct.TraceLayer.html#method.new_for_http)—as documented [here](https://docs.rs/tower-http/0.1.1/tower_http/trace/index.html#recording-fields-on-the-span))—should address your use-case _without_ needing to introduce a new, slightly different API. Am I incorrect?

hannobraun:
Oh, you're right! I was so focused on the `DefaultMakeSpan`, I somehow missed that you can pass your own `MakeSpan` implementation there. That addresses my use case.

Thank you!

davidpdrsn:
I've opened. I still think it would be nice to address this inconsistency. Will review tomorrow.

tower-http/src/trace/on_request.rs
```diff
@@ -60,10 +60,15 @@ impl DefaultOnRequest {
 
     /// Set the [`Level`] used for [tracing events].
     ///
+    /// Please note that while this will set the level for the tracing events
+    /// themselves, it might cause them to lack expected information, like
+    /// request method or path. You can address this using
+    /// [`DefaultMakeSpan::level`].
+    ///
```
davidpdrsn, line 67: This bit feels redundant to me. The default span fields are already shown in the example output at the top of [the docs](https://docs.rs/tower-http/0.1.1/tower_http/trace/index.html). I kinda think we should remove this paragraph. Same goes for the change in `tower-http/src/trace/on_response.rs`.

davidpdrsn (changes requested):
Sorry the delay!

I think this looks except for one small thing wrt the docs.

Also wanna add a line in the changelog?

You should also merge/rebase master to fix CI.

hannobraun:
Thanks for the review and also sorry for the delay! I'm on vacation. I'll take care of this at some point during the next weeks.

davidpdrsn:
No worries :)

tower-http/src/trace/on_request.rs
hannobraun: Hey @davidpdrsn, I don't understand what you mean here. It's true that the span fields are shown in the example output in the docs. But I was very confused when I used these `level` methods to set the log level for request/response, and then saw log messages that **did not** match this example output. I added the paragraph to address this point of confusion.

hannobraun:
Rebased on top of master (no changes were necessary). Added changelog entry as requested. Did not address the review comment, as I disagree with the request (see comment there).

davidpdrsn:
@hannobraun Sorry for the delay on this. I think it looks good as is 😊 Do you wanna merge/rebase master? I would like to include this in 0.1.2 which I'm hoping to release next week.

davidpdrsn (approved):

davidpdrsn:
@hannobraun Thanks for working on this! Sorry for taking so long with reviews.

hannobraun:
Looks like you got it sorted before I had a chance to look into it. Thanks!

And don't worry about the time it took. There are only so many hours in the day, after all :grin:

PR #131, opened 2021-08-29
Remove duplicate code from compression/layer.rs

Madoshakalaka:
struct `Compression` and struct `CompressionLayer` has a exactly same set of methods.

[CompressionLayer code](https://github.com/tower-rs/tower-http/blob/e25d4f099e792e38daafac74bf6238b3f855774a/tower-http/src/compression/layer.rs#L32-L77)

[Compression code](https://github.com/tower-rs/tower-http/blob/e25d4f099e792e38daafac74bf6238b3f855774a/tower-http/src/compression/service.rs#L38-L84)

It turns out `CompressionLayer`'s set of methods are not used anywhere and they are not in the documentation or any API. Removing them doesn't impact things and tidies up the code.

In fact, `CompressionLayer::new()` returns a `Compression` instance and those methods only need to exist there.

### DIFF
--- a/tower-http/src/compression/layer.rs
+++ b/tower-http/src/compression/layer.rs
@@ -1,5 +1,4 @@
 use super::Compression;
-use crate::compression_utils::AcceptEncoding;
 use tower_layer::Layer;
 
 /// Compress response bodies of the underlying service.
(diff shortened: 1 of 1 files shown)


davidpdrsn:
I don't understand what you mean. They are part of the public api https://docs.rs/tower-http/0.1.1/tower_http/compression/struct.CompressionLayer.html

davidpdrsn:
The layers config isn't used when apply it to the service. That's a bug but removing the methods isn't right.

Madoshakalaka:
@davidpdrsn I'll close this then. I'll produce a bug fix later!

davidpdrsn:
Cool. Thanks for noticing!

PR #132, opened 2021-08-29
Bug fix: [de]compression layers ignore configuration

Madoshakalaka:
* Decompression layer and compression layer now accepts encoding configuration

* Added regression test to make sure compression layer registers configuration

whoa, so compression layers and decompression layers just ignored those `.set_br`/`.no_br` calls...I fixed it!

### DIFF
--- a/tower-http/src/compression/layer.rs
+++ b/tower-http/src/compression/layer.rs
@@ -18,7 +18,10 @@ impl<S> Layer<S> for CompressionLayer {
     type Service = Compression<S>;
 
     fn layer(&self, inner: S) -> Self::Service {
-        Compression::new(inner)
+        let mut outer = Compression::new(inner);
+        println!("layer's accept: {:?}", self.accept);
+        outer.accept = self.accept;
+        outer
     }
 }
 
@@ -31,7 +34,7 @@ impl CompressionLayer {
     /// Sets whether to enable the gzip encoding.
     #[cfg(feature = "compression-gzip")]
     #[cfg_attr(docsrs, doc(cfg(feature = "compression-gzip")))]
-    pub fn gzip(self, enable: bool) -> Self {
+    pub fn gzip(mut self, enable: bool) -> Self {
         self.accept.set_gzip(enable);
         self
     }
@@ -39,7 +42,7 @@ impl CompressionLayer {
     /// Sets whether to enable the Deflate encoding.
     #[cfg(feature = "compression-deflate")]
     #[cfg_attr(docsrs, doc(cfg(feature = "compression-deflate")))]
-    pub fn deflate(self, enable: bool) -> Self {
+    pub fn deflate(mut self, enable: bool) -> Self {
         self.accept.set_deflate(enable);
         self
     }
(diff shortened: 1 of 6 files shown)


Madoshakalaka:
whoops, forgot to remove those `println!`'s I added while debugging

tower-http/src/compression/layer.rs
```diff
@@ -31,48 +33,145 @@ impl CompressionLayer {
     /// Sets whether to enable the gzip encoding.
     #[cfg(feature = "compression-gzip")]
     #[cfg_attr(docsrs, doc(cfg(feature = "compression-gzip")))]
-    pub fn gzip(self, enable: bool) -> Self {
+    pub fn gzip(mut self, enable: bool) -> Self {
```
davidpdrsn, line 36: Why are all these `mut`s necessary?

Madoshakalaka: This is the bug exactly. The next line `self.accept.set_gzip(enable)` wouldn't do anything. Because `set_gzip` was an adaptor instead of a setter, it returns a new AcceptEncoing here and is immediately dropped. Thus the configuration fails to register.

See below code change in `AcceptEncoding`'s struct. `AcceptEncoing::set_gzip` is now a setter instead of an adaptor, see below. the mut here is necessary because the next line mutates self.

tower-http/src/compression_utils.rs
```diff
@@ -78,21 +78,18 @@ impl AcceptEncoding {
     }
 
     #[allow(dead_code)]
-    pub(crate) fn set_gzip(mut self, enable: bool) -> Self {
+    pub(crate) fn set_gzip(& mut self, enable: bool) {
```
davidpdrsn, line 81: Seems you didn't run `cargo rustfmt`

tower-http/src/compression/layer.rs
davidpdrsn: Makes sense.

davidpdrsn (commented):
Thanks for fixing this! I think the changes look good, just a couple of minor nit picks.

You might also wanna merge/rebase master to fix CI.

Should probably also mention this fix in the changelog.

tower-http/src/compression/layer.rs
```diff
@@ -18,7 +18,9 @@ impl<S> Layer<S> for CompressionLayer {
     type Service = Compression<S>;
 
     fn layer(&self, inner: S) -> Self::Service {
-        Compression::new(inner)
+        let mut outer = Compression::new(inner);
+        outer.accept = self.accept;
+        outer
```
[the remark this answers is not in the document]
Madoshakalaka: yep!

tower-http/src/compression/mod.rs
```diff
@@ -166,7 +166,10 @@ fn encodings(headers: &HeaderMap, accept: AcceptEncoding) -> Vec<(Encoding, f32)
 
             Some((encoding, qval))
         })
-        .collect::<Vec<(Encoding, f32)>>()
+        .collect::<Vec<(Encoding, f32)>>();
+
+
+    ret
```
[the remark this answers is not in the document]
Madoshakalaka: sorry I don't know how it sneaked in there

tower-http/src/compression_utils.rs
Madoshakalaka: I have just run it!

davidpdrsn (approved):

PR #137, opened 2021-09-02
ServeDir: reuse buf and reponse Stream, optimize the default chunk buf size

ttys3:
~~we need wait tokio release the next version of `tokio-util` since this PR depends on the commit https://github.com/tokio-rs/tokio/commit/d0dd74a0583b3ed7c8c7f68a4279fe9709bbf158~~

`tokio-util` 0.6.8 has been released (https://crates.io/crates/tokio-util/0.6.8)

## Motivation

With this PR, the `ServeDir` file sending speed up from `6MiB/s` to `1318M/s` on my machine (target `release` build).

The default buf size changed to `64KiB`, which is a reasonable value in some other popular Rust web framework. I think we are ok with this value.

## Solution

1. reuse the buf ( do not `new` on every `poll`)
2. change the default buffer chunk size to a rather reasonable value
3. add `with_capacity()` method allow to customize the buffer capacity (currently this method is internal use only)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,3 +4,7 @@ members = [
   "examples/warp-key-value-store",
   "examples/tonic-key-value-store",
 ]
+
+[patch.crates-io]
+tokio = { git = "https://github.com/tokio-rs/tokio.git", branch = "master"}
+tokio-util = { git = "https://github.com/tokio-rs/tokio.git", branch = "master" }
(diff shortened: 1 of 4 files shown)


davidpdrsn (commented):
Sorry for taking so long to review this 😞

But thanks a lot of fixing it!

davidpdrsn (approved):

PR #140, opened 2021-09-06
fix: Don't recompress HTTP responses

### LINKED ISSUE #138: Compression layer recompresses responses
### Version

v0.1.1

### Description

If a layer below the compression layer returns a response whose body is already compressed, the compression-layer will happily re-compress the already compressed bytes, because it does not check the `Content-Encoding` before applying its body transformation.

https://github.com/tower-rs/tower-http/blob/36157ce3ed7cbf30d2c8016ec8054b703308c585/tower-http/src/compression/future.rs#L34-L67

This behavior is fine, if the content that passes through the layer is known to never be compressed already. In a complex application where the data comes from multiple sources, this might not always be the case. Therefore, the layer should probably check whether there already is a `Content-Encoding` on the response, and not apply any further transformations in this case. This would allow using the compression layer as a kind of "catch-all", at the outer edge of the layer stack.

If you agree I'm happy to send a PR 👍. I'd probably integrate this check into `supports_transparent_compression`.

NeoLegends:
## Motivation

Closes #138 
## Solution

extended the `supports_transparent_compression` function to check for a `content-encoding` and added a test

### DIFF
--- a/tower-http/src/compression/mod.rs
+++ b/tower-http/src/compression/mod.rs
@@ -172,11 +172,13 @@ fn encodings(headers: &HeaderMap, accept: AcceptEncoding) -> Vec<(Encoding, f32)
 #[cfg(test)]
 mod tests {
     use super::*;
+    use async_compression::tokio::write::{BrotliDecoder, BrotliEncoder};
     use bytes::BytesMut;
     use flate2::read::GzDecoder;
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks for fixing this 👌

PR #141, opened 2021-09-06
feat: Allow specifying a minimum body size for compression

### LINKED ISSUE #139: Only compress responses above a certain size
### Motivation

This is something I've seen in other web servers, where compression is only turned on for responses of e. g. `>= 1kB`. In particular, there's no use compressing an empty response.

### Proposal

In the compression layer, check the body's `size_hint()`, and do not compress if there is an exact size hint less than a configured value.

I'd say if the size hint is not exact we default to compressing responses in any case.

### Alternatives

Don't care and keep the current behavior.

NeoLegends:
## Motivation

Closes #139 

## Solution

The layer now inspects the response body's `size_hint` (only if exact) and `content-length` header to measure the size of the content. If it's below a preconfigured threshold (default: 32 bytes) compression won't happen.

Now, the only remaining question I have is what the default should be. I think having a default `> 0` is very helpful, because it stops the layer from wasting time compressing 0-sized and "very-small" responses. The only question is, what's "very-small"? nginx uses a default of 20 bytes, and I used 32 because it's a power of two and close to 20. Alternatives I see:
- 0 bytes, keep the behavior identical to the current implementation. Users need to actively "do something" though to improve behavior on empty responses, whereas with a `> 0` default they get that "for free".
- 1 byte, just to save us from compressing 0-sized responses
- 256 / 512 / 1024, a bit larger, low risk of wasting bandwith on http / text files

~~Lastly, do we really need a `u64` for the limit? A `u16` would probably suffice, given that it would be odd to have a minimum size above 65k.~~ Did that now.

### DIFF
--- a/tower-http/src/compression/layer.rs
+++ b/tower-http/src/compression/layer.rs
@@ -11,7 +11,7 @@ use tower_layer::Layer;
 #[derive(Clone, Debug, Default)]
 pub struct CompressionLayer {
     accept: AcceptEncoding,
-    min_size: u64,
+    min_size: u16,
 }
 
 impl<S> Layer<S> for CompressionLayer {
@@ -89,7 +89,7 @@ impl CompressionLayer {
     /// Passing `0` makes the layer compress every response.
     ///
     /// The default is 32 bytes.
-    pub fn min_size(mut self, min: u64) -> Self {
+    pub fn min_size(mut self, min: u16) -> Self {
         self.min_size = min;
         self
     }
--- a/tower-http/src/compression/layer.rs
+++ b/tower-http/src/compression/layer.rs
@@ -1,4 +1,4 @@
-use super::Compression;
+use super::{Compression, MIN_SIZE_DEFAULT};
 use crate::compression_utils::AcceptEncoding;
 use tower_layer::Layer;
 
@@ -10,8 +10,8 @@ use tower_layer::Layer;
 /// See the [module docs](crate::compression) for more details.
 #[derive(Clone, Debug, Default)]
 pub struct CompressionLayer {
-    _priv: (),
     accept: AcceptEncoding,
+    min_size: u64,
 }
 
 impl<S> Layer<S> for CompressionLayer {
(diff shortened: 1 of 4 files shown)


davidpdrsn (approved):
I think it looks good. Thanks for doing this!

davidpdrsn:
@NeoLegends Wanna fix the merge conflicts?

tower-http/src/compression/layer.rs
```diff
@@ -78,6 +88,21 @@ impl CompressionLayer {
...
+    /// bodies.
+    ///
+    /// Any response smaller than `min` will not be compressed. The response size
+    /// is determined by inspecting [`Body::size_hint()`] and the `content-length`
+    /// header.
+    ///
+    /// Passing `0` makes the layer compress every response.
+    ///
+    /// The default is 32 bytes.
+    pub fn min_size(mut self, min: u16) -> Self {
```
davidpdrsn, line 102: What do you think about making this an `Option<u16>` and using `None` to mean "compression everything"? I generally find that a bit more explicit.

davidpdrsn (commented):
I think it looks good! Thanks for working on it. I only had one API question.

Just a heads up, once https://github.com/tower-rs/tower-http/pull/156 is merged I think you'll get some merge conflicts when rebasing on `master`.

tower-http/src/compression/layer.rs
davidpdrsn: Should probably be `impl Into<Option<u16>>` which will allow passing `u16`s directly and thus a bit more ergonomic.

davidpdrsn (changes requested):
Actually I'm thinking we can use some of the API introduced by https://github.com/tower-rs/tower-http/pull/155 here as well. I think we should work on it there and circle back when its done. So changing my review to "request changes" just as a reminder to myself not to merge this yet.

davidpdrsn:
We have decided to include this as part of https://github.com/tower-rs/tower-http/pull/172. You're welcome to have a look at that and leave feedback on the API 😊

Thanks for your work on this!

PR #142, opened 2021-09-06
add max-age to ServeDir

c5soft:
It's important for static file service to support cache controlling, so i start a step by add max-age to ServeDir

### DIFF
--- a/tower-http/src/services/fs/serve_dir.rs
+++ b/tower-http/src/services/fs/serve_dir.rs
@@ -1,7 +1,7 @@
 use super::AsyncReadBody;
 use bytes::Bytes;
 use futures_util::ready;
-use http::{header, HeaderValue, Request, Response, StatusCode, Uri};
+use http::{header, HeaderMap, HeaderValue, Request, Response, StatusCode, Uri};
 use http_body::{combinators::BoxBody, Body, Empty};
 use percent_encoding::percent_decode;
 use std::{
@@ -28,6 +28,7 @@ use tower_service::Service;
 pub struct ServeDir {
     base: PathBuf,
     append_index_html_on_directories: bool,
+    max_age: u64,
 }
 
 impl ServeDir {
@@ -39,6 +40,7 @@ impl ServeDir {
         Self {
             base,
             append_index_html_on_directories: true,
+            max_age: 5 * 60,
         }
     }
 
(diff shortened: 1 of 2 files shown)


tower-http/src/services/fs/serve_dir.rs
```diff
@@ -113,7 +125,14 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
                 });
 
             let file = File::open(full_path).await?;
-            Ok(Output::File(file, mime))
+            let mut headers: HeaderMap<HeaderValue> = HeaderMap::new();
+            headers.insert(header::CONTENT_TYPE, mime);
+            headers.insert(
+                header::CACHE_CONTROL,
+                HeaderValue::from_str(&format!("max-age={}", max_age)).unwrap(),
+            );
```
davidpdrsn, line 133: I think it would be nicer to just have the two headers as separate fields on `Output::File`. Then we don't need to allocate a `HeaderMap` and can add the headers one by one in the future.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -212,7 +231,8 @@ impl Future for ResponseFuture {
                 let body = ResponseBody(body);
 
                 let mut res = Response::new(body);
-                res.headers_mut().insert(header::CONTENT_TYPE, mime);
+                //res.headers_mut().insert(header::CONTENT_TYPE, headers);
```
davidpdrsn, line 234: Should probably just remove this.

```suggestion
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -51,6 +53,15 @@ impl ServeDir {
         self.append_index_html_on_directories = append;
         self
     }
+    /// Cache-Control:max-age=300.
+    ///
+    /// This speeds up user's feeling for static sites, especially large files.
+    ///
+    /// Defaults to 300 seconds.
```
davidpdrsn, line 60: I think these docs should be phrased differently. What will setting this actually make the browser do?

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -51,6 +53,15 @@ impl ServeDir {
...
+    pub fn max_age(mut self, seconds: u64) -> Self {
```
davidpdrsn, line 61: I think accepting a `std::time::Duration` is nicer and then mentioning the docs that it will be rounded down to the nearest second. Because thats what `Duration::as_secs` does.

```suggestion
    pub fn max_age(mut self, duration: Duration) -> Self {
```

davidpdrsn (changes requested):
Thanks for working on this! It makes to support.

davidpdrsn:
@c5soft Do you still wanna drive this home?

c5soft:
Rewrited with a new PR

c5soft:
Rewrited with a new PR

PR #143, opened 2021-09-10
Improve set header

### LINKED ISSUE #127: SetResponseHeaderLayer et al are hard to use with literal HeaderValues
### Version

0.1.1

### Platform
Linux

### Description

SetResponseHeaderLayer (and associated functions) use the MakeHeaderValue trait for the value. This is implemented on `HeaderValue` itself (for setting a constant) but I'm at a loss for how to use it because of the unrestricted type `T` in its generic signature

Given this Axum code:
```rust
    let hostname = HeaderValue::from_static("example");

    let app = Router::new()
        .route("/v1/ok", get(handle_ok))
        .layer(tower_http::set_header::SetResponseHeaderLayer::overriding(
            "x-hostname".parse().unwrap(),
            hostname,
        ));
```

Compilation fails with:

```
   --> src/main.rs:198:16
    |
198 |         .layer(tower_http::set_header::SetResponseHeaderLayer::overriding(
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `T`
```

I have no idea how to pass this type. It's not constrained in the docs and I don't even know what it //is//.

jplatte:
## Motivation

Makes the code more maintainable, and docs neater (providing rustdoc with a clear tagline / rest of the docs for overview pages).

## Solution

Refactor stuff.

## Other

I didn't rename `T` in `Set{Request,Response}HeaderLayer` to `ReqBody` / `ResBody` here, but I think that might also make sense.

All of this refactoring happened in an attempt to fix #127, which I still believe is possible but I've spent enough time on this for now.

### DIFF
--- a/tower-http/src/lib.rs
+++ b/tower-http/src/lib.rs
@@ -99,7 +99,7 @@
 //! };
 //! use tower::{ServiceBuilder, Service, ServiceExt};
 //! use hyper::Body;
-//! use http::{Request, Response, HeaderValue, header::USER_AGENT};
+//! use http::{Request, HeaderValue, header::USER_AGENT};
 //!
 //! #[tokio::main]
 //! async fn main() {
--- a/tower-http/src/services/redirect.rs
+++ b/tower-http/src/services/redirect.rs
@@ -47,6 +47,7 @@ use tower_service::Service;
 pub struct Redirect<ResBody> {
     status_code: StatusCode,
     location: HeaderValue,
+    // Covariant over ResBody, no dropping of ResBody
     _marker: PhantomData<fn() -> ResBody>,
 }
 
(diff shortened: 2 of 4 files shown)


jplatte:
The last commit is a breaking change due to adding / strengthening trait bounds. Other people could have replicated the strange previous bound (and lack of `Clone` bound) in their own generic code, which would then break with this change.

davidpdrsn (commented):
Nice!

What do you think about leaving the bounds as they were and changing them for 0.2?

jplatte:
Sure. Do you want to merge the rest of this PR on the command line or should I split the PR?

davidpdrsn:
I would appreciate it if you would split it. Then I can merge directly from GH.

davidpdrsn (approved):
Thanks!

PR #144, opened 2021-09-17
Fix build warning

hannobraun:
## Motivation

The build should be warning-free.

## Solution

Fixed warning about renamed lint by using the new name.


davidpdrsn (approved):
Thanks!

PR #145, opened 2021-09-17
ServeDir & ServeFile last modified headers handling

zenria:
~Note: This PR also includes https://github.com/tower-rs/tower-http/pull/137, I will modify it when it will be merged~

## Motivation

ServeDir & ServeFile should respect http caching best practice by setting `Last-Modified` response header and handling `If-Modified_since` & `If-Unmodified-Since` requests headers.

## Solution

The `Last-Modified` header is added on responses of ServeFile & ServeDir. 

`If-Modified_since` & `If-Unmodified-Since` requests headers are handled so the service will respond `NOT_MODIFIED` or `PRECONDITION_FAILED` depending on values set into the headers.

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -35,6 +35,7 @@ tokio = { version = "1", optional = true, default_features = false }
 tokio-util = { version = "0.6.8", optional = true, default_features = false, features = ["io"] }
 tower = { version = "0.4.1", optional = true }
 tracing = { version = "0.1", default_features = false, optional = true }
+httpdate = { version = "1.0", optional = true }
 
 [dev-dependencies]
 bytes = "1"
@@ -74,7 +75,7 @@ add-extension = []
 auth = ["base64"]
 cors = []
 follow-redirect = ["iri-string", "tower/util"]
-fs = ["tokio/fs", "tokio-util/io", "tokio/io-util", "mime_guess", "mime", "percent-encoding"]
+fs = ["tokio/fs", "tokio-util/io", "tokio/io-util", "mime_guess", "mime", "percent-encoding", "httpdate"]
 map-request-body = []
 map-response-body = []
 metrics = ["tokio/time"]
--- a/tower-http/src/services/fs/mod.rs
+++ b/tower-http/src/services/fs/mod.rs
@@ -226,8 +226,7 @@ impl IfModifiedSince {
     fn from_header_value(value: &http::header::HeaderValue) -> Option<IfModifiedSince> {
         String::from_utf8(value.as_bytes().to_vec())
             .ok()
-            .map(|value| httpdate::parse_http_date(&value).ok())
-            .flatten()
+            .and_then(|value| httpdate::parse_http_date(&value).ok())
             .map(|time| IfModifiedSince(time.into()))
     }
 }
--- a/tower-http/src/services/fs/mod.rs
+++ b/tower-http/src/services/fs/mod.rs
@@ -248,7 +248,7 @@ impl IfUnmodifiedSince {
 }
 
 fn check_modified_headers(
-    modified: &Option<LastModified>,
+    modified: Option<&LastModified>,
     if_unmodified_since: Option<IfUnmodifiedSince>,
     if_modified_since: Option<IfModifiedSince>,
 ) -> Option<StatusCode> {
--- a/tower-http/src/services/fs/mod.rs
+++ b/tower-http/src/services/fs/mod.rs
@@ -1,8 +1,10 @@
 //! File system related services.
 
 use bytes::Bytes;
-use http::{HeaderMap, Response, StatusCode};
+use futures_util::Stream;
+use http::{HeaderMap, HeaderValue, Response, StatusCode};
 use http_body::{combinators::BoxBody, Body, Empty};
+use httpdate::HttpDate;
 use pin_project_lite::pin_project;
 use std::fs::Metadata;
 use std::{ffi::OsStr, path::PathBuf};
@@ -10,12 +12,11 @@ use std::{
     io,
     pin::Pin,
     task::{Context, Poll},
+    time::SystemTime,
 };
 use tokio::fs::File;
 use tokio::io::{AsyncRead, AsyncReadExt, Take};
 use tokio_util::io::ReaderStream;
-
-use futures_util::Stream;
 
 mod serve_dir;
 mod serve_file;
@@ -202,4 +203,77 @@ fn response_from_io_error(
         }
         _ => Err(err),
     }
+}
+
+struct LastModified(HttpDate);
+
+impl From<SystemTime> for LastModified {
+    fn from(time: SystemTime) -> Self {
+        LastModified(time.into())
+    }
+}
+
+struct IfUnmodifiedSince(HttpDate);
+struct IfModifiedSince(HttpDate);
+
+impl IfModifiedSince {
+    /// Check if the supplied time means the resource has been modified.
+    fn is_modified(&self, last_modified: &LastModified) -> bool {
+        self.0 < last_modified.0
+    }
+
+    /// convert a header value into a IfModifiedSince, invalid values are silentely ignored
+    fn from_header_value(value: &http::header::HeaderValue) -> Option<IfModifiedSince> {
+        String::from_utf8(value.as_bytes().to_vec())
+            .ok()
+            .map(|value| httpdate::parse_http_date(&value).ok())
+            .flatten()
+            .map(|time| IfModifiedSince(time.into()))
+    }
+}
+
+impl IfUnmodifiedSince {
+    /// Check if the supplied time passes the precondtion.
+    fn precondition_passes(&self, last_modified: &LastModified) -> bool {
+        self.0 >= last_modified.0
+    }
+
+    /// convert a header value into a IfModifiedSince, invalid values are silentely ignored
+    fn from_header_value(value: &http::header::HeaderValue) -> Option<IfUnmodifiedSince> {
+        String::from_utf8(value.as_bytes().to_vec())
+            .ok()
+            .map(|value| httpdate::parse_http_date(&value).ok())
+            .flatten()
+            .map(|time| IfUnmodifiedSince(time.into()))
+    }
+}
+
+fn check_modified_headers(
+    modified: &Option<LastModified>,
+    if_unmodified_since: Option<IfUnmodifiedSince>,
+    if_modified_since: Option<IfModifiedSince>,
+) -> Option<StatusCode> {
+    if let Some(since) = if_unmodified_since {
+        let precondition = modified
+            .as_ref()
+            .map(|time| since.precondition_passes(time))
+            .unwrap_or(false);
+
+        if !precondition {
+            return Some(StatusCode::PRECONDITION_FAILED);
+        }
+    }
+
+    if let Some(since) = if_modified_since {
+        let unmodified = modified
+            .as_ref()
+            .map(|time| !since.is_modified(&time))
+            // no last_modified means its always modified
+            .unwrap_or(false);
+        if unmodified {
+            return Some(StatusCode::NOT_MODIFIED);
+        }
+    }
+
+    None
 }
--- a/tower-http/src/services/fs/mod.rs
+++ b/tower-http/src/services/fs/mod.rs
@@ -242,8 +242,7 @@ impl IfUnmodifiedSince {
     fn from_header_value(value: &http::header::HeaderValue) -> Option<IfUnmodifiedSince> {
         String::from_utf8(value.as_bytes().to_vec())
             .ok()
-            .map(|value| httpdate::parse_http_date(&value).ok())
-            .flatten()
+            .and_then(|value| httpdate::parse_http_date(&value).ok())
             .map(|time| IfUnmodifiedSince(time.into()))
     }
 }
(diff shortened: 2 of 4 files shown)


davidpdrsn:
@zenria do you merge/rebase the latest master? Seems some things have changed making this hard to review.

zenria:
> @zenria do you merge/rebase the latest master? Seems some things have changed making this hard to review.

I've just rebased my branch on master :)

davidpdrsn:
Thank you! Given the size of this change I'm pushing it to 0.2 as I would like to get 0.1.2 our relatively soon.

tower-http/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tokio-util = { version = "0.6", optional = true, default_features = false, featu
 tower = { version = "0.4.1", optional = true }
 tracing = { version = "0.1", default_features = false, optional = true }
 percent-encoding = { version = "2.1.0", optional = true }
+headers = { version = "0.3", optional = true }
```
davidpdrsn, line 37: I'm not sure we should depend on `headers`. Using the stuff in `http::headers` is fine imo. Not worth adding another dependency for.

davidpdrsn (changes requested):

davidpdrsn:
We made some more changes to CI end of last week so you'll have to rebase `master` again to get CI working.

Just a heads up there are also some other PRs landing this week which touch these files so you'll probably get some merge conflicts soon.

tower-http/Cargo.toml
zenria: I'll have a look on how to remove this dependency

zenria:
I've rebased my branch, you were right, there was some merge conflicts ;)

I've got rid of `headers` crate, but I was forced to introduce a dependency on `httpdate` to parse and generate headers.

davidpdrsn:
Thanks! I'll take a look some time this week.

davidpdrsn:
Just a heads up. I'm waiting on https://github.com/tower-rs/tower-http/pull/173 being merged before diving into this.

davidpdrsn:
Removing this from the 0.2 milestone. It wont require breaking changes so we can always ship it in 0.2.1

shepmaster:
Thank you for your work on this; I'm excited to have this functionality in tower-http!

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -254,12 +268,22 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
                 full_path.set_extension(new_extension);
             }
 
+            let metadata = std::fs::metadata(&full_path)?;
```
davidpdrsn, line 271: This should use [`tokio::fs::metadata`](https://docs.rs/tokio/latest/tokio/fs/fn.metadata.html) to not block the runtime. We already do some things with the metadata on `master` so if you rebase it would be relatively easy to fix.

davidpdrsn (changes requested):
Looks good overall! Using `httpdate` for parsing is fine. Hyper is also using it.

I'm sorry but it looks like you'll have to rebase `master` again 😞 At least `ServeFile` has been simplified quite a bit which should make this simpler as well.

zenria:
I've rebased my branch :)

tower-http/src/services/fs/mod.rs
```diff
@@ -213,3 +214,76 @@ fn response_from_io_error(
...
+    /// Check if the supplied time means the resource has been modified.
+    fn is_modified(&self, last_modified: &LastModified) -> bool {
+        self.0 < last_modified.0
+    }
+
+    /// convert a header value into a IfModifiedSince, invalid values are silentely ignored
+    fn from_header_value(value: &http::header::HeaderValue) -> Option<IfModifiedSince> {
+        String::from_utf8(value.as_bytes().to_vec())
+            .ok()
+            .map(|value| httpdate::parse_http_date(&value).ok())
```
shepmaster, line 239: `map` followed by `flatten` should be `flat_map`

tower-http/src/services/fs/mod.rs
```diff
@@ -213,3 +214,76 @@ fn response_from_io_error(
...
+    fn precondition_passes(&self, last_modified: &LastModified) -> bool {
+        self.0 >= last_modified.0
+    }
+
+    /// convert a header value into a IfModifiedSince, invalid values are silentely ignored
+    fn from_header_value(value: &http::header::HeaderValue) -> Option<IfUnmodifiedSince> {
+        String::from_utf8(value.as_bytes().to_vec())
+            .ok()
+            .map(|value| httpdate::parse_http_date(&value).ok())
+            .flatten()
```
shepmaster, line 256: Ditto

tower-http/src/services/fs/mod.rs
```diff
@@ -213,3 +214,76 @@ fn response_from_io_error(
...
+            .map(|time| IfUnmodifiedSince(time.into()))
+    }
+}
+
+fn check_modified_headers(
+    modified: &Option<LastModified>,
```
shepmaster, line 262: Usually you don’t want references to options, but an optional reference. `Option::as_ref` Can be used

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -304,6 +307,17 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
...
+        let if_unmodified_since = req
+            .headers()
+            .get(header::IF_UNMODIFIED_SINCE)
+            .map(IfUnmodifiedSince::from_header_value)
+            .flatten();
+        let if_modified_since = req
+            .headers()
+            .get(header::IF_MODIFIED_SINCE)
+            .map(IfModifiedSince::from_header_value)
+            .flatten();
```
shepmaster, line 319: Ditto here x2

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -424,7 +459,7 @@ fn append_slash_on_path(uri: Uri) -> Uri {
 enum Output {
     File(FileRequest),
     Redirect(HeaderValue),
-    NotFound,
+    StatusCode(StatusCode),
```
shepmaster, line 462: Feels unfortunate that this went from one possibility to hundreds when it seemingly only needed two possibilities.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -749,7 +787,6 @@ mod tests {
             .unwrap();
         let res = svc.oneshot(request).await.unwrap();
 
-        assert_eq!(res.headers()["content-type"], "text/plain");
```
shepmaster, line 752: Why’s this get removed?

tower-http/src/services/fs/serve_file.rs
```diff
@@ -1,8 +1,7 @@
-//! Service that serves a file.
```
shepmaster, line 1: Why’s this get removed?

shepmaster (commented):
My comments don’t carry any weight in this repository, just actively reading through to see what changed.

tower-http/src/services/fs/serve_dir.rs
zenria: Sorry, this is a merge error!

tower-http/src/services/fs/serve_file.rs
zenria: this too :(

tower-http/src/services/fs/serve_dir.rs
zenria: Yes, you're right!

tower-http/src/services/fs/mod.rs
zenria: yes! you're right. (btw this is an option, as there is no `flat_map` on that type, I always forget I can use `and_then` to do that.)

tower-http/src/services/fs/mod.rs
zenria: in that case this does not matter, but I agree.

tower-http/src/services/fs/serve_dir.rs
zenria: the `check_modified_headers` can output 2 different status codes: `304 Not Modified` and `412 Precondition Failed` this is why I'm not using 2 more enum variants.

tower-http/src/services/fs/mod.rs
```diff
@@ -213,3 +214,74 @@ fn response_from_io_error(
...
+    fn precondition_passes(&self, last_modified: &LastModified) -> bool {
+        self.0 >= last_modified.0
+    }
+
+    /// convert a header value into a IfModifiedSince, invalid values are silentely ignored
+    fn from_header_value(value: &http::header::HeaderValue) -> Option<IfUnmodifiedSince> {
+        String::from_utf8(value.as_bytes().to_vec())
+            .ok()
+            .and_then(|value| httpdate::parse_http_date(&value).ok())
+            .map(|time| IfUnmodifiedSince(time.into()))
```
Nehliin, line 255:
```suggestion
        std::str::from_utf8(value.as_bytes()))
            .ok()
            .and_then(|value| httpdate::parse_http_date(value).ok())
            .map(|time| IfUnmodifiedSince(time.into()))
```

tower-http/src/services/fs/serve_file.rs
```diff
@@ -446,4 +446,58 @@ mod tests {
...
+        // -- If-Modified-Since
+
+        let svc = ServeFile::new("../README.md");
+        let req = Request::builder()
+            .header(header::IF_MODIFIED_SINCE, last_modified)
+            .body(Body::empty())
+            .unwrap();
+
+        let res = svc.oneshot(req).await.unwrap();
+        assert_eq!(res.status(), StatusCode::NOT_MODIFIED);
```
Nehliin, line 472: Could we please also check that the body is empty here like the spec states?

tower-http/src/services/fs/serve_file.rs
```diff
@@ -446,4 +446,58 @@ mod tests {
...
+
+        let svc = ServeFile::new("../README.md");
+        let req = Request::builder()
+            .header(header::IF_MODIFIED_SINCE, "Fri, 09 Aug 1996 14:21:40 GMT")
+            .body(Body::empty())
+            .unwrap();
+
+        let res = svc.oneshot(req).await.unwrap();
+        assert_eq!(res.status(), StatusCode::OK);
```
Nehliin, line 481: Let's check that the file is actually part of the body here as a sanity check.

tower-http/src/services/fs/serve_file.rs
```diff
@@ -446,4 +446,58 @@ mod tests {
...
+        // -- If-Unmodified-Since
+
+        let svc = ServeFile::new("../README.md");
+        let req = Request::builder()
+            .header(header::IF_UNMODIFIED_SINCE, last_modified)
+            .body(Body::empty())
+            .unwrap();
+
+        let res = svc.oneshot(req).await.unwrap();
+        assert_eq!(res.status(), StatusCode::OK);
```
Nehliin, line 492: Same here

tower-http/src/services/fs/serve_file.rs
```diff
@@ -446,4 +446,58 @@ mod tests {
...
+
+        let svc = ServeFile::new("../README.md");
+        let req = Request::builder()
+            .header(header::IF_UNMODIFIED_SINCE, "Fri, 09 Aug 1996 14:21:40 GMT")
+            .body(Body::empty())
+            .unwrap();
+
+        let res = svc.oneshot(req).await.unwrap();
+        assert_eq!(res.status(), StatusCode::PRECONDITION_FAILED);
```
Nehliin, line 501: And this shouldn't return the body.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -1124,4 +1160,63 @@ mod tests {
             &format!("bytes */{}", file_contents.len())
         )
     }
+    #[tokio::test]
+    async fn last_modified() {
```
Nehliin, line 1164: Same comments here like in the tests for serve_file

Nehliin (approved):
Nicely done! Thanks for the PR! I have some comments regarding the tests but those extra checks should still pass with this implementation. Also have a comment regarding some unnecessary heap allocations. 

I will approve this since I personally think my comments can be easily fixed in a follow up pr by either you or me and this has been in the works for some time now.  But it's up to @davidpdrsn if he wants to block this on my comments.

zenria:
Ok, thank you @Nehliin I'll update the PR with the CI & suggested fixes :)

tower-http/src/services/fs/mod.rs
```diff
@@ -213,3 +214,74 @@ fn response_from_io_error(
...
+    fn is_modified(&self, last_modified: &LastModified) -> bool {
+        self.0 < last_modified.0
+    }
+
+    /// convert a header value into a IfModifiedSince, invalid values are silentely ignored
+    fn from_header_value(value: &http::header::HeaderValue) -> Option<IfModifiedSince> {
+        String::from_utf8(value.as_bytes().to_vec())
+            .ok()
+            .and_then(|value| httpdate::parse_http_date(&value).ok())
+            .map(|time| IfModifiedSince(time.into()))
```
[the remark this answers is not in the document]
zenria: Ok, I was not aware of this function, thanks a lot!

davidpdrsn (approved):

photino:
I cannot find the `Last-Modified` header in the response for `tower-http` 0.4.4.

davidpdrsn:
Please open a new issue instead of resurrecting such and old PR.

PR #147, opened 2021-09-18
Update trait bounds of Set{Request,Response}HeaderLayer constructors

jplatte:
#143 plus breaking change. I'll rebase once that is merged (just so GitHub doesn't show #143's commits as being part of this one anymore).

PR #148, opened 2021-10-04
Remove unnecessary generic parameter from Set{Request,Response}HeaderLayer

### LINKED ISSUE #127: SetResponseHeaderLayer et al are hard to use with literal HeaderValues

jplatte:
## Motivation

Currently, `SetRequestHeaderLayer` and `SetResponseHeaderLayer` have to have a turbofish supplied when used with a plain `HeaderValue` (or `Option<HeaderValue>`) instead of a closure, which is annoying.

## Solution

Adjust the generics to make things work out :)

Unfortunately, this conflicts with #147 and as such error messages from closures with wrong parameter types will continue to be reported rather late somewhere in the usage of these layers rather than at the construction site. However, I consider this change to be more valuable than #147.

This is a breaking change because it changes the number of generic parameters and previous `::<_, Body>` turbofish usage will thus become invalid.

Properly fixes #127

### DIFF
--- a/examples/tonic-key-value-store/src/main.rs
+++ b/examples/tonic-key-value-store/src/main.rs
@@ -280,7 +280,7 @@ async fn make_client(
         // Decompress response bodies
         .layer(DecompressionLayer::new())
         // Set a `User-Agent` header
-        .layer(SetRequestHeaderLayer::<_, Request<BoxBody>>::overriding(
+        .layer(SetRequestHeaderLayer::overriding(
             header::USER_AGENT,
             HeaderValue::from_static("tonic-key-value-store"),
         ))
--- a/examples/warp-key-value-store/src/main.rs
+++ b/examples/warp-key-value-store/src/main.rs
@@ -2,7 +2,7 @@ use bytes::Bytes;
 use hyper::{
     body::HttpBody,
     header::{self, HeaderValue},
-    Body, Request, Response, Server, StatusCode,
+    Body, Response, Server, StatusCode,
 };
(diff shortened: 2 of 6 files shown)


davidpdrsn (approved):
I think this looks good and is worth doing. I've seen several people be confused about the turbofish. Wanna rebase/merge the lastest `master` and add a note to the changelog? Then it'll be gtm.

jplatte:
Done.

davidpdrsn:
Thanks!

Issue #135, opened 2021-09-01
Proposal for X-Request-Id


DoumanAsh:
## Feature Request

Add hanlding of `x-request-id` and/or automatically generate it when missing

### Motivation

Request-id is useful if we want to be able to efficiently correlate logs between client and server.
If client doesn't generate himself, we still can benefit by sending such id from server as we can ask client later to provide it if there is something wrong with response

### Proposal

Implementation example https://github.com/DoumanAsh/tower-http-req-id

For now there is only UUID generator, but I made code generic enough to allow crate's user to supply own generators.

Basically idea is that middleware checks whether header is present, if it is then it stores it in extensions.
If not it generates and adds it to extension.

Following that once request is resolved it adds the same header to Response

### Alternatives

I'll publish it as 3pp crate

DoumanAsh:
@davidbarsky ping, what do you think about adding such middleware to this crate?

davidpdrsn:
I think it makes sense! I've actually played around with a request-id layer at work.

The API I have in mind is

```rust
struct SetRequestIdLayer {
    // ...
}

impl SetRequestId {
    fn uuid() -> Self;

    fn custom_generator<T>(gen: T) -> Self
    where
        T: IdGenerator;

    // Defaults to X-Request-Id
    fn header_name(self, name: HeaderName) -> Self;
}

trait IdGenerator {
    fn gen(&mut self) -> HeaderValue;
}
```

What do you think?

DoumanAsh:
With API that you have in mind, how user access the id?
Should he always inspect header for when it is needed?
I don't really like extensions myself as they are quite ugly to use due to lack of compile-time awareness, but it makes your id typed at least, allowing a simple way to copy it when it is for example integer like uuid 😄

davidpdrsn:
Hm good question. Certainly a `RequestId(pub HeaderValue)` extension would be easiest but certainly has drawbacks as you mention. Do you have any other ideas?

DoumanAsh:
There is another way to make it less ugly for user:
1. Insert `X-Request-ID` if user client didn't set it.
2. Use extractor to get `id` from header
3. Profit

That should at least remove need for user to work directly with headers himself.

PR #150, opened 2021-10-06
Add `SetRequestId` and `PropagateRequestId` middleware

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/135

Example usage:

```rust
use http::{Request, Response, header::HeaderName};
use tower::{Service, ServiceExt, ServiceBuilder};
use tower_http::request_id::{
    SetRequestIdLayer, PropagateRequestIdLayer, MakeRequestId, RequestId,
};
use hyper::Body;
use std::sync::{Arc, atomic::{AtomicU64, Ordering}};

// A `MakeRequestId` that increments an atomic counter
#[derive(Clone, Default)]
struct MyMakeRequestId {
    counter: Arc<AtomicU64>,
}

impl MakeRequestId for MyMakeRequestId {
    fn make_request_id<B>(&mut self, request: &Request<B>) -> Option<RequestId> {
        let request_id = self.counter
            .fetch_add(1, Ordering::SeqCst)
            .to_string()
            .parse()
            .unwrap();

        Some(RequestId::new(request_id))
    }
}

let x_request_id = HeaderName::from_static("x-request-id");

ServiceBuilder::new()
    // set `x-request-id` header on all requests
    .layer(SetRequestIdLayer::new(
        x_request_id.clone(),
        MyMakeRequestId::default(),
    ))
    // propagate `x-request-id` headers from request to response
    .layer(PropagateRequestIdLayer::new(x_request_id))
    .service(handler);
```

Or slightly shorter using `ServiceBuilderExt`:

```rust
use tower_http::ServiceBuilderExt;

ServiceBuilder::new()
    .set_x_request_id(MyMakeRequestId::default())
    .propagate_x_request_id()
    .service(handler);
```

These are two different middleware since they need to be applied around any logging you might have. The docs have an example.


davidpdrsn:
@DoumanAsh what do you think about this?

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,575 @@
...
+    type Error = S::Error;
+    type Future = S::Future;
+
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
+        if let Some(request_id) = req.headers().get(&self.header_name).cloned() {
```
DoumanAsh, line 377: `clone` should be done inside branch when `req.extensions().get::<RequestId>().is_none()`

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,575 @@
...
+            make_request_id,
+        }
+    }
+
+    /// Create a new `SetRequestIdLayer` that uses `x-request-id` as the header name.
+    pub fn x_request_id(make_request_id: M) -> Self
+    where
+        M: MakeRequestId,
+    {
+        SetRequestIdLayer {
```
DoumanAsh, line 288: Self::new(HeaderName::from_static(X_REQUEST_ID), make_request_id)`

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,575 @@
...
+use std::{future::Future, pin::Pin};
+use tower_layer::Layer;
+use tower_service::Service;
+
+pub(crate) const X_REQUEST_ID: &str = "x-request-id";
+
+/// Trait for producing [`RequestId`]s.
+///
+/// Used by [`SetRequestId`].
+pub trait MakeRequestId {
```
DoumanAsh, line 228: Might as well `MakeRequestId: Clone` since it is anyway required for it to be useful

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,575 @@
...
+    type Output = Result<Response<B>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let mut response = futures_core::ready!(this.inner.poll(cx))?;
+
+        if let Some(current_id) = response.headers().get(&*this.header_name) {
+            if response.extensions().get::<RequestId>().is_none() {
+                let current_id = current_id.clone();
+                response.extensions_mut().insert(current_id);
```
DoumanAsh, line 502: I don't think putting it inside `Response`'s extensions is useful?
Generally you just want to return back header to user for request identification
If you're concerned about using logging middleware, then it is not necessary too.
In my case I just extract request id from request and pass it to future alongside future that resolves response

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,575 @@
...
+            }
+        } else if let Some(request_id) = this.request_id.take() {
```
DoumanAsh, line 504: it shouldn't be `else if`
It should `if let` as you definitely want to propagate request id if it is set

DoumanAsh (approved):
That looks pretty good and generic too for most use cases
I left some comments based on my own middleware

tower-http/src/request_id.rs
davidpdrsn: Its an `else if` because If the response already has a request id for some reason, perhaps another middleware that conditionally applies it, then I don't think we should override it.

tower-http/src/request_id.rs
davidpdrsn: I think its useful to also have available in response extensions as it makes it easier to access the request id if you don't know which header value is being used.

tower-http/src/request_id.rs
davidpdrsn: It's only required for the layer and not the service. So not strictly required.

tower-http/src/request_id.rs
DoumanAsh: that sounds like a programming error to me
Well, it is up to you, but I think user shouldn't have multiple layers that set request id

tower-http/src/request_id.rs
DoumanAsh: Maybe, but I just do not see case where you can have access to response but not request itself?

davidpdrsn: In a `.map_response` combinator.

tower-http/src/request_id.rs
davidpdrsn: I don't think it hurts to support.

tower-http/src/request_id.rs
DoumanAsh: Ah I see if there is use case then it makes sense

tower-http/src/request_id.rs
davidpdrsn: I mean I'm just speculating so maybe not 🤷‍♂️ Would like to hear what others think.

tower-http/src/builder.rs
```diff
@@ -302,6 +302,58 @@ pub trait ServiceBuilderExt<L>: crate::sealed::Sealed<L> {
...
+    ///
+    /// [`tower_http::request_id`]: crate::request_id
+    #[cfg(feature = "request-id")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "request-id")))]
+    fn propagate_request_id(
+        self,
+        header_name: HeaderName,
+    ) -> ServiceBuilder<Stack<crate::request_id::PropagateRequestIdLayer, L>>;
+
+    /// TODO
```
Nehliin, line 349: Forgotten to add this?

tower-http/src/builder.rs
```diff
@@ -302,6 +302,58 @@ pub trait ServiceBuilderExt<L>: crate::sealed::Sealed<L> {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "request-id")))]
+    fn set_request_id<M>(
+        self,
+        header_name: HeaderName,
+        make_request_id: M,
+    ) -> ServiceBuilder<Stack<crate::request_id::SetRequestIdLayer<M>, L>>
+    where
+        M: crate::request_id::MakeRequestId;
+
+    /// TODO
```
Nehliin, line 321: Same here as below

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,573 @@
...
+        )
+    }
+}
+
+/// Set request id headers and extensions on requests.
+///
+/// See the [module docs](self) for an example.
+///
+/// If [`MakeRequestId::make_request_id`] returns `Some(_)` and the request doesn't already have a
+/// header with the same name, then the header will be inserted.
```
Nehliin, line 312: Maybe worth adding the part about not overwriting the header to the module docs as well?

tower-http/src/request_id.rs
Nehliin: I agree that I don't think it hurts to support.

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,573 @@
...
+            Arc,
+        },
+    };
+    use tower::{ServiceBuilder, ServiceExt};
+
+    #[allow(unused_imports)]
+    use super::*;
+
+    #[tokio::test]
+    async fn test_something() {
```
Nehliin, line 530: Maybe worth adding a test for the cutsom header as well?

Nehliin (commented):
Really like this feature! Will be super nice when its merged, I only had a couple of small comments

tower-http/src/request_id.rs
```diff
@@ -0,0 +1,624 @@
...
+            .map_request(|request: Request<_>| {
+                // `set_x_request_id` should set the extension if its missing
+                assert_eq!(request.extensions().get::<RequestId>().unwrap().0, "foo");
+                request
+            })
+            .propagate_x_request_id()
+            .service_fn(handler);
+
+        let req = Request::builder()
+            .header("x-request-id", "foo")
```
Nehliin, line 582: Just to clarify, its expected to produce the same result even if this would have `bar` as header value here right?

Nehliin (approved):

tower-http/src/request_id.rs
davidpdrsn: Yep. I've changed the test in 69692cec3744f to make that more clear.

Nehliin (approved):
🎉

MarcusGrass (approved):
LGTM, convenient implementation

Nehliin (approved):

PR #151, opened 2021-10-15
Switch from pin-project to pin-project-lite

### LINKED ISSUE #115: Migrate to pin-project-lite
Tower recently migrated to pin-project-lite (https://github.com/tower-rs/tower/pull/595) so it makes sense to do the same for tower-http. It should be a fairly mechanical operation. Off the top of my head tower-http isn't using any of pin-projects more advanced features.

jplatte:
Fixes #115.

Contains a decent amount of duplicated code, most of which can be removed once https://github.com/rust-lang/rust/issues/51085 is stable. It might be possible to reduce this duplication using a macro (but I'm not sure whether macros are allowed to take the place of entire match arms, rather than just patterns / expressions), if you want I'll look into that.

### DIFF
--- a/tower-http/src/auth/require_authorization.rs
+++ b/tower-http/src/auth/require_authorization.rs
@@ -116,7 +116,7 @@ use http::{
     Request, Response, StatusCode,
 };
 use http_body::Body;
-use pin_project::pin_project;
+use pin_project_lite::pin_project;
 use std::{
     fmt,
     future::Future,
@@ -279,31 +279,41 @@ where
     }
 }
 
-/// Response future for [`RequireAuthorization`].
-#[pin_project]
-pub struct ResponseFuture<F, B> {
-    #[pin]
-    kind: Kind<F, B>,
+pin_project! {
+    /// Response future for [`RequireAuthorization`].
+    pub struct ResponseFuture<F, B> {
+        #[pin]
+        kind: Kind<F, B>,
+    }
 }
 
 impl<F, B> ResponseFuture<F, B> {
     fn future(future: F) -> Self {
         Self {
-            kind: Kind::Future(future),
+            kind: Kind::Future { future },
         }
     }
 
     fn invalid_auth(res: Response<B>) -> Self {
         Self {
-            kind: Kind::Error(Some(res)),
+            kind: Kind::Error {
+                response: Some(res),
+            },
         }
     }
 }
 
-#[pin_project(project = KindProj)]
-enum Kind<F, B> {
-    Future(#[pin] F),
-    Error(Option<Response<B>>),
+pin_project! {
+    #[project = KindProj]
+    enum Kind<F, B> {
+        Future {
+            #[pin]
+            future: F,
+        },
+        Error {
+            response: Option<Response<B>>,
+        },
+    }
 }
 
 impl<F, B, E> Future for ResponseFuture<F, B>
@@ -314,8 +324,8 @@ where
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.project().kind.project() {
-            KindProj::Future(future) => future.poll(cx),
-            KindProj::Error(response) => {
+            KindProj::Future { future } => future.poll(cx),
+            KindProj::Error { response } => {
                 let response = response.take().unwrap();
                 Poll::Ready(Ok(response))
             }
--- a/tower-http/src/compression/body.rs
+++ b/tower-http/src/compression/body.rs
@@ -14,57 +14,82 @@ use bytes::{Buf, Bytes};
 use futures_util::ready;
 use http::HeaderMap;
 use http_body::Body;
-use pin_project::pin_project;
+use pin_project_lite::pin_project;
 use std::{
     io,
+    marker::PhantomData,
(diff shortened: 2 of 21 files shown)


tower-http/src/compression/body.rs
```diff
@@ -14,124 +14,216 @@ use bytes::{Buf, Bytes};
...
+
+#[cfg(feature = "compression-deflate")]
+type DeflateBody<B> = WrapBody<ZlibEncoder<B>>;
+#[cfg(not(feature = "compression-deflate"))]
+type DeflateBody<B> = (Never, PhantomData<B>);
+
+#[cfg(feature = "compression-br")]
+type BrotliBody<B> = WrapBody<BrotliEncoder<B>>;
+#[cfg(not(feature = "compression-br"))]
+type BrotliBody<B> = (Never, PhantomData<B>);
```
davidpdrsn, line 181: Just checking but are these things necessary because you cannot put `cfg`s inside `pin_project_lite::pin_project!`?

davidpdrsn (approved):
Thanks @jplatte. It looks good to me 😊

A merge/rebase of `master` should fix CI.

tower-http/src/compression/body.rs
jplatte: Yeah, exactly.

davidpdrsn:
Just a small heads up I think I'll merge https://github.com/tower-rs/tower-http/pull/156 before this, which might lead to some merge conflicts here.

jplatte:
Rebased. Also had to switch the new `cors` module to pin-project-lite, but that was straight-forward.

tower-http/src/cors.rs
```diff
@@ -4,7 +4,7 @@
 //!
 //! ```
 //! use http::{Request, Response, Method, header};
-//! use hyper::Body;
+//! use hypefut: futureBody;
```
davidpdrsn, line 7: I guess this needs to be

```suggestion
//! use hyper::Body;
```

davidpdrsn (approved):
Yep thats cool. Thanks!

tower-http/src/cors.rs
jplatte: Looks like rust-analyzer used some outdated spans when doing updating uses of a field I renamed 😅

hawkw: huh, that seems like maybe something that deserves an upstream bug report?

tower-http/src/auth/require_authorization.rs
```diff
@@ -279,31 +279,41 @@ where
...
+    #[project = KindProj]
+    enum Kind<F, B> {
+        Future {
+            #[pin]
+            future: F,
+        },
+        Error {
+            response: Option<Response<B>>,
+        },
+    }
```
hawkw, line 316: not a blocker, but is there a reason for changing this from a tuple-like enum to a struct-like enum? AFAICT this just adds extra changed lines to the diff without an obvious reason, but I might be missing something...

tower-http/src/compression/body.rs
```diff
@@ -14,124 +14,216 @@ use bytes::{Buf, Bytes};
...
+            inner: DeflateBody<B>,
+        },
+        Brotli {
+            #[pin]
+            inner: BrotliBody<B>,
+        },
+        Identity {
+            #[pin]
+            inner: B,
+        },
```
hawkw, line 204: similarly, i'm not really convinced these needed to be changed from tuple-like to struct-like just so we could name the fields `inner`; especially since this type is mostly destructured in matches where we already bound the inner field to the name `inner`.

tower-http/src/compression/body.rs
```diff
@@ -14,124 +14,216 @@ use bytes::{Buf, Bytes};
...
-            BodyInner::Identity(inner) => inner,
+            BodyInner::Brotli { inner } => inner.read.get_ref().get_ref().get_ref().get_ref(),
+            BodyInner::Identity { inner } => inner,
+
+            #[cfg(not(feature = "compression-gzip"))]
+            BodyInner::Gzip { inner } => match inner.0 {},
+            #[cfg(not(feature = "compression-deflate"))]
+            BodyInner::Deflate { inner } => match inner.0 {},
+            #[cfg(not(feature = "compression-br"))]
+            BodyInner::Brotli { inner } => match inner.0 {},
```
hawkw, line 63: this certainly _feels_ weird to a reader who doesn't know what's going on here...it might be worth adding a comment explaining why this code is necessary, with a reference to https://github.com/rust-lang/rust/issues/51085

but, not a blocker.

hawkw (approved):
this looks good to me; i commented on a couple of minor notes that aren't really blocking.

tower-http/src/auth/require_authorization.rs
davidpdrsn: pin-project-lite doesn't support tuple enums afaik.

davidpdrsn:
@jplatte wanna fix the last round of conflicts? Then I'll merge 🚀

jplatte:
Done. Also added a reference to the exhaustive match feature tracking issue to the first `match never_enum {}` occurrence in each `body.rs`.

davidpdrsn:
Thanks a lot for working on this!

Issue #146, opened 2021-09-17
CompressionLayer should perhaps check response's MIME type


shamblesides:
### Version

0.1.1

### Platform

Linux 5.4.0-84-generic #94-Ubuntu SMP Thu Aug 26 20:27:37 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Description

It seems like the CompressionLayer compresses all responses, including images, which do not need to be compressed. When I put a CompressionLayer in front of ServeDir for example, I would expect that only text files should be compressed.

davidpdrsn:
Makes sense!

Being able to add a predicate that takes the `http::response::Parts` from the inner service's response and returns whether or not it should be compressed would probably work fine.

Issue #154, opened 2021-11-09
Use `http::request::Parts` instead of `Request<B>` in methods that produce things from requests


davidpdrsn:
We have a couple of methods that look something like this

```rust
fn make_thing<B>(&mut self, request: &Request<B>) -> Thing
```

Such as [`MakeClassifier`](https://docs.rs/tower-http/0.1.1/tower_http/classify/trait.MakeClassifier.html) and [`MakeSpan`](https://docs.rs/tower-http/0.1.1/tower_http/trace/trait.MakeSpan.html).

I've been thinking if it would make sense to change those methods to

```rust
fn make_thing(&mut self, parts: &http::request::Parts) -> Thing
```

This should in theory make rust's type inference better since there is one less generic param. It would also just be simpler overall to understand I think. Users cannot do anything with the request body anyway since there no trait bounds on it, they only have a shared reference to the request, and the functions aren't async.

Thoughts @hawkw @LucioFranco @olix0r?

olix0r:
@davidpdrsn I agree with the intent, but I think this is going to be a bit cumbersome given the current `http::Request` API. Specifically, we only have [`Request::into_parts`](https://docs.rs/http/0.2.5/http/request/struct.Request.html#method.into_parts) which requires fully decomposing the request. It would be a lot more natural if we had a `Request::parts() -> &Parts`...

davidpdrsn:
Yep would require using `into_parts `and `from_parts` inside the middleware. Users shouldn't have to worry about that though I think.

You're right that being able to borrow the parts would make it more convenient.

olix0r:
One thing to consider (from hyperium/http#511) is that switching to `&Parts` requires that the caller has ownership of the request. If you only have a `&http::Request<B>` reference, you can't produce a `&Parts`. So, `&Parts` may be suitable in very specific cases but is probably not the best general pattern to pursue, as it has some unexpected side-effects on composability.

davidpdrsn:
Hm yeah thats a valid point 😕

I think given the concerns I'll be conservative and close this for now. Lets just keep things as they are.

PR #155, opened 2021-11-09
Determine whether to compress by a predicate

MarcusGrass:
Fixes https://github.com/tower-rs/tower-http/issues/146 by implementing a predicate which takes http::response::Parts by ref and returns a boolean indicating whether the response body should be compressed.

This implementation demands that the filter is Copy, which may be reasonable.

Putting this up as a draft because if you like it we could do something similar for Decompression, since that uses the same default predicate to determine compressibility.


tower-http/src/compression/compression_filter.rs
```diff
@@ -0,0 +1,26 @@
+use crate::compression_utils::default_compression_filter_predicate;
+
+/// A filter which any response Parts needs to pass to be compressed
+pub trait CompressionFilter: Copy {
```
davidpdrsn, line 4: I think requiring `Copy` might be too limiting in some cases. For examples users might want to implement `CompressionFilter` for something like

```rust
struct MyFilter
    compression_enabled: Arc<AtomicBool>,
}
```

This isn't possible since `Arc` isn't `Copy` but only `Clone`.

So I think we should require `CompressionFilter: Clone` instead.

What do you think?

tower-http/src/compression/compression_filter.rs
```diff
@@ -0,0 +1,26 @@
...
+    /// Predicate which takes response parts and returns true if the response should be compressed
+    fn should_compress(&self, parts: &http::response::Parts) -> bool;
```
davidpdrsn, line 6: I know it was me who suggested using `request::Parts` in the first place but we talked about it in https://github.com/tower-rs/tower-http/issues/154 and decided to stick with `Request`. So I think we should do that here as well to remain consistent.

```suggestion
    fn should_compress<B>(&self, request: &Request<B>) -> bool;
```

tower-http/src/compression/compression_filter.rs
```diff
@@ -0,0 +1,26 @@
...
+        (self)(parts)
+    }
+}
+
+/// Default compression filter that proxies `default_compression_filter_predicate` which looks at
+/// headers to determine whether compression is suitable
+#[derive(Default, Copy, Clone, Debug)]
+pub struct DefaultCompressionFilter {
+
+}
```
davidpdrsn, line 20: This is equivalent to

```suggestion
#[derive(Default, Copy, Clone, Debug)]
#[non_exhaustive]
pub struct DefaultCompressionFilter;
```

tower-http/src/compression/compression_filter.rs
```diff
@@ -0,0 +1,26 @@
...
+
+impl CompressionFilter for DefaultCompressionFilter {
+    fn should_compress(&self, parts: &http::response::Parts) -> bool {
+        default_compression_filter_predicate(&parts.headers)
```
davidpdrsn, line 24: What do you think about inlining `default_compression_filter_predicate` here? Might be slightly simpler since we don't have who places to look.

tower-http/src/compression/service.rs
```diff
@@ -12,19 +13,24 @@ use tower_service::Service;
...
+    pub fn new(service: S) -> Compression<S> {
         Self {
             inner: service,
             accept: AcceptEncoding::default(),
+            compression_filter: DefaultCompressionFilter{}
         }
     }
+}
+
+impl<S, P: CompressionFilter> Compression<S, P> {
```
davidpdrsn, line 33: Unless the trait bound is specifically necessary for lots of methods in the impl block then you commonly leave it. I would do

```suggestion
impl<S, P> Compression<S, P> {
```

tower-http/src/compression/service.rs
```diff
@@ -82,16 +88,26 @@ impl<S> Compression<S> {
         self.accept.set_br(false);
         self
     }
+
+    /// Replaces the current compression filter, `DefaultCompressionFilter` being the default
+    pub fn with_compression_filter<CF: CompressionFilter>(self, compression_filter: CF) -> Compression<S, CF> {
```
davidpdrsn, line 93:
```suggestion
    /// Replace the current compression filter.
    ///
    /// The default predicate is [`DefaultCompressionFilter`] which disables compression of gRPC
    /// (gRPC has its own protocol specific compression system) and responses who's
    /// mime type starts with `image/`.
    ///
    /// # Example
    ///
    /// ```
    /// todo!();
    /// ```
    pub fn compression_filter<CF>(self, compression_filter: CF) -> Compression<S, CF> 
    where
        CF: CompressionFilter
    {
```

- Expands the docs a bit to explain what the default compression does. Do we always apply the standard filtering regardless if this has been customized by the user? For example I don't think we should ever compress gRPC responses.
- I don't believe we normally use `with_` prefix for builder methods like this. For example `Trace` has `make_span_with`, `on_request`, etc.
- I personally prefer `where` clauses to `Foo: SomeTrait` 🤷 
- Makes `DefaultCompressionFilter` a link.
- We should probably also have an example.

tower-http/src/compression/compression_filter.rs
davidpdrsn: Also wondering if `CompressionPredicate` would be a better name? Tower already has [`Filter`](https://docs.rs/tower/0.4.10/tower/filter/struct.Filter.html) which is something very different 🤔

davidpdrsn (commented):
I think this is looking good! Mostly minor comments about the API.

davidpdrsn:
We've also made some changes to CI. You'll have to rebase/merge `master` to get CI working I think.

tower-http/src/compression/compression_filter.rs
MarcusGrass: It's a bit problematic since compression looks at the response while decompression looks at the request, since we're using `http::Response<B>` instead of headers they can't be easily merged

davidpdrsn: Wait a minute. Why does decompression need to determined by a predicate? Shouldn't we always decompress things if they're compressed?

davidpdrsn: The client advertising that it supports compression should always be fine since its up to the server if it actually wants to compress a resource. I think those checks are basically just for additional safety but could perhaps be removed. I looked at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) and [the spec](https://httpwg.org/specs/rfc7233.html#header.range) and they don't say `accept-encoding` cannot be combined with `range`.

davidpdrsn: I think we should wait with touching decompression in this PR at least. To keep the scope smaller.

MarcusGrass:
Moved to https://github.com/tower-rs/tower-http/pull/171

Issue #120, opened 2021-08-18
Have `ServeDir` and `ServeFile` serve precompressed files


paolobarbolini:
### Motivation

Currently tower-http has `tower_http::services::{ServeDir, ServeFile}` for serving files from the filesystem and `tower_http::compression::Compression` for compressing responses on the fly, but no way of serving precompressed files.

Since our service is built into a docker container we could precompress static assets at build time and have the server use the precompressed versions.

### Proposal

Add an option to both `ServeDir` and `ServeFile` to enable serving precompressed files based on the client's `Accept-Encoding` header.
This would allow us to gzip and brotli compress at the maximum compression level without incurring in the runtime cost of recompressing a file multiple times for every new request that comes in.

### Alternatives

Have static files be served by nginx, which has the [ngx_http_gzip_static_module module](http://nginx.org/en/docs/http/ngx_http_gzip_static_module.html) for serving precompressed files.

davidpdrsn:
What should happen if the file is requested by a client that doesn't support the compression that was used?

paolobarbolini:
> What should happen if the file is requested by a client that doesn't support the compression that was used?

The uncompressed version of the file would still need to be present, so that's the version that would be served.

The way the ngx_http_gzip_static_module module works is that you have both the uncompressed and the compressed versions of the files and the module chooses the best option for each client.

davidpdrsn:
It would be neat if this was something `Compression` could handle in a more general way. Not sure how that would work though.

I think this is a good thing to support regardless. Pretty much the same thing was requested for axum here https://github.com/tokio-rs/axum/issues/66

paolobarbolini:
> It would be neat if this was something `Compression` could handle in a more general way. Not sure how that would work though.

Thinking a bit about this idea I wouldn't want to have to compile the compression crates just to support serving precompressed files, so it would have to live under a separate crate feature.

PR #156, opened 2021-11-11
Precompressed files

Nehliin:
## Motivation
Suggestion for an implementation that should fix https://github.com/tower-rs/tower-http/issues/120. 

## Solution
The user specifies which types of precompression is available in the same directory as the original file by specifying `precompressed_[gzip|br|deflate]` to the `ServeFile` service. It is not required to have the compression feature enabled since no compression/decompression actually happens on the server. The `ServeFile` service looks for copies of the original file with the extension `.gz`, `.br`or `.zz` depending on the compression used/specified. If a client requests the file and doesn't support any of the precompressed variants the "normal" file  will be served instead. This means it's expected that the uncompressed version of the file also exists if all clients should be able to request the file. Otherwise a 404 will be returned if the client doesn't support the compression algorithm.


Nehliin:
Cleaned up a bit and added more tests and support for `ServeDir` so I think its ready for review now

tower-http/src/services/fs/serve_file.rs
```diff
@@ -74,7 +117,7 @@ impl ServeFile {
     }
 }
 
-impl<R> Service<R> for ServeFile {
+impl<ReqBody> Service<Request<ReqBody>> for ServeFile {
```
davidpdrsn, line 120: Its a shame that this required a breaking change. I think for 0.2 we should audit all services and make sure they implement `Service<Request<B>>` so we can make changes like this in the future.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -124,8 +173,27 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
                     HeaderValue::from_str(mime::APPLICATION_OCTET_STREAM.as_ref()).unwrap()
                 });
 
+            if let Some(file_extension) =
+                negotiated_encoding.and_then(|encoding| encoding.to_file_extention())
```
davidpdrsn, line 177: Just checking, but `to_file_extension` only returns `None` for "identity" encoding meaning technically `.filter(|encoding| *encoding != Encoding::Identity);` also handles the filtering. I don't necessarily think we have to change it, just making sure I understand why we're doing this check.

davidpdrsn (approved):
Great work! I think it looks good as it.

I'll put this on 0.2 milestone.

tower-http/src/services/fs/serve_dir.rs
Nehliin: I actually didn't have the filter before but the problem was the negotiated_encoding will then be passed as `Some(Encoding::Identity)` to the ResponseFuture which means the content encoding header would be set to `identity`. That should be fine but thought it would be nice to not change the behavior when there is no compression.

Nehliin:
Thanks! I'll fix the CI as well either today or tomorrow

davidpdrsn:
There were some general CI issues which I'm fixing in https://github.com/tower-rs/tower-http/pull/158

tower-http/src/services/fs/serve_dir.rs
davidpdrsn: Ah yes that makes sense!

Nehliin:
Ah great, yeah the the `cargo hack` issue should have been there before my pr as well unless I'm missing something :thinking:

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -297,6 +377,114 @@ mod tests {
...
+        let body = res.into_body().data().await.unwrap().unwrap();
+        let body = String::from_utf8(body.to_vec()).unwrap();
+        assert!(body.starts_with("\"This is a test file!\""));
+    }
+
+    #[tokio::test]
+    async fn only_precompressed_variant_existing() {
+        let svc = ServeDir::new("../test-files").precompressed_gzip();
+
+        let request = Request::builder().body(Body::empty()).unwrap();
```
shepmaster, line 465: Should this specify the `.uri` to match?

tower-http/src/content_encoding.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn to_str(self) -> &'static str {
+        match self {
+            Encoding::Gzip => "gzip",
+            Encoding::Deflate => "deflate",
+            Encoding::Brotli => "br",
+            Encoding::Identity => "identity",
+        }
+    }
+
+    pub(crate) fn to_file_extention(self) -> Option<&'static OsStr> {
```
shepmaster, line 29: typo: "extension"

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -62,6 +64,48 @@ impl ServeDir {
...
         self
     }
+
+    /// Informs the service that it should look for a precompressed (by the gzip algorithm)
+    /// version of _any_ file in the directory with the correct file extension  that can be
+    /// served if the client have the correct corresponding `Accept-Encoding`.
+    /// (i.e `dir/foo.txt.gz` when the served directory is `dir` and the requested file is `foo.txt`)
+    /// If not the uncompressed version will be served instead. Note that this means
+    /// that both the precompressed version(s) and the uncompressed file is expected
+    /// to be present in the same directory.
```
shepmaster, line 74:
```suggestion
    /// Informs the service that it should look for a precompressed gzip
    /// version of _any_ file in the directory.
    ///
    /// If the client has an `Accept-Encoding` header that allows the gzip encoding, 
    /// the file `dir/foo.txt.gz` will be served instead of `dir/foo.txt`.
    /// If the precompressed file is not available, or the client doesn't support it,
    /// the uncompressed version will be served instead. 
    /// Both the precompressed version and the uncompressed version are expected
    /// to be present in the directory.
```

(ditto similar suggestions to the sibling methods)

I'm not sure where it should be mentioned about the `q` settings for browser preference.

davidpdrsn (approved):

tower-http/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- `ServeDir` and `ServeFile`: Ability to serve precompressed files ([#156])
```
davidpdrsn, line 10: I think its missing a link to `[#156]` 🤔

Nehliin: Huh? Isn't that the last part `([#156])` or do I need to do something else as well?

davidpdrsn: You need to add link for it like its done for the other changelog entries. GH doesn't pick it up automatically unfortunately.

It'll look like this otherwise:

image

Nehliin: ah right completely missed that, fixed now!

c5soft:
Hi there, I'm wondering if it is a design problem that ServeDir/File will fail when precompressed_gzip enabled and path/file.ext.gz not exists but path/file.ext exists. this should works without content-encoding. 
In my opinion, negotiated_encoding should be determined by: 1. precompressed_variants 2.client Accept-Encoding, 3.precompressed file exists.  Any one of three conditions is not statisfied will cause uncompressed file used. Multiple prcompressed type co-statisfied, we my consider br first, then gzip, the third deflate.

Nehliin:
Hi @c5soft if the developer has explicitly indicated that there should exist an precompressed variant I think its fair to return 404 to clients that supports the encoding (all browsers) since it would be an error on the side of the developer and it makes it easy to notice after a basic check. Otherwise the developer might think it serves precompressed files without them actually existing.

Another option would be to remove the `precompressed_[gzip|br|deflate|` and automatically always look for precompressed variants and fallback if they don't exist. However that wouldn't be zero cost for developers who know they don't have precompressed files. We would have to do additional system calls and checks to look for the precompressed variants in almost all cases since again, all large browsers supports the content encodings.

davidpdrsn:
> if the developer has explicitly indicated that there should exist an precompressed variant I think its fair to return 404

I didn't catch that about this implementation. Not sure thats ideal since `ServeDir` serves any file in the directory and all subdirectories. So you'd have to precompress all files even if you're only really interested in precompressing a single file. You might also be serving images which don't benefit much from compression afaik.

Nehliin:
@davidpdrsn Yeah that's fair, I didn't think about the `ServeDir` case becoming quite inconvenient to work with if you have images etc. I can create a PR to fallback if it doesn't exist. I presume we want to change the `ServeFile` service then as well to keep it consistent?

davidpdrsn:
That would be great 😊 and yeah adding it to `ServeFile` as well would be nice.

Issue #153, opened 2021-10-21
License text not included in published crate


Jake-Shadle:
The `MIT` license text in the root `LICENSE` is not part of the package crate, which is a requirement of the `MIT` license.

- https://tldrlegal.com/license/mit-license

PR #157, opened 2021-11-12
Move license into crate

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/153


seanmonstar:
Usually you want a license in the root (too), implying it applies to all in the repo (examples, docs, etc).

davidpdrsn:
@seanmonstar how do you tell cargo where to find the license then?

seanmonstar:
This is the options I've seen:

- Just have 1 `Cargo.toml`, and a `src` directory at the root. Having it inside an extra directory I usually only see when there's multiple crates in a repository.
- If a repo has multiple crates, then the license file is sometimes symlinked into the other directories.
- Or, looks like the Tokio repo just puts a full copy in each place.

davidpdrsn:
Yep I looked at tokio when I fixed it in axum as well. Kinda think I just wanna do that as its easy.

I think when setting up the repo I didn't realize you could have a workspace with one of the crates being in the root of the repo. It could be converted to that. I don't plan on having multiple published crates in this repo.

LucioFranco (approved):

PR #158, opened 2021-11-12
Fix CI

davidpdrsn:
Copies parts of tower's and axum's CI workflows.


tower-http/src/lib.rs
```diff
@@ -210,7 +210,7 @@
     nonstandard_style,
     missing_docs
 )]
-#![deny(unreachable_pub, rustdoc::broken_intra_doc_links, private_in_public)]
```
davidpdrsn, line 213: This lint is check on CI.

.github/workflows/CI.yml
```diff
@@ -43,30 +49,73 @@ jobs:
...
-      run: cargo hack check --each-feature --no-dev-deps --all
+      run: cargo hack check --each-feature --no-dev-deps --workspace
 
   test-versions:
     # Test against the stable, beta, and nightly Rust toolchains on ubuntu-latest.
     needs: check
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: [stable, beta, nightly, 1.40.0]
```
davidpdrsn, line 54: tower-http doesn't declare an MSRV and it never compiled on 1.40 in the first place. Going to 1.51 for const generics.

davidpdrsn: But we still don't specify MSRV anywhere. Think I wanna keep it like that for now.

LucioFranco (approved):

PR #163, opened 2021-11-15
`ServeDir`: Add cache control feature: max-age default to 300 secs 

c5soft:
`ServeDir`: Add cache control feature: max-age . Based on 0.1.2, no breaking changes, used duration.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- `ServeDir`: Add cache control feature: max-age default to 300 secs  
+- PR: `ServeDir`: Add cache control feature: max-age default to 300 secs, from [c5soft](https://github.com/c5soft/tower-http)
 
 # 0.1.2 (November 13, 2021)
 
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- `ServeDir`: Add cache control feature: max-age default to 300 secs  
 
 # 0.1.2 (November 13, 2021)
 
--- a/tower-http/src/services/fs/serve_dir.rs
+++ b/tower-http/src/services/fs/serve_dir.rs
@@ -84,48 +84,13 @@ use tower_service::Service;
 /// #     .expect("server error");
 /// # };
 /// ```
-/// # Axum Example
-/// ```
-/// use axum::{
-///     error_handling::HandleErrorExt,
-///     http::StatusCode,
-///     routing::{get, service_method_routing},
-///     Router,
-/// };
-/// use tower_http::services::ServeDir;
-/// 
-/// async fn hello() -> &'static str {
-///     "Hello world"
-/// }
-/// 
-/// #[tokio::main]
-/// async fn main() {
-///     // build our application with a single route
-///     let app = Router::new().route("/hello", get(hello)).fallback(
-///         service_method_routing::get(ServeDir::new("public")).handle_error(
-///             |error: std::io::Error| {
-///                 Ok::<_, std::convert::Infallible>((
-///                     StatusCode::INTERNAL_SERVER_ERROR,
-///                     format!("Unhandled internal error: {}", error),
-///                 ))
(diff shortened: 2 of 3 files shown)


tower-http/src/services/fs/serve_dir.rs
```diff
@@ -83,11 +84,48 @@ use tower_service::Service;
...
+///     );
+/// 
+///     // run it with hyper on localhost:3000
+///     axum::Server::bind(&"0.0.0.0:3000".parse().unwrap())
+///         .serve(app.into_make_service())
+///         .await
+///         .unwrap();
+/// }
+/// ```
+
```
davidpdrsn, line 122: I don't think tower-http's examples should include axum. I think its more appropriate for axum's examples to include tower-http.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -120,6 +159,16 @@ impl ServeDir {
         self.buf_chunk_size = chunk_size;
         self
     }
+
+    /// Cache-Control:max-age=300.
+    ///
+    /// This speeds up user's feeling for static sites, especially large files.
+    ///
+    /// Defaults to 300 seconds.
```
davidpdrsn, line 167:
```suggestion
    /// Set the [`Cache-Control: max-age`][mdn] response directive
    ///
    /// This indicates that the response can be cached and reused for a period of time.
    ///
    /// Defaults to 300 seconds (5 minutes).
    ///
    /// Pass `None` to not specify a `max-age`.
    ///
    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -120,6 +159,16 @@ impl ServeDir {
...
+    pub fn max_age(mut self, dur: Duration) -> Self {
+        self.max_age = dur;
+        self
+    }
```
davidpdrsn, line 171: I think it makes sense to allow users to set this to `None` if for some reason they don't wanna specify a `max-age`. Could also set it to `Duration::ZERO` but I think an explicit `Option<Duration>` is more clear.

```suggestion
    pub fn max_age(mut self, max_age: Option<Duration>) -> Self {
        self.max_age = max_age;
        self
    }
```

This'll require some more changes elsewhere.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -228,7 +278,7 @@ fn append_slash_on_path(uri: Uri) -> Uri {
 }
 
 enum Output {
-    File(File, HeaderValue, usize),
+    File(File, HeaderValue, usize, Duration),
```
davidpdrsn, line 281: Given the number of fields of this variant I think we should move to named fields:

```suggestion
    File {
        file: File,
        mime: HeaderValue,
        buf_chunk_size: usize,
        max_age: Option<Duration>,
    },
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -251,38 +301,45 @@ impl Future for ResponseFuture {
...
+                        }
+                    };
                 let body = AsyncReadBody::with_capacity(file, chunk_size).boxed();
                 let body = ResponseBody(body);
 
                 let mut res = Response::new(body);
                 res.headers_mut().insert(header::CONTENT_TYPE, mime);
+                res.headers_mut().insert(
+                    header::CACHE_CONTROL,
+                    HeaderValue::from_str(&format!("max-age={}", max_age.as_secs())).unwrap(),
```
davidpdrsn, line 341: To avoid allocating a string for each response (only to create a `HeaderValue`) I think should instead change the field from a `Duration` (or rather `Option<Duration>`) to `Option<HeaderValue>` and then generate the header value directly when the user sets it. Cloning a `HeaderValue` is cheaper since its reference counted.

tower-http/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- `ServeDir`: Add cache control feature: max-age default to 300 secs
```
davidpdrsn, line 10: This should contain a link to the PR (like other entries in the changelog)

davidpdrsn (changes requested):
Thanks! I had a few comments. Let me know if you have any questions 😊

c5soft:
Closed and PR again

PR #168, opened 2021-11-16
Redo PR:ServeDir: Add optional cache  control feature: max-age . Based on 0.1.2

c5soft:
Add cache control feature: max-age=300 as default, test code included


tower-http/CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
-- None.
+- PR: `ServeDir`: Add cache control feature: max-age default to 300 secs, from [c5soft](https://github.com/c5soft/tower-http)
```
davidpdrsn, line 10: Plase format changelog items similarly to the other items.

```suggestion
- `ServeDir`: Support setting `Cache-Control: max-age` directive ([#168])

[#168]: https://github.com/tower-rs/tower-http/pull/168
```

tower-http/Cargo.toml
```diff
@@ -16,7 +16,7 @@ bytes = "1"
 futures-core = "0.3"
 futures-util = { version = "0.3", default_features = false, features = [] }
 http = "0.2"
-http-body = "0.4.1"
+http-body = "0.4"
```
davidpdrsn, line 19: This needs to remain 0.4.1 since tower-http uses features shipped in 0.4.1. If we only depend on 0.4 a user might get 0.4.0 which would lead to compile errors.

```suggestion
http-body = "0.4.1"
```

tower-http/Cargo.toml
```diff
@@ -42,7 +42,7 @@ hyper = { version = "0.14", features = ["full"] }
 once_cell = "1"
 tokio = { version = "1", features = ["full"] }
 tower = { version = "0.4.10", features = ["buffer", "util", "retry", "make"] }
-tracing-subscriber = "0.2"
+tracing-subscriber = "0.3"
```
davidpdrsn, line 45: Updating this will also require adding the `env-filter` feature. I don't think we should update it in this PR. Lets do that in another PR.

```suggestion
tracing-subscriber = "0.2"
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -120,6 +124,16 @@ impl ServeDir {
         self.buf_chunk_size = chunk_size;
         self
     }
+
+    /// Cache-Control:max-age=300.
+    ///
+    /// This speeds up user's feeling for static sites, especially large files.
+    ///
+    /// Defaults to 300 seconds.
+    pub fn max_age(mut self, dur: Option<Duration>) -> Self {
```
davidpdrsn, line 133: Actually I think this will be more ergonomic.

```suggestion
    pub fn max_age(mut self, max_age: impl Into<Option<Duration>>) -> Self {
```

Hyper does the same for configs.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -120,6 +124,16 @@ impl ServeDir {
         self.buf_chunk_size = chunk_size;
         self
     }
+
+    /// Cache-Control:max-age=300.
+    ///
+    /// This speeds up user's feeling for static sites, especially large files.
+    ///
+    /// Defaults to 300 seconds.
```
davidpdrsn, line 132:
```suggestion
    /// Set the [`Cache-Control: max-age`][mdn] response directive
    ///
    /// This indicates that the response can be cached and reused for a period of time.
    ///
    /// Defaults to 300 seconds (5 minutes).
    ///
    /// Pass `None` to not specify a `max-age`.
    ///
    /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -183,7 +198,17 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
                 });
 
             let file = File::open(full_path).await?;
-            Ok(Output::File(file, mime, buf_chunk_size))
+            let cache_control = if let Some(max_age) = max_age {
+                Some(HeaderValue::from_str(&format!("max-age={}", max_age.as_secs())).unwrap())
```
davidpdrsn, line 202: Like I wrote [here](https://github.com/tower-rs/tower-http/pull/163#discussion_r749189630)

> To avoid allocating a string for each response (only to create a `HeaderValue`) I think should instead change the field from a `Duration` (or rather `Option<Duration>`) to `Option<HeaderValue>` and then generate the header value directly when the user sets it. Cloning a `HeaderValue` is cheaper since its reference counted.

So this needs to happen in the `max_age` builder method, not here.

davidpdrsn (changes requested):
Thanks!

Although, why did you close https://github.com/tower-rs/tower-http/pull/163? Seems it would have been easier to continue the review there so I didn't have to copy/paste previous comments back and forth.

c5soft:
Sir, i am beginner to git, I added one more feature  to ServeDir, but don't know how to PR again without delete the prevouse one.  [https://github.com/c5soft/tower-http](https://github.com/c5soft/tower-http)
```
   /// If path.file.ext and path.file.ext.gz(.br) coexists retrieve compressed file instend.
    ///
    /// This is will speed up file transfering.
    ///
    /// gzip/brotli file can be generated by gzipper:
    /// gzipper c --exclude jpg,png,ico --threshold 10240 public
    ///
    /// Defaults to `true`.
    pub fn lookup_compressed_file(mut self, is_lookup: bool) -> Self {
        self.lookup_compressed_file = is_lookup;
        self
    }
     ...
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -113,13 +118,37 @@ impl ServeDir {
         self
     }
 
+    /// If path.file.ext and path.file.ext.gz(.br) coexists retrieve compressed file instend.
```
Nehliin, line 121: This has already been implemented in #156

davidpdrsn:
Actually it just occurred to me that we have [`SetResponseHeader`](https://docs.rs/tower-http/0.1.2/tower_http/set_header/response/index.html) which can be used instead of this. Since its just a static header that is always set and has on corresponding logic. Therefore I don't think we should add this as a feature to `ServeDir`.

@c5soft What do you think? Let us know if you have questions about using [`SetResponseHeader`](https://docs.rs/tower-http/0.1.2/tower_http/set_header/response/index.html).

c5soft:
Good ideal, but how can I use SetResponseHeader and ServeDir dually with fallback in axum ?

davidpdrsn:
```rust
let static_files = ServeDir::new(".").handle_error(|err| {
    (
        StatusCode::INTERNAL_SERVER_ERROR,
        format!("Something went wrong: {}", err),
    )
});
let static_files = SetResponseHeader::if_not_present(
    static_files,
    header::CACHE_CONTROL,
    HeaderValue::from_static("max-age=300"),
);

let app = Router::new().nest("/", get(static_files));
```

Issue #113, opened 2021-08-03
Don't return file content for HEAD requests in ServeFile, ServeDir


jplatte:
As discussed in tokio-rs/axum#84, `ServeFile` and `ServeDir` could be improved re. handling of HEAD requests. Responses to HEAD requests should be identical to the corresponding GETs response, but without a body.

The implementation should probably use `fs::metadata` instead of opening the file to fulfill HEAD requests more efficiently.

jplatte:
I wonder whether it makes sense to even support all methods by default. Just ignoring `POST` & `PUT` bodies and returning the file contents on `TRACE` / `DELETE` / `OPTIONS` seems kind of weird. This probably applies to other services in tower-http as well.

davidpdrsn:
Fixing this will unfortunately be a breaking change since `ServeFile` implements `Service` for any type `R` rather than `Service<Request<ReqBody>>`.

davidpdrsn:
> I wonder whether it makes sense to even support all methods by default. Just ignoring POST & PUT bodies and returning the file contents on TRACE / DELETE / OPTIONS seems kind of weird. This probably applies to other services in tower-http as well.

I agree. Method handling could be improved for `ServeDir` and `ServeFile`. Off the top of my head I don't any other services have similar issues.

jplatte:
Oh I see, there's only one more service currently which is the redirect one (which probably *should* work the same regardless of HTTP method).

Nehliin:
https://github.com/tower-rs/tower-http/pull/156 Have the same required breaking change since it needs to inspect the `Accepted-Encoding` headers so I could pick this up after that PR is reviewed

PR #169, opened 2021-11-16
Properly support HEAD requests for ServeFile/ServeDir

Nehliin:
## Motivation
Fixes: https://github.com/tower-rs/tower-http/issues/113

## Solution
Looks at the metadata to determine the `Content-length` which is set as an additional header in the response. The response contains no body as expected for `HEAD` requests. It uses the same fallback strategy we use for opening files based on negotiated encoding to figure out which metadata should be read.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 # Unreleased
 
 - `ServeDir` and `ServeFile`: Ability to serve precompressed files ([#156])
+- `ServeDir` and `ServeFile`: Properly support HEAD requests which return no body and have the `Content-length` header set ([#169])
 - `Trace`: Add `DefaultMakeSpan::level` to make log level of tracing spans easily configurable ([#124])
 - Change the response body error type of `Compression` and `Decompression` to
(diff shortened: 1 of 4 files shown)


davidpdrsn (approved):
LGTM!

Is `HEAD` requests something to consider for [range request](https://github.com/tower-rs/tower-http/pull/173)? Might user agents send a `HEAD` request to check if the server supports range requests?

cc @MarcusGrass

Nehliin:
@davidpdrsn yeah I would presume that would be one of the main use cases for `HEAD` requests

MarcusGrass:
> LGTM!
> 
> Is `HEAD` requests something to consider for [range request](https://github.com/tower-rs/tower-http/pull/173)? Might user agents send a `HEAD` request to check if the server supports range requests?
> 
> cc @MarcusGrass

Yeah we do wan't that, I implemented it in read_partial then axed that when I heard that Oscar was doing something with HEAD, it's easy to add though!

davidpdrsn:
Alright. Then lets get this merged and make any necessary adjustments in https://github.com/tower-rs/tower-http/pull/173.

Nehliin:
👍 Could we merge #178 first to avoid extra conflicts?

Nehliin:
Rebased now and added some extra tests. It should now use the same fallback strategy like we use for opening files based on negotiated encoding.

davidpdrsn (approved):

PR #172, opened 2021-11-16
Add compression predicate API

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/139
Fixes https://github.com/tower-rs/tower-http/issues/146

This introduces new API for specifying "compression predicates". This allows users using the `Compression` middleware to disable compression of specific responses. Could be small responses, images, gRPC, or something completely custom.

Example from the docs:

```rust
use tower_http::compression::{
    Compression,
    predicate::{Predicate, NotForContentType, DefaultPredicate},
};
use tower::util::service_fn;

let service = service_fn(|_: ()| async {
    Ok::<_, std::io::Error>(http::Response::new(()))
});

// build our custom compression predicate
// its recommended to still include `DefaultPredicate` as part of
// custom predicates
let predicate = DefaultPredicate::new()
    // don't compress responses who's `content-type` starts with `application/json`
    .and(NotForContentType::new("application/json"));

let service = Compression::new(service).compress_when(predicate);
```


MarcusGrass (approved):
LGTM!

tower-http/src/compression/mod.rs
```diff
@@ -194,4 +195,54 @@ mod tests {
...
+        let res = svc.ready().await.unwrap().call(req).await.unwrap();
+
+        // read the compressed body
+        let mut body = res.into_body();
+        let mut data = BytesMut::new();
+        while let Some(chunk) = body.data().await {
+            let chunk = chunk.unwrap();
+            data.extend_from_slice(&chunk[..]);
+        }
+        assert!(String::from_utf8(data.to_vec()).is_err());
```
Nehliin, line 281: If we want to check that it actually compresses to the expected string I think we have dev dependencies to actually decompress brotli stuff synchronously after my ServeFile/Dir changes. If for whatever reason this starts returning garbage data instead of the expected compressed data. But its not important just an FYI

Nehliin (approved):
:+1: Looks good to me!

NeoLegends (approved):
Really like the API here!

Nehliin (approved):

tower-http/src/compression/predicate.rs
```diff
@@ -0,0 +1,242 @@
...
+//! [`CompressionLayer::compress_when`]: super::CompressionLayer::compress_when
+
+use http::{header, Extensions, HeaderMap, StatusCode, Version};
+use http_body::Body;
+use std::{fmt, sync::Arc};
+
+/// Predicate used to determine if a response should be compressed or not.
+pub trait Predicate: Clone {
+    /// Should this response be compressed or not?
+    fn should_compress<B>(&self, response: &http::Response<B>) -> bool
```
davidpdrsn, line 16: Note to self: This should probably take `&mut self` as per [`tower::filter::Predicate`](https://docs.rs/tower/0.4.11/tower/filter/trait.Predicate.html)

Issue #128, opened 2021-08-24
Handle Range Requests in ServeDir


theduke:
## Feature Request

`ServeDir` currently does not respect [Range headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests).

This is pretty important for serving larger files like videos, or for allowing browsers to resume downloads.

davidpdrsn:
We can probably get inspiration from [warp's implementation](https://github.com/seanmonstar/warp/blob/master/src/filters/fs.rs).

theduke:
I also remember the `actix-web` implementation to be pretty solid.

I think it does a few more things like adding `content-disposition` headers.

theduke:
Implementation is here: https://github.com/actix/actix-web/tree/master/actix-files/src

Quite a bit more advanced and configurable.

davidpdrsn:
I don't mind changing the internals, as long as we don't introduce breaking changes.

theduke:
I'll draft a PR.
I already refactored the actix-web code once for a warp project.

If it's too similar it would mean adding their license though, I reckon.

Ideally actix-web could somehow be convinced to share the implemenation...

davidpdrsn:
> If it's too similar it would mean adding their license though, I reckon.

For sure.

> Each project implementing this from scratch, usually in an incomplete fashion, was one of my big annoyances with the state of Rust frameworks.

Thats the whole purpose of tower-http. To provide building blocks that aren't framework specific :)

apps4uco:
Hi, Im very interested in this. Would it be possible to consider in the api, not just reading from files, but reading from database BLOBs/CLOBs, sled database and block caches like bcache and nbd.

davidpdrsn:
Is there one way to read from all those things or would we have to have different implementations for thing to read from?

programatik29:
Anyone working on this?

davidpdrsn:
I don't believe so.

apps4uco:
I found a crate that could be adapted/incorporated for this (dependencies include http, http-body)

<https://crates.io/crates/http-serve>

"Rust helpers for serving HTTP GET and HEAD responses with hyper 0.14.x and tokio.

This crate supplies two ways to respond to HTTP GET and HEAD requests:

the serve function can be used to serve an Entity, a trait representing reusable, byte-rangeable HTTP entities. Entity must be able to produce exactly the same data on every call, know its size in advance, and be able to produce portions of the data on demand"

[...]

PR #173, opened 2021-11-16
Support `Range` requests for `ServeDir`

MarcusGrass:
Adresses Handle Range Requests in ServeDir #128 with some outstanding questionmarks mentioned in that issue


tower-http/src/services/fs/read_stream.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let reader = match this.reader.as_pin_mut() {
+            Some(r) => r,
+            None => return Poll::Ready(None),
+        };
+        if this.buf.capacity() == 0 {
+            this.buf.reserve(*this.capacity);
+        }
+
+        match poll_read_buf(reader, cx, &mut this.buf) {
```
Nehliin, line 83: IMO It would be nice with some small comments in this going through the expected poll flow with regards to reuse of buffer etc.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -192,6 +203,32 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
...
+    if let Ok(as_str) = range_header.to_str() {
+        // produces ("bytes", "0-1023")
+        return if let Some((_, range)) = as_str.split_once("=") {
+            // produces ("0", "1023")
+            if let Some((start_incl, end_incl)) = range.split_once("-") {
+                if let Ok(start) = start_incl.parse() {
+                    if let Ok(end) = end_incl.parse() {
+                        return Some(ParsedRangeHeader::Valid(ContentRange {
+                            start_inclusive: start,
+                            end_inclusive: end
```
Nehliin, line 218: Do we check start < end somewhere?

MarcusGrass: Not explicitly, it's done here https://github.com/tower-rs/tower-http/blob/dfd3e04059b9672bbf71521854b040c08ef5cbab/tower-http/src/services/fs/serve_dir.rs#L331 and would panic during the body creation, might not be what we want, I also would like an ending range-value to be an invalid parse, it's a good suggestion to check beforehand!

MarcusGrass: Fixed, but the parsing leaves a lot to be desired and should potentially be reworked

Nehliin: :+1: Yeah its nice to avoid panics that can be caused by external clients

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -192,6 +203,34 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
...
+                }
+            }
+            // Range header at least on form bytes=<some_range> should produce a 416
+            Some(ParsedRangeHeader::Unsatisfiable)
+        } else {
+            Some(ParsedRangeHeader::Malformed("Range header value not in format `bytes=<range>`, `=` missing".to_owned()))
+        }
+    }
+    Some(ParsedRangeHeader::Malformed("Range not parseable as utf8".to_owned()))
+}
```
davidpdrsn, line 232: Wondering if there is way to write this using the fact that `?` works on `Option` as well 🤔

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -192,6 +203,34 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
     }
 }
 
+fn parse_range(headers: &HeaderMap) -> Option<ParsedRangeHeader> {
+    // ex:  `Range: bytes=0-1023`
+    let range_header = headers.get(header::RANGE)?;
+    if let Ok(as_str) = range_header.to_str() {
+        // produces ("bytes", "0-1023")
+        return if let Some((_, range)) = as_str.split_once("=") {
```
davidpdrsn, line 211: Should we check that the stuff to the left of `=` is actually `bytes`?

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -192,6 +203,34 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
...
+                    if let Ok(end) = end_incl.parse() {
+                        if start < end {
+                            return Some(ParsedRangeHeader::Valid(ContentRange {
+                                start_inclusive: start,
+                                end_inclusive: end
+                            }));
+                        }
+                    }
+                }
+            }
```
davidpdrsn, line 224: Might be good to run some tests with resuming downloads from a browser and see the actual header values that are sent and make sure this handles all of them.

We talked about the many forms the `Range` header can take so might good to make sure we cover the necessary cases.

davidpdrsn: Should probably also have some unit tests specifically for this function.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -228,11 +267,37 @@ fn append_slash_on_path(uri: Uri) -> Uri {
...
+struct ContentRange {
+    start_inclusive: u64,
+    end_inclusive: u64,
+}
+
+impl ContentRange {
+    fn range_size(&self) -> u64 {
+        self.end_inclusive - self.start_inclusive + 1
+    }
+}
```
davidpdrsn, line 299: Could we use https://doc.rust-lang.org/stable/std/ops/struct.RangeInclusive.html instead of this?

tower-http/src/services/fs/read_stream.rs
```diff
@@ -0,0 +1,143 @@
...
+        // This value is `None` if the stream has terminated.
+        #[pin]
+        reader: Option<R>,
+        // Working buffer, used to optimize allocations.
+        buf: BytesMut,
+        capacity: usize,
+        max_read_bytes: Option<usize>,
+        read_bytes: usize,
+    }
+}
```
davidpdrsn, line 52: I've just noticed that `AsyncReadExt` has a [`take`](https://docs.rs/tokio/1.14.0/tokio/io/trait.AsyncReadExt.html#method.take) adapter.

> Creates an adaptor which reads at most `limit` bytes from it.

Do you think we could use that instead of this custom version of `ReaderStream`?

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -251,8 +316,44 @@ impl Future for ResponseFuture {
...
+                                        .status(StatusCode::BAD_REQUEST)
+                                        .body(body_from_bytes(Bytes::from(reason)))
+                                }
+                            }
+                        } else {
+                            let body = AsyncReadBody::with_capacity(file_request.file, file_request.chunk_size).boxed();
+                            let body = ResponseBody(body);
+                            builder.body(body)
+                        }
+                    },
```
davidpdrsn, line 356: I recommend adding a git pre-push hook that reminds you to run `cargo fmt` 😅

davidpdrsn (commented):
Good progress so far! I left some suggestions and questions.

MarcusGrass:
Thinking about offloading the read-ranges part to its own crate and feature-gate debug info

Nehliin (commented):
Nice! only had a few comments

Nehliin:
> Thinking about offloading the read-ranges part to its own crate and feature-gate debug info

Yeah the parsing of ranges could maybe be uplifted to some generic http crate. 

Do you mean the detaild error messages when you say debug-info? I think that could make sense since I'm not sure how detailed error messages we need for something that comes form the HTTP spec. The different status codes already tell you a lot and it means we can avoid some allocations.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -245,9 +251,25 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
...
+            let (mut file, maybe_encoding) =
                 open_file_with_fallback(full_path, negotiated_encodings).await?;
-            Ok(Output::File(file, mime, buf_chunk_size, maybe_encoding))
+            let file_size = file.metadata().await?.len();
+            let maybe_range = range_header.map(|range| parse_range(&range, file_size));
+            if let Some(ParsedRangeHeader::Range(ranges)) = maybe_range.as_ref() {
+                if ranges.len() == 1 {
+                    file.seek(SeekFrom::Start(*ranges[0].start()))
+                        .await
+                        .unwrap();
```
Nehliin, line 262: Could we avoid this unwrap here?

tower-http/src/services/fs/mod.rs
```diff
@@ -138,6 +139,16 @@ where
...
         }
     }
+
+    fn with_capacity_limited(
+        read: T,
+        capacity: usize,
+        max_read_bytes: u64,
+    ) -> AsyncReadBody<Take<T>> {
+        AsyncReadBody {
+            reader: ReaderStream::with_capacity(read.take(max_read_bytes as u64), capacity),
```
davidpdrsn, line 149: `max_read_bytes` seems to already be a `u64`

```suggestion
            reader: ReaderStream::with_capacity(read.take(max_read_bytes), capacity),
```

tower-http/src/services/fs/parse_range.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+    } else {
+        return ParsedRangeHeader::Malformed(format!(
+            "Range: {} is not acceptable, range does not start with 'bytes='",
+            range
+        ));
+    }
+    if ranges.is_empty() {
+        panic!("Programming error parsing range {}", range)
+    } else {
```
davidpdrsn, line 80: We should aim for `tower-http` _never_ to panic because of user input. If someone can make the server panic by sending a specific header thats technically an attack vector that can be used to do DoS.

tower-http/src/services/fs/parse_range.rs
```diff
@@ -0,0 +1,309 @@
...
+    for range in ranges {
+        bounds.push((range.start(), range.end()));
+    }
+    for i in 0..bounds.len() {
+        for j in i + 1..bounds.len() {
+            if bounds[i].0 <= bounds[j].1 && bounds[j].0 <= bounds[i].1 {
+                return true;
+            }
+        }
+    }
```
davidpdrsn, line 105: I'm wondering if there is a way to write these loops without looping over the indexes 🤔

Something like:

```rust
for x in &thing {
    for y in &thing {
    }
}
```

That would also avoid potential panics from `bounds[i]` being out of bounds (no pun intended 😜)

tower-http/src/services/fs/parse_range.rs
```diff
@@ -0,0 +1,309 @@
...
+    false
+}
+
+fn split_once<'a>(s: &'a str, pat: &'a str) -> Option<(&'a str, &'a str)> {
+    let mut iter = s.split(pat);
+    let left = iter.next();
+    left.and_then(|l| iter.next().map(|r| (l, r)))
+}
```
davidpdrsn, line 113: `?` works on `Option` so this should work:

```suggestion
fn split_once<'a>(s: &'a str, pat: &'a str) -> Option<(&'a str, &'a str)> {
    let mut iter = s.split(pat);
    let left = iter.next()?;
    let right = iter.next()?;
    Some((left, right))
}
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -376,6 +470,13 @@ fn empty_body() -> ResponseBody {
     ResponseBody::new(body)
 }
 
+fn body_from_bytes(bytes: Bytes) -> ResponseBody {
+    let body = BoxBody::new(Full::from(bytes))
+        .map_err(|err| match err {})
+        .boxed();
```
davidpdrsn, line 476: The initial `BoxBody::new` shouldn't be necessary since there is a `.boxed()` at the end.

```suggestion
    let body = Full::from(bytes)
        .map_err(|err| match err {})
        .boxed();
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -245,9 +251,25 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
...
                 });
 
-            let (file, maybe_encoding) =
+            let (mut file, maybe_encoding) =
                 open_file_with_fallback(full_path, negotiated_encodings).await?;
-            Ok(Output::File(file, mime, buf_chunk_size, maybe_encoding))
+            let file_size = file.metadata().await?.len();
+            let maybe_range = range_header.map(|range| parse_range(&range, file_size));
+            if let Some(ParsedRangeHeader::Range(ranges)) = maybe_range.as_ref() {
+                if ranges.len() == 1 {
```
davidpdrsn, line 259: Is it correct that this means if the client asks for multiple ranges we'll only return the first one? Not an expert on the spec for `Range` but is that allowed? Would returning some sort of "range not satisfiable" status be more correct?

I remember that supporting multiples ranges is quite annoying so I'm fine with not supporting it. Its something we should document regardless.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -315,49 +346,112 @@ impl Future for ResponseFuture {
...
                         }
-
-                        Err(err) => {
-                            return Poll::Ready(
-                                super::response_from_io_error(err)
-                                    .map(|res| res.map(ResponseBody::new)),
+                        if let Some(range) = file_request.maybe_range {
+                            match range {
+                                ParsedRangeHeader::Range(ranges) => {
+                                    let range = &ranges[0];
```
davidpdrsn, line 362: Is there a way to write this such it'll never panic? I'm always scared of stuff like this if we refactor the code someday in a way that `ranges` can be empty here.

tower-http/src/services/fs/serve_dir.rs
davidpdrsn: Edit: Ah I see you're handling multipart ranges further down. Adding a comment about that here would be fine I think.

davidpdrsn (commented):
Really great work on this! Getting closer 😊

tower-http/src/services/fs/parse_range.rs
```diff
@@ -0,0 +1,309 @@
...
+                        range
+                    ));
+                }
+                if let Ok(start) = start.parse::<u64>() {
+                    if end == "" {
+                        ranges.push(RangeInclusive::new(start, file_size_bytes - 1));
+                        continue;
+                    }
+                    if let Ok(end) = end.parse::<u64>() {
+                        if end >= file_size_bytes {
```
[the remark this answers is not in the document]
MarcusGrass: The end is an inclusive index so end == file_size_bytes would be an overflow

tower-http/src/services/fs/parse_range.rs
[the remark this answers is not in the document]
MarcusGrass: I had it as split_once but CI complains that it was added too soon sadly

tower-http/src/services/fs/parse_range.rs
```diff
@@ -0,0 +1,309 @@
...
+            if sep_count != 1 {
+                return ParsedRangeHeader::Malformed(format!(
+                    "Range: {} is not acceptable, contains multiple dashes (-).",
+                    range
+                ));
+            }
+            if let Some((start, end)) = split_once(range, "-") {
+                if start == "" {
+                    if let Ok(end) = end.parse::<u64>() {
+                        if end >= file_size_bytes {
```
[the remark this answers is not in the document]
MarcusGrass: same as below

tower-http/src/services/fs/parse_range.rs
```diff
@@ -0,0 +1,309 @@
...
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum ParsedRangeHeader {
+    Range(Vec<RangeInclusive<u64>>),
+    Unsatisfiable(String),
+    Malformed(String),
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
MarcusGrass: True, speaks even more to offloading it

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -245,9 +251,25 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
                     HeaderValue::from_str(mime::APPLICATION_OCTET_STREAM.as_ref()).unwrap()
                 });
 
-            let (file, maybe_encoding) =
+            let (mut file, maybe_encoding) =
                 open_file_with_fallback(full_path, negotiated_encodings).await?;
-            Ok(Output::File(file, mime, buf_chunk_size, maybe_encoding))
+            let file_size = file.metadata().await?.len();
```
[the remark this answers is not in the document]
MarcusGrass: Yeah, it would be nice to get the size of the file on every request

tower-http/src/services/fs/serve_dir.rs
MarcusGrass: Woops, yeah, should be ?

tower-http/src/services/fs/parse_range.rs
MarcusGrass: Yeah, this should(TM) be impossible though as a bad header should always be a Malformed or Unsatisfiable, any valid header contains a range, scratching my head over this as something needs to be returned but if it's a Malformed then that's misleading because that should have been caught earlier, Unsatisfiable as well. If we end up here it's because I wrote a bug

tower-http/src/services/fs/parse_range.rs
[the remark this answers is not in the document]
davidpdrsn: Good point about fuzzing! [`headers`](https://docs.rs/headers/0.3.5/headers/struct.Range.html) has some `Range` parsing but dunno if it supports everything we need. It also pulls in a bunch of dependencies that we might not want.

tower-http/src/services/fs/parse_range.rs
davidpdrsn: Returning `Malformed` with the body saying something like "something went wrong in tower-http, please file an issue" is fine imo.

I'm mostly worried about this code getting refactored in the future such that this is no longer unreachable.

tower-http/src/services/fs/parse_range.rs
MarcusGrass: I was thinking about that but we shouldn't be able to go out of bounds here as `i` and `j` are both constricted, additionally the unconstrained double loob will make a comparison that is in the worst case n*n, while the constrained one will be a lot fewer, 
for non overlapping ranges a, b, and c we in that case get 
`aa, ab, ac,

ba, bb, bc,

ca, cb, cc`

vs

` ab, ac,

  bc`

tower-http/src/services/fs/mod.rs
MarcusGrass: Yeah, missed removing that when migrating from usize -> u64, nice catch

tower-http/src/services/fs/parse_range.rs
davidpdrsn: Hm 🤔 This code is only needed if the client sends multiple ranges right? If so, could the parsing return early and avoid these checks all together?

davidpdrsn:
I think it should be nice to also support ranges for `ServeFile` though that can be done in a follow up PR.

tower-http/src/services/fs/parse_range.rs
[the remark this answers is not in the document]
jplatte: Maybe the `headers` crate would be open to feature-gating some functionality in the next breaking-change release.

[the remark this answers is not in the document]
davidpdrsn: Yeah. The three of us work at embark so we've been chatting about it. We've decided to try switching to using `headers` and possibly contributing back improvements if we find any.

tower-http/src/services/fs/parse_range.rs
MarcusGrass: If the length of ranges is 1 it doesn't check overlap

MarcusGrass:
On my quest to parse headers I found some other crates doing it, one small and possibly abandoned that has some deps https://github.com/DCjanus/range_header
[Rust-http-range](https://github.com/bancek/rust-http-range) is more used [crates io](https://crates.io/crates/http-range).

It's faster, benching on the same inputs i get about 2.5x faster on that testing a few different normal range inputs.

It will however panic extremely quickly when fuzzing, less than half a second of running.

Additionally it deemes and empty range to be correct, some examples:
```
`bytes=				` => `[]`
`bytes= ,  ` => `[]`
`bytes=, , ` => `[]`
`bytes=-7  ` => `[HttpRange { start: 993, length: 7 }]`
`bytes= - 1` => `[HttpRange { start: 999, length: 1 }]`
`bytes=04-,` => `[HttpRange { start: 4, length: 996 }]`
`bytes=-44, ` => `[HttpRange { start: 956, length: 44 }]`
`bytes=-97  ` => `[HttpRange { start: 903, length: 97 }]`
`bytes=7-7, ` => `[HttpRange { start: 7, length: 1 }]`
`bytes=, ,, ` => `[]`
`bytes=4-06,` => `[HttpRange { start: 4, length: 3 }]`
`bytes=7-,4-,` => `[HttpRange { start: 7, length: 993 }, HttpRange { start: 4, length: 996 }]`
`bytes=, , , ` => `[]`
`bytes=7-				` => `[HttpRange { start: 7, length: 993 }]`
```
I guess that's a matter of preference, whether an empty range is satisfiable but according to my understanding of the spec it isn't.
Additionally:
`bytes=04-,` => `[HttpRange { start: 4, length: 996 }]`
Should probably not be valid either.

It's hard to know if any other strangeness in parsing could be found with fuzzing since it crashes so soon.

Per our earlier discussion the headers library suffers from many of the same issues.

I could try to fixup that project but it's a string parser with no tests and I really just want to be done with this before my head explodes so I just pushed my implementation that I know works

davidpdrsn:
We've agreed to move forward with @MarcusGrass's parsing implementation since it is stable (doesn't crash on bad input) and supports our use cases. Also agreed to move the crate into the tower-http workspace and give publish permissions to more people. Once thats done we can finally merge this 🥳

http-range-header/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "http-range-header"
+version = "0.3.0"
+edition = "2018"
+license = "MIT"
+readme = "./README.md"
+description = "No-dep range header parser"
+repository = "https://github.com/MarcusGrass/parse-range-headers"
+homepage = "https://github.com/MarcusGrass/parse-range-headers"
```
Nehliin, line 9: Old links?

http-range-header/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "http-range-header"
+version = "0.3.0"
+edition = "2018"
```
Nehliin, line 4: 2021 instead?

Nehliin (commented):
Haven't looked through everything yet but looks very nice!

http-range-header/Cargo.toml
davidpdrsn: Can't go to 2021 because we still support older rust versions :(

Nehliin: ah right was thinking we could still keep tower-http on 2018 but ofc you wouldn't be able to compile it then with older versions of rust.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -347,6 +375,16 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
     }
 }
 
+fn try_parse_range(
+    maybe_range_ref: Option<&String>,
```
Nehliin, line 379: &str instead?

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -406,58 +456,118 @@ impl Future for ResponseFuture {
...
+                            .header(header::ACCEPT_RANGES, "bytes")
+                            .header(header::CONTENT_LENGTH, size.to_string());
+                        if let Some(encoding) = file_request.maybe_encoding {
+                            builder = builder
+                                .header(header::CONTENT_ENCODING, encoding.into_header_value());
+                        }
+                        if let Some(reasonable_range) = file_request.maybe_range {
+                            match reasonable_range {
+                                Ok(ranges) => {
+                                    if let Some(range) = ranges.get(0) {
```
Nehliin, line 477: .first() instead of .get(0) perhaps?

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -406,58 +456,118 @@ impl Future for ResponseFuture {
...
+                            .header(header::CONTENT_TYPE, file_request.mime_header_value)
+                            .header(header::ACCEPT_RANGES, "bytes")
+                            .header(header::CONTENT_LENGTH, size.to_string());
+                        if let Some(encoding) = file_request.maybe_encoding {
+                            builder = builder
+                                .header(header::CONTENT_ENCODING, encoding.into_header_value());
+                        }
+                        if let Some(reasonable_range) = file_request.maybe_range {
+                            match reasonable_range {
+                                Ok(ranges) => {
```
Nehliin, line 476: To avoid some of the extra layers of the extra conditionals it might be worth matching on `Ok(Some(range))` and `Ok(None)`

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -406,58 +456,118 @@ impl Future for ResponseFuture {
...
+                        };
+                        let mut builder = Response::builder()
+                            .header(header::CONTENT_TYPE, file_request.mime_header_value)
+                            .header(header::ACCEPT_RANGES, "bytes")
+                            .header(header::CONTENT_LENGTH, size.to_string());
+                        if let Some(encoding) = file_request.maybe_encoding {
+                            builder = builder
+                                .header(header::CONTENT_ENCODING, encoding.into_header_value());
+                        }
+                        if let Some(reasonable_range) = file_request.maybe_range {
```
Nehliin, line 474: would be nice to maybe break this part out to a separate method to make the poll method a bit shorter and easier to overlook.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -329,14 +336,35 @@ impl<ReqBody> Service<Request<ReqBody>> for ServeDir {
 
             match request_method {
                 Method::HEAD => {
-                    let (metadata, maybe_encoding) =
+                    let (meta, maybe_encoding) =
                         file_metadata_with_fallback(full_path, negotiated_encodings).await?;
-                    Ok(Output::FileMetadata(metadata, mime, maybe_encoding))
+                    let maybe_range = try_parse_range(range_header.as_ref(), meta.len());
```
Nehliin, line 341: I think it would be odd for a client to send request headers in a HEAD request but I guess it doesn't hurt to support

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -914,4 +1029,86 @@ mod tests {
...
+            .uri("/README.md")
+            .header(
+                "Range",
+                format!("bytes={}-{}", bytes_start_incl, requested_len - 1),
+            )
+            .body(Body::empty())
+            .unwrap();
+        let res = svc.oneshot(req).await.unwrap();
+
+        assert_eq!(res.status(), StatusCode::RANGE_NOT_SATISFIABLE);
```
Nehliin, line 1088: Same for the other test cases but do we want to check that we return the expected content range header here even if its not satisfiable? Since we always seem to set that in the service

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -406,58 +456,118 @@ impl Future for ResponseFuture {
...
-                ),
-            },
+                    Err(err) => {
+                        return Poll::Ready(
+                            super::response_from_io_error(err)
+                                .map(|res| res.map(ResponseBody::new)),
+                        )
+                    }
+                };
+                Poll::Ready(Ok(response.unwrap()))
```
Nehliin, line 569: Is it worth returning the match directly to avoid having to have this Poll::Ready so far down? Maybe not a problem if the range stuff gets broken out to separate function

Nehliin (changes requested):
Great work! I left some comments but you don't have to fix all of them. The only thing I'd like to see is breaking out the range stuff from the poll method so its a bit easier to overlook and isn't as long.

tower-http/src/services/fs/serve_dir.rs
MarcusGrass: Won't compile, takes an `Option<String>::as_ref() -> Option<&String>`

tower-http/src/services/fs/serve_dir.rs
MarcusGrass: It's kind of odd but per our discussion, head requests should return the same headers as a get without the body, and we only check this if the client actual requests range in their HEAD

davidpdrsn:
Why did you move the crate out of the workspace again? I still think it makes sense to have in tree since its an important part of the feature.

tower-http/src/services/fs/serve_dir.rs
davidpdrsn: You can use [`Option::as_deref`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.as_deref)

MarcusGrass:
> Why did you move the crate out of the workspace again? I still think it makes sense to have in tree since its an important part of the feature.

I couldn't get it to pass CI and couldn't bring myself to struggle more with this PR, thinking about creating an ietf2616 compliant, no-dependency, no-panic header parsing lib

tower-http/src/services/fs/serve_dir.rs
davidpdrsn: To it feels like something we should support

tower-http/src/services/fs/serve_dir.rs
davidpdrsn: These kinds of small refactorings we can always make. I don't think this should be blocking.

davidpdrsn (approved):

PR #174, opened 2021-11-17
ServeDir/File: add check_precompressed_file to serve precompressed files

c5soft:
I'm wondering if it is a design problem that ServeDir/File will fail when precompressed_gzip enabled and path/file.ext.gz not exists but path/file.ext exists, this should works without content-encoding. **It is very common that small files and image/movie files should remaind uncompressed(only uncompressed files exist) while large files have precompressed and uncompressed copies coexist.** [This PR is my solution to fix this problem](https://github.com/tower-rs/tower-http/pull/174).

### DIFF
--- a/test-files/only_uncompressed.txt
+++ b/test-files/only_uncompressed.txt
@@ -1,0 +1,1 @@
+"This is a test file!"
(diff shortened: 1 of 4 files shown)


Nehliin:
Ah looks like we both did the same thing (#175). Unfortunately `.exists()` relies on `std::fs::metadata` which is blocking which we would like to avoid. But thanks for the pr even though we might not use this specifically and especially big thanks for bringing up the issue in the first place!

c5soft:
Very good, your solution is better !

PR #175, opened 2021-11-17
Fallback to uncompressed file version if precompressed can't be found

Nehliin:
## Motivation

Prevents 404s when `.precompressed_gzip()` and family is specified but no precompressed version is found. Mostly useful for the `ServeDir` use case.

## Solution
Simply remove the extension and retry if an encoding is specified but no file is found.

### DIFF
--- a/test-files/missing_precompressed.txt
+++ b/test-files/missing_precompressed.txt
@@ -1,0 +1,1 @@
+Test file!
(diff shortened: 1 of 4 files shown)


davidpdrsn (approved):
Nice!

c5soft:
Great! But one exception should be considered, that is, when br and gzip both enabled and client-accepted, negotiated_encoding may be gzip where path/file.ext.gz not exists, this  should not be fallback to uncompressed but br.

davidpdrsn:
That sounds like a good idea to me 👍

PR #178, opened 2021-11-19
Fallback to other precompressed files if one variant isn't found

Nehliin:
## Motivation
If for example a given ServeFile/Dir have `precompressed_gzip` and `precompressed_br` set but only the `br` variant actually exists, and the calling client supports both encodings, it will now try to fallback to `br` instead of the uncompressed file variant.  

## Solution
Provide all possible negotiated encodings in the `open_file_with_fallback` function and try them each in the preferred order until a file is found or return a 404


c5soft:
Very good. Having being fully reviewed the new code I guess there may be one exception: when path/file.ext.gz and path/file.ext.br co-existed and actived, gzip will allways be selected for Chrome not br. as Chrome allways send "Accept-Encoding: gzip, deflate, br" not "Accept-Encoding: br, deflate, gzip" .

Nehliin:
@c5soft again, I'm not sure that would be a problem since it supports both. Is it because brotli has better compression density? If chrome would prefer brotli it should probably add weights to the encodings. (I'm sure they have reasons not to though)

davidpdrsn:
> when path/file.ext.gz and path/file.ext.br co-existed and actived, gzip will allways be selected for Chrome not br. as Chrome allways send "Accept-Encoding: gzip, deflate, br" not "Accept-Encoding: br, deflate, gzip" .

I'm gonna review this today but I don't feel its appropriate to build in overrides for specific user-agents. I chrome prefers br over gzip I guess it should send that first in `accept-encoding` 🤷

c5soft:
Chrome will send "Accept-Encoding: gzip, deflate, br" when https remote host or http local host. 
Chrome will send "Accept-Encoding: gzip, deflate" when http remote host.
I modified the algorithm of preferred_encoding in [PR177](https://github.com/tower-rs/tower-http/pull/177), let br be the first then gzip.

c5soft:
I have not noticed any browser send Accept-Encoding in format: "br;q=1,gzip;q=0.9,deflate;q=0.5", commonly no ";q=" but comma only seperated  as chrome/firefox done. Why sticked to the old algorithm ?

Nehliin:
Relying on that chrome always will put `br` last also feels a bit shaky. Changing the reading order just for chrome will break if they decide to change the order in the header.

c5soft:
Why Chrome send  "Accept-Encoding: gzip, deflate, br" not  "Accept-Encoding: br, deflate, gzip" ?
I don't know, but I noticed:  trait Iterator::max_by rule as "If several elements are equally maximum, the last element is returned".

c5soft:
One posible solution: Can we introduce a way to give the developer a choice to force a Encoding be the first one if awailable ?

davidpdrsn:
> One posible solution: Can we introduce a way to give the developer a choice to force a Encoding be the first one if awailable ?

You can already use [`MapRequest`](https://docs.rs/tower/0.4.11/tower/util/struct.MapRequest.html) to change the headers. I think its fine to require users to do that if you want custom things like this.

c5soft:
Can we do it this way ?:
```
 let static_files = ServeDir::new(static_path)
        .precompressed_br(1.0)
        .precompressed_gzip(0.8)
``

davidpdrsn:
No.

I think we should stick with what we have now and re-evaluate later if more people ask about it.

davidpdrsn (approved):

PR #180, opened 2021-11-20
WIP: Make `Trace` compatible with OpenTelemetry

davidpdrsn:
OpenTelemetry defines a bunch of [conventions for HTTP
spans][conventions]. This makes adopting those conventions easy when
using `Trace`:

```rust
.layer(TraceLayer::new_for_http().opentelemetry_server())
```

This changes the span for the request to:

- Name: HTTP request
- Fields:
  - `http.flavor`: "1.1"
  - `http.host`: "localhost:3000"
  - `http.method`: "GET"
  - `http.route`: "/users/123". Can be configured by user to instead be
  the path pattern for the route, ie `/users/:id`
  - `http.scheme`: "http". Must be set by user as middleware don't know
  this
  - `http.target`: "/users/123"
  - `http.user_agent`: "curl/7.64.1"
  - `http.client_ip`: Must be extracted by user in a "make service"
  - `http.status_code`: 200
  - `request_id`: Requires #150.
  - `otel.kind`: "server"
  - `otel.status_code`: Unset or "ERROR"
  - `exception.message`: Set by user based on response classification.
  - `exception.details`: Set by user based on response classification.

The fields are a bit different for HTTP clients hence the name
`opentelemetry_server`. I don't think supporting clients initially is
required. We can always add it later.

TODO

- [x] Add `Trace::opentelemetry_server` method. Currently it only exists
  on the layer
- [ ] Docs, including getting request ids
- [x] Example using axum and `opentelemetry_jaeger`
- [ ] Changelog
- [x] Request id integration
- [x] Implement callbacks for functions

[conventions]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md

### DIFF
--- a/examples/axum-key-value-store/Cargo.toml
+++ b/examples/axum-key-value-store/Cargo.toml
@@ -13,5 +13,10 @@ tokio = { version = "1.2.0", features = ["full"] }
 tower = { version = "0.4.5", features = ["full"] }
 tower-http = { path = "../../tower-http", features = ["full"] }
 tracing = "0.1"
-tracing-subscriber = "0.2"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 axum = "0.3"
+uuid = { version = "0.8", features = ["v4"] }
+
+opentelemetry = { version = "0.16", features = ["rt-tokio"] }
+opentelemetry-jaeger = { version = "0.15", features = ["rt-tokio"] }
+tracing-opentelemetry = "0.16"
--- a/examples/axum-key-value-store/Cargo.toml
+++ b/examples/axum-key-value-store/Cargo.toml
@@ -7,16 +7,15 @@ publish = false
 license = "MIT"
 
 [dependencies]
+axum = "0.3"
 hyper = { version = "0.14.15", features = ["full"] }
+opentelemetry = { version = "0.16", features = ["rt-tokio"] }
+opentelemetry-jaeger = { version = "0.15", features = ["rt-tokio"] }
 structopt = "0.3.21"
 tokio = { version = "1.2.0", features = ["full"] }
 tower = { version = "0.4.5", features = ["full"] }
 tower-http = { path = "../../tower-http", features = ["full"] }
 tracing = "0.1"
-tracing-subscriber = { version = "0.3", features = ["env-filter"] }
-axum = "0.3"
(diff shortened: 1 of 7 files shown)


davidpdrsn:
When https://github.com/tower-rs/tower-http/pull/181 is merged I'm gonna update it to use an opentelemetry config.

davidpdrsn:
I've removed this from the 0.2 milestone. Been thinking that it might be overly complicated and that a simpler solution should be possible. So think it needs more time to bake. Since its not a breaking change it doesn't _have_ be included in 0.2

Oliboy50:
when this will ship, I'll definitely replace the use of [warp's trace filter](https://github.com/seanmonstar/warp/blob/master/src/filters/trace.rs) with tower + tower_http in my projects 🙏 

the biggest fail for "warp's trace filter" when working with OpenTelemetry, is that it does not set `http.status_code` in spans

Oliboy50:
> `http.route`: "/users/123". Can be configured by user to instead be the path pattern for the route, ie /users/:id

@davidpdrsn It would be an amazing feature 😍 (this is so much useful to classify incoming requests by their name or "static path" instead of their `http.target` in dashboards) but I don't really get how we could do that with the current PR?

Could you give an example of usage you have in mind to go from `/users/123` to `/users/:id`? 
Is it something crate users will have to do manually or will it be possible to let tower handles it automatically?

davidpdrsn:
I don't think this will ship any time soon, definitely not in its current form. I think the implementation in this PR is too complicated. It shouldn't require implementing four traits...

The `/users/1` vs `/users/:id` issue is part of it. I'm not sure whether frameworks allow middleware to get the matched path rather than the literal request path. With axum its possible with [`MatchedPath`](https://docs.rs/axum/latest/axum/extract/struct.MatchedPath.html) but I don't know about warp.

I imagine that someday I (or someone else) will make a crate that contains an axum specific middleware. At Embark Studios we've been working on an implementation [here](https://github.com/EmbarkStudios/server-framework/blob/main/src/middleware/trace/opentelemetry.rs) but its not published to crates.io yet. Don't have a timeline on that I'm afraid.

In fact I think I'll just close this PR. I don't like having stale PR lingering.

Issue #164, opened 2021-11-15
Make sure all middleware implement `Service<Request<ReqBody>>` and require `Response = Response<ResBody>`


davidpdrsn:
For maximum flexibility in the future I think it makes sense to always implement `Service` like this:

```rust
impl<ReqBody, ResBody> Service<Request<ReqBody>> for Foo<S>
where
    S: Service<Request<ReqBody>, Response = Response<ResBody>>
```

That is they require the request type to be `http::Request<_>` and require the response of the inner service to be `http::Response<_>`.

If you only implement `Service<R>` then you don't have access to headers or extensions which might be necessary for implementing new features, same goes for the response. Case and point https://github.com/tower-rs/tower-http/pull/156 which is a breaking change. That could have been avoided had `ServeFile` implemented `Service<Request<_>>`

For 0.2 I think we should go through all services fix this.

PR #182, opened 2021-11-22
Enforce http::Request<B> and http::Response<U> for our services

Nehliin:
## Motivation

Fixes: https://github.com/tower-rs/tower-http/issues/164 and hopefully makes it less likely to require breaking changes in the future.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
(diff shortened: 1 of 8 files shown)


davidpdrsn (approved):

privettoli:
Did we consider using a trait for the request and response to provide http::Request? I have a need to have a custom request and it seems to be incompatible with the current implementation. Would replacement of http::Request with IntoHttpRequest trait be an acceptable flexible solution?

jplatte:
Please open a new issue for feature requests / discussions, instead of commenting on old PRs.

That said, I don't think this is something we will at. If you need to convert types between two layers, you should be able to do so with another layer. If it's only for request or response, not both, I'd recommend tower's map_request / map_response.

Issue #184, opened 2021-11-22
Implement `ServeFile` in terms of `ServeDir`


davidpdrsn:
`ServeFile` and `ServeDir` have quite a few things in common. They both:

- Serve static files
- Support `HEAD` requests
- Support `Range` requests (https://github.com/tower-rs/tower-http/pull/173)
- Support precompressed files
- Support `If-Not-Modified` headers (https://github.com/tower-rs/tower-http/pull/145)

All of these features basically have to be implemented twice, and any bugs found have to be fixed in both places.

I think we should investigate removing this duplication by implementing `ServeFile` as a specialized version of `ServeDir`. Since `ServeFile` doesn't care about the request URI it should work to replace the incoming requests path with the one given to `ServeFile`'s constructor and then call `ServeDir`. Then we'd always serve the same file without so much duplication.

This would add some additional overhead to `ServeFile`, such as protecting against directory traversal attacks, which technically aren't necessary. However I think a little additional overhead is worth making the implementation simpler.

I imagine `ServeFile` would be defined as

```rust
struct ServeFile {
    serve_dir: ServeDir,
    path: std::path::PathBuf,
}
```

or something like that.

PR #187, opened 2021-11-23
Make `ServieFile` specialized version of `ServeDir`

Nehliin:
## Motivation
Fixes: https://github.com/tower-rs/tower-http/issues/184 and reduces the need to implement features twice in both `ServeFile` and `ServeDir`.

## Solution
This implementaiton specializes `ServeDir` to avoid unnecessary overhead for the `ServeFile` usecase. `ServeFile` now simply wraps a `ServeDir` service.


tower-http/src/services/fs/serve_dir.rs
```diff
@@ -92,9 +92,53 @@ use tower_service::Service;
...
+// with almost no overhead
+#[derive(Clone, Debug)]
+enum ServeVariant {
+    Directory {
+        append_index_html_on_directories: bool,
+    },
+    SingleFile {
+        mime: HeaderValue,
+    },
+}
```
davidpdrsn, line 112: This is a great idea! Much simpler than the design I had in mind 👍

davidpdrsn (approved):

tower-http/src/services/fs/serve_dir.rs
Nehliin: Thanks!

c5soft:
Nice! I have to give up my implementation of merging work.

PR #192, opened 2021-11-26
Simplify `AuthorizeRequest` and `AsyncAuthorizeRequest` traits

### LINKED ISSUE #190: Tell `AuthorizeRequest::unauthorized_response` why the request could not be authorized
### Motivation

The current authorization services returns an `Option` ([see docs](https://docs.rs/tower-http/0.1.2/tower_http/auth/trait.AsyncAuthorizeRequest.html) for `AsyncAuthorizeRequest`). This works but has one major flaw: there is no way to tell why authorization failed. One use case where this is needed is determining what status code to send. 401 is sent if no `Authorization` header is present and 403 is sent in case of (say) an invalid/expired token.

### Proposal

Replace the `Option` with a `Result`. That way, we can inform `{Async}AuthorizeRequest::unauthorized_response` why the request could not be authorized so it can respond appropriately.

This does force the user to handle the error case but that is inline with how Rust handles errors.

### Alternatives

- Use request extensions to figure out what went wrong
  This is janky and error prone. Using `Result` is a much better way.

davidpdrsn:
Fixes #190

This changes the `AuthorizeRequest` and `AsyncAuthorizeRequest` traits to be quite a bit simpler I think.

## Sync

Before: https://github.com/tower-rs/tower-http/blob/master/tower-http/src/auth/async_require_authorization.rs#L234-L269

After

```rust
pub trait AuthorizeRequest<B> {
    type ResponseBody;

    fn authorize(
        &mut self,
        request: &mut Request<B>,
    ) -> Result<(), Response<Self::ResponseBody>>;
}
```

- `authorize` now returns `Result<(), Response>`
- `Ok(())` means the request was accepted and will be passed on the inner service
- `Err(response)` means the request couldn't be authorized and the response will be returned without calling the inner service.

## Async

Before: https://github.com/tower-rs/tower-http/blob/master/tower-http/src/auth/async_require_authorization.rs#L234-L269

After

```rust
pub trait AsyncAuthorizeRequest<B> {
    type RequestBody;
    type ResponseBody: Body;
    type Future: Future<Output = Result<Request<Self::RequestBody>, Response<Self::ResponseBody>>>;

    fn authorize(&mut self, request: Request<B>) -> Self::Future;
}
```

- `authorize` now receives an owned request. This allows you to move the request into the future, do some async stuff which yields some data (such as user id from a database), and then save that in a request extension.
- This wouldn't be possible if the future received a `&mut Request` because it would require GATs.
- Therefore the future must resolve to `Result<Request, Response>` so the request can be passed on the inner service.
- Authorization is allowed to change the request body type. Might be useful for someone to buffer and parse the request body to consider it for auth. Then one could change the request body from a generic `B` to [`http_body::Full`](https://docs.rs/http-body/0.4.4/http_body/struct.Full.html).
- This is essentially the same as [`AsyncPredicate`](https://docs.rs/tower/0.4.11/tower/filter/trait.AsyncPredicate.html) from tower.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   be used instead ([#170]) (BREAKING)
 - **fs**: Changed response body type of `ServeDir` and `ServeFile` to
   `ServeFileSystemResponseBody` and `ServeFileSystemResponseFuture` ([#187]) (BREAKING)
+- **auth**: Change `AuthorizeRequest` and `AsyncAuthorizeRequest` traits to be simpler ([#???]) (BREAKING)
 
 ## Removed
 
--- a/tower-http/src/auth/async_require_authorization.rs
+++ b/tower-http/src/auth/async_require_authorization.rs
@@ -14,30 +14,36 @@
 //! #[derive(Clone, Copy)]
 //! struct MyAuth;
 //!
-//! impl AsyncAuthorizeRequest for MyAuth {
-//!     type Output = UserId;
-//!     type Future = BoxFuture<'static, Option<UserId>>;
+//! impl<B> AsyncAuthorizeRequest<B> for MyAuth
+//! where
+//!     B: Send + Sync + 'static,
+//! {
 //!     type ResponseBody = Body;
+//!     type Future = BoxFuture<'static, Result<Request<B>, Response<Self::ResponseBody>>>;
 //!
-//!     fn authorize<B>(&mut self, request: &Request<B>) -> Self::Future {
+//!     fn authorize(&mut self, mut request: Request<B>) -> Self::Future {
 //!         Box::pin(async {
-//!             // ...
-//!             # None
+//!             if let Some(user_id) = check_auth(&request).await {
+//!                 // Set `user_id` as a request extension so it can be accessed by other
+//!                 // services down the stack.
+//!                 request.extensions_mut().insert(user_id);
+//!
+//!                 Ok(request)
+//!             } else {
+//!                 let unauthorized_response = Response::builder()
+//!                     .status(StatusCode::UNAUTHORIZED)
+//!                     .body(Body::empty())
+//!                     .unwrap();
+//!
+//!                 Err(unauthorized_response)
+//!             }
 //!         })
 //!     }
+//! }
 //!
-//!     fn on_authorized<B>(&mut self, request: &mut Request<B>, user_id: UserId) {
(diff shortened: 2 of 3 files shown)


davidpdrsn:
@hamza1311 what do you think?

hamza1311:
API looks good. Better than the simple swapping of Option with Result that I suggested

davidpdrsn:
`AuthorizeRequest` and `AsyncAuthorizeRequest` are now also implemented for closures which is quite ergonomic 😊

tower-http/src/auth/async_require_authorization.rs
```diff
@@ -232,40 +263,36 @@ where
...
-    type Future: Future<Output = Option<Self::Output>>;
+    /// Set this to `B` if you need to change the request body type.
+    type RequestBody;
 
     /// The body type used for responses to unauthorized requests.
-    type ResponseBody: Body;
+    type ResponseBody;
+
+    /// The Future type returned by `authorize`
+    type Future: Future<Output = Result<Request<Self::RequestBody>, Response<Self::ResponseBody>>>;
```
Nehliin, line 276: This is really neat!

tower-http/src/auth/async_require_authorization.rs
```diff
@@ -232,40 +263,36 @@ where
...
-    /// The output type of doing the authorization.
+pub trait AsyncAuthorizeRequest<B> {
+    /// The type of request body returned by `authorize`.
     ///
-    /// Use `()` if authorization doesn't produce any meaningful output.
-    type Output;
-
-    /// The Future type returned by `authorize`
-    type Future: Future<Output = Option<Self::Output>>;
+    /// Set this to `B` if you need to change the request body type.
```
Nehliin, line 269: Great to have this as part of the docs :+1:

Nehliin (approved):
Looks a lot cleaner to me!

tower-http/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   be used instead ([#170]) (BREAKING)
 - **fs**: Changed response body type of `ServeDir` and `ServeFile` to
   `ServeFileSystemResponseBody` and `ServeFileSystemResponseFuture` ([#187]) (BREAKING)
+- **auth**: Change `AuthorizeRequest` and `AsyncAuthorizeRequest` traits to be simpler ([#???]) (BREAKING)
```
Nehliin, line 42: ah forgot to mention this, that we can put in the proper pr number here now

davidpdrsn: Oh yeah good catch!

PR #196, opened 2021-12-23
Update minimum dependency versions

kylewlacy:
This PR updates the minimum dependency versions in `Cargo.toml` to prevent compilation failures.

## Motivation

Originally, after trying to update our project to tower-http 0.2.0, we got a compilation error from tower-http itself after updating.

I did some sleuthing and found that the version of `futures-util` in our lockfile was too old for tower-http. Doing `cargo update -p futures-util` was good enough to solve the immediate problem, but it seems like the kind of thing that should be upstreamed!

## Solution

I updated all dependencies listed in `Cargo.toml` to the minimum versions needed to still compile. I basically used the following to verify that I didn't miss any dependencies:

```
$ git clone 'https://github.com/kylewlacy/tower-http.git'
$ cd tower-http
$ git checkout origin/update-minimum-dependency-versions
$ cargo +nightly check -Z minimal-versions --all-features --lib
(Initial compilation will fail due to an upstream error in crc32)
$ cargo update -p crc32
$ cargo +nightly check -Z minimal-versions --all-features --lib
(This succeeds with the changes from this PR, but would fail on master)
```

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -13,11 +13,11 @@ keywords = ["io", "async", "futures", "service", "http"]
 rust-version = "1.51"
 
 [dependencies]
-bitflags = "1.0"
+bitflags = "1.3.1"
 bytes = "1"
 futures-core = "0.3"
-futures-util = { version = "0.3", default_features = false, features = [] }
-http = "0.2"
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks!

PR #199, opened 2022-01-01
CORS middleware should set the Vary header

### LINKED ISSUE #198: CORS middleware should set the `Vary` header
From reading https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin it sounds like we should be setting `Vary` in cors responses. Probably also for other headers than `Access-Control-Allow-Origin`.

shepmaster:
The headers are based on [corsware][] which states:

> These are all headers which can influence the outcome of any given
> CORS request.

Fixes #198

[corsware]: https://github.com/atorstling/corsware/blob/3467805a74145b788f8234f9f994caf045509b31/src/lib.rs#L223-L229

### DIFF
--- a/tower-http/src/cors.rs
+++ b/tower-http/src/cors.rs
@@ -81,6 +81,12 @@ pub struct CorsLayer {
 
 const WILDCARD: &str = "*";
 
+const VARY_HEADERS: [HeaderName; 3] = [
+    header::ORIGIN,
+    header::ACCESS_CONTROL_REQUEST_METHOD,
+    header::ACCESS_CONTROL_REQUEST_HEADERS,
+];
+
 impl CorsLayer {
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
Thanks! This looks good to me.

It also matches [what actix-web does](https://github.com/actix/actix-extras/blob/master/actix-cors/src/middleware.rs#L118-L134) (as of two days ago 😅)

I'll fix CI in a separate PR.

jplatte:
Changelog entry is missing, followup PR for that would probably be good?

davidpdrsn:
Yeah I noticed that. I'll just fill it in when doing a release.

davidpdrsn:
Actually I just did it in c99dd0a42b525

PR #200, opened 2022-01-04
fix link from readme to license

dimbleby:
## Motivation

The LICENSE file for this project was lost in the great reset of 9ca9e7cc56a43.

## Solution

The README still says MIT license and links to the missing file, so I am taking the liberty of sending you an MR that reinstates that file.

It's exactly as it previously was, except that I've added 2022 to the copyright range.


davidpdrsn:
Is it not enough to have the license here https://github.com/tower-rs/tower-http/blob/master/tower-http/LICENSE?

dimbleby:
Aha, that very much looks like enough!  Give me a moment and I'll repurpose this to fix the link from the README instead...

davidpdrsn (approved):

Issue #194, opened 2021-12-16
Preflight response access control headers should match request


tqwewe:
### Version

`tower-http v0.2.0`

### Platform

`Darwin MacBook-Pro-2 21.1.0 Darwin Kernel Version 21.1.0`

### Description

Preflight response should match request `access-control-allow-origin` and `access-control-allow-headers` and should not use wild card `*`.

Currently, cors issues occur due to preflight response headers not matching the requested headers for origin and allowed headers. The response should copy the headers from the request (if allowed in the configuration).

<https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#credentialed_requests_and_wildcards>

> - The server **must not** specify the "`*`" wildcard for the `Access-Control-Allow-Origin` response-header value, but must instead specify an explicit origin; for example: `Access-Control-Allow-Origin: https://example.com`
> - The server **must not** specify the "`*`" wildcard for the `Access-Control-Allow-Headers` response-header value, but must instead specify an explicit list of header names; for example, `Access-Control-Allow-Headers: X-PINGOTHER, Content-Type`
> - The server must not specify the "`*`" wildcard for the `Access-Control-Allow-Methods` response-header value, but must instead specify an explicit list of method names; for example, `Access-Control-Allow-Methods: POST, GET`

The above rules don't seem to be handled correctly by tower-http.

jplatte:
I agree that there is a problem here, but I don't think the suggested solution makes sense.

IMO `CorsLayer::permissive` should simply not include `allow_credentials(true)` and we should add a note to the docs that you can't use wildcards together with `allow_credentials(true)`. Attempting to bypass the rules documented at the provided link by having the server respond with different CORS headers based on the request headers will inevitably lead to security vulnerabilities. Additionally, I'm pretty sure preflight requests don't include all of the headers of the main request, so for `allow_headers` this wouldn't work anyways.

davidpdrsn:
> A future version could potentially even panic when attempting to use a CORS middleware that sets a wildcard in combination with allow_credentials(true) since that is apparently never correct.

iirc actix’s cors implementation does that. Worth investigating for sure.

tqwewe:
@jplatte
> I'm pretty sure preflight requests don't include all of the headers of the main request, so for allow_headers this wouldn't work anyways.

I believe they do under `access-control-request-headers`, but I could be wrong. But this is what fixed it for me.

[...]

PR #203, opened 2022-01-13
Set acess-control-allow-origin to request origin when using Any (#194)

tilpner:
## Motivation

Cross-site requests with credentials are currently prevented by browsers when CorsLayer is set to allow any origin, because it replies with a literal '*' instead of the requests origin.

Example error:
> The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.

## Solution

This *fixes* #194, by replying with the request's Origin when the CorsLayer is set to allow Any origin.
It does that by removing existing special-casing of the Any origin.

cc @jplatte, who was arguing against this solution

### DIFF
--- a/tower-http/src/cors.rs
+++ b/tower-http/src/cors.rs
@@ -769,7 +769,7 @@ where
         match self.project().inner.project() {
             KindProj::CorsCall {
                 future,
-                allow_origin,
+                allow_origin: _,
                 origin,
                 allow_credentials,
                 expose_headers,
@@ -779,7 +779,7 @@ where
 
                 headers.insert(
                     header::ACCESS_CONTROL_ALLOW_ORIGIN,
-                    response_origin(allow_origin.take().unwrap(), origin),
+                    origin.clone()
                 );
 
                 if let Some(allow_credentials) = allow_credentials {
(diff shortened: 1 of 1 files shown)


jplatte:
In case of the origin specifically, this could already be done in user code by using a closure that always returns `true`. I think that would be better than removing the obvious mapping from `Any` to `'*'`.

There's no equivalent to that for headers though, but I guess you don't care about that since this PR doesn't do anything with allow-headers?

jplatte:
@davidpdrsn Wdyt about this: Instead of `AnyOr<_>`, how about representing all of [`allow_headers`, `allow_origin`, `allow_methods`, `expose_headers`] as their own type similar to the `Origin` type, but have those types simply hold a boxed `FnMut` that can access the request? Then those types can have constructors like `Origin::any()` and `Origin::from_list(vec![...])` and `Origin::from_request_origin()`. That makes it easy to do the thing that some people want and allow anything including headers, which could also be exposed as another `Cors` constructor (though I think the current `Cors::permissive()` should still exist in some form, just without the ineffective `allow_credentials(true)`).

tilpner:
While a closure does solve my specific usecase, it's not really a good API, with `Any` and `|_, _| true` meaning different variants of "allow everything" (in addition to the low discoverability). And it might make sense to preserve the ability of returning literal asterisks instead of the request origin, although I can't come up with a usecase on the spot. Overall, your proposal of eliminating AnyOr seems preferable to this, so I'll close the PR for now.

Issue #201, opened 2022-01-06
Building with fs and only compression-gzip doesnt work


steffengy:
## Bug Report

Building with `tower-http = { version = "0.2", features = ["fs", "compression-gzip"] }`

### Version
tower-http v0.2.0 (*)

### Platform

Windows x64

### Description

```
error[E0004]: non-exhaustive patterns: `(true, Deflate)` and `(true, Brotli)` not covered
  --> C:\Users\steff\.cargo\registry\src\github.com-1ecc6299db9ec823\tower-http-0.2.0\src\compression\future.rs:48:26
   |
48 |         let body = match (should_compress, self.encoding) {
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ patterns `(true, Deflate)` and `(true, Brotli)` not covered
   |
   = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms
   = note: the matched value is of type `(bool, Encoding)`
```

PR #205, opened 2022-01-14
Clean up some #[cfg]s in compression/body.rs

Nehliin:
## Motivation
Was originally looking into https://github.com/tower-rs/tower-http/issues/201 but kind of rabbit holed and felt a strong urge to clean up some of the `#[cfg]` we use in compression body.

## Solution
The `pin_project_lite` macro (looks like it's the same for the regular `#[pin_project]` macro) doesn't play well with `#[cfg]` attributes which led to some messy workarounds that I wanted to get rid of. I found this comment https://github.com/taiki-e/pin-project-lite/issues/3#issuecomment-774461161 by @tesaguri which provides a macro workaround which solves the problem and makes things a bit cleaner. It's licensed under CC0 1.0 which means it should be fine to use afaik.  It's still a bit messy but more contained to the macro itself at least.

### DIFF
--- a/tower-http/src/compression/body.rs
+++ b/tower-http/src/compression/body.rs
@@ -23,6 +23,8 @@ use std::{
 };
 use tokio_util::io::StreamReader;
 
+use super::pin_project_cfg::pin_project_cfg;
+
 pin_project! {
     /// Response body of [`Compression`].
     ///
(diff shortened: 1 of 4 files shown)


davidpdrsn (approved):
Nice!

davidpdrsn:
CI should be fixed by https://github.com/tower-rs/tower-http/pull/207

PR #208, opened 2022-01-18
predicate: Fix comment for images and SizeAbove

kevinji:
The original comment referred to when compression *would* happen, not when it would not.

## Motivation

A comment documented the behavior incorrectly.

## Solution

Fixed the comment.

### DIFF
--- a/tower-http/src/compression/predicate.rs
+++ b/tower-http/src/compression/predicate.rs
@@ -89,8 +89,8 @@ where
 /// This will compress responses unless:
 ///
 /// - They're gRPC, which has its own protocol specific compression scheme.
-/// - Its not an image as determined by the `content-type` starting with `image/`.
-/// - The response is larger than 32 bytes.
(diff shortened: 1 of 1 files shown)


tower-http/src/compression/predicate.rs
```diff
@@ -89,8 +89,8 @@ where
 /// This will compress responses unless:
 ///
 /// - They're gRPC, which has its own protocol specific compression scheme.
-/// - Its not an image as determined by the `content-type` starting with `image/`.
-/// - The response is larger than 32 bytes.
+/// - It's not an image as determined by the `content-type` starting with `image/`.
```
Nehliin, line 92: The `not` should be removed since we _do not_ compress images.

kevinji:
Addressed the comment and updated the commit message.

davidpdrsn (approved):

Nehliin (approved):

Issue #212, opened 2022-01-22
Please undo yanking of version 0.1


TethysSvensson:
## Bug Report

Please undo the yanking of version 0.1.2, since it's breaking the ecosystem. It's currently impossible to install e.g. trunk using cargo:

```
$ cargo install trunk
    Updating crates.io index
  Downloaded trunk v0.14.0
  Downloaded 1 crate (123.7 KB) in 0.40s
  Installing trunk v0.14.0
error: failed to compile `trunk v0.14.0`, intermediate artifacts can be found at `/tmp/cargo-installQy7ES2`

Caused by:
  failed to select a version for the requirement `tower-http = "^0.1"`
  candidate versions found which didn't match: 0.2.1, 0.0.0
  location searched: crates.io index
  required by package `trunk v0.14.0`
```

### Version

0.1.2

davidpdrsn:
It was yanked due to a security vulnerability: https://github.com/rustsec/advisory-db/blob/main/crates/tower-http/RUSTSEC-2021-0135.md

So I would rather that trunk updates to a version that doesn't have this vulnerability, that is 0.2.1

davidpdrsn:
I've made https://github.com/tower-rs/tower-http/pull/213 which fixes the bug in 0.1.x, when that is approved I'll release 0.1.3 which should make trunk installable again.

davidpdrsn:
0.1.3 is published and trunk is installable again:

```
❯ cargo install trunk
    Updating crates.io index
  Downloaded trunk v0.14.0
  Downloaded 1 crate (123.7 KB) in 0.49s
  Installing trunk v0.14.0
  Downloaded anyhow v1.0.53
  Downloaded tokio-rustls v0.23.2
  Downloaded structopt v0.3.26
  Downloaded socket2 v0.4.3
  Downloaded webpki-roots v0.22.2
  Downloaded serde v1.0.134
  Downloaded serde_derive v1.0.134
  Downloaded crc32fast v1.3.1
  Downloaded libc v0.2.113
  Downloaded tower-http v0.1.3 <-- the new version
  Downloaded 10 crates (1.2 MB) in 1.47s
   Compiling libc v0.2.113
   // ...
   Compiling trunk v0.14.0
    Finished release [optimized] target(s) in 2m 26s
  Installing /Users/davidpdrsn/.cargo/bin/trunk
   Installed package `trunk v0.14.0` (executable `trunk`)
```

@TethysSvensson Thanks for reporting this!

[...]

PR #213, opened 2022-01-22
Backport #204 to 0.1.x

davidpdrsn:
This backports the fix in https://github.com/tower-rs/tower-http/issues/212 such that we can release 0.1.3 which should fix https://github.com/tower-rs/tower-http/issues/212.


davidpdrsn:
> I think it could also make sense to add a test + windows machine i ci if possible to ensure it doesn't regress and also doesn't happen on other platforms

Yes thats a good idea! Can do that in a follow up

PR #214, opened 2022-01-26
Add `CatchPanic`

davidpdrsn:
This adds a `CatchPanic` middleware that catches panics and converts them into `500` responses.

## TODO

- [x] Module docs with examples
- [x] Support supplying closure to generate response in case users wanna return a fancy JSON response or something like that.

### DIFF
--- a/tower-http/src/catch_panic.rs
+++ b/tower-http/src/catch_panic.rs
@@ -3,12 +3,11 @@
 use bytes::Bytes;
 use futures_core::ready;
 use futures_util::future::{CatchUnwind, FutureExt};
-use http::{Request, Response, StatusCode};
+use http::{HeaderValue, Request, Response, StatusCode};
 use http_body::{combinators::UnsyncBoxBody, Body, Full};
 use pin_project_lite::pin_project;
 use std::{
     any::Any,
-    borrow::Cow,
     future::Future,
     panic::AssertUnwindSafe,
     pin::Pin,
@@ -17,26 +16,47 @@ use std::{
 use tower_layer::Layer;
 use tower_service::Service;
 
+use crate::BoxError;
+
 /// Layer that applies the [`CatchPanic`] middleware that catches panics and converts them into
 /// `500 Internal Server` responses.
 ///
 /// See the [module docs](self) for an example.
 #[derive(Debug, Clone, Copy, Default)]
-#[non_exhaustive]
-pub struct CatchPanicLayer;
+pub struct CatchPanicLayer<T> {
+    panic_handler: T,
+}
 
-impl CatchPanicLayer {
-    /// Create a new `CatchPanicLayer`.
+impl CatchPanicLayer<DefaultResponseForPanic> {
+    /// Create a new `CatchPanicLayer` with the default panic handler.
     pub fn new() -> Self {
-        CatchPanicLayer {}
+        CatchPanicLayer {
+            panic_handler: DefaultResponseForPanic,
+        }
     }
 }
 
-impl<S> Layer<S> for CatchPanicLayer {
-    type Service = CatchPanic<S>;
+impl<T> CatchPanicLayer<T> {
+    /// Create a new `CatchPanicLayer` with a custom panic handler.
+    pub fn custom(panic_handler: T) -> Self
+    where
+        T: ResponseForPanic,
+    {
+        Self { panic_handler }
+    }
+}
+
+impl<T, S> Layer<S> for CatchPanicLayer<T>
+where
+    T: Clone,
+{
+    type Service = CatchPanic<S, T>;
 
     fn layer(&self, inner: S) -> Self::Service {
-        CatchPanic::new(inner)
+        CatchPanic {
+            inner,
+            panic_handler: self.panic_handler.clone(),
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,202 @@
...
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        match std::panic::catch_unwind(AssertUnwindSafe(|| self.inner.call(req))) {
+            Ok(future) => ResponseFuture {
+                kind: Kind::Future {
+                    future: AssertUnwindSafe(future).catch_unwind(),
```
davidpdrsn, line 85: Requiring the service and future to be `UnwindSafe` doesn't work. It doesn't compile when you actually try and use it.

I'm honestly not very familiar with `AssertUnwindSafe` but I suppose this is fine since there is no `unsafe` code 🤔 

@jplatte maybe you know?

tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,202 @@
...
+            },
+        }
+    }
+}
+
+fn panic_err_to_response<E>(
+    err: Box<dyn Any + Send + 'static>,
+) -> Response<UnsyncBoxBody<Bytes, E>> {
+    let message: Cow<_> = if let Some(s) = err.downcast_ref::<String>() {
+        s.to_owned().into()
```
kosta, line 144: Microoptimization: Since we're owning and thowing away the `Box` anyway, we might do a `downcast_mut()` and `swap()`ing with an empty `String`? (saves the `to_owned()`)

tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,202 @@
...
+        let body = hyper::body::to_bytes(res).await.unwrap();
+        assert_eq!(&body[..], b"Service panicked: service panic");
+    }
+
+    #[tokio::test]
+    async fn panic_in_future() {
+        let svc = ServiceBuilder::new()
+            .layer(CatchPanicLayer::new())
+            .service_fn(|_: Request<Body>| async {
+                panic!("future panic");
```
kosta, line 190: One of the tests should panic with a format string, so that both panic payload cases (`&str` and `String`) are tested.

tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,202 @@
...
+    let message: Cow<_> = if let Some(s) = err.downcast_ref::<String>() {
+        s.to_owned().into()
+    } else if let Some(s) = err.downcast_ref::<&str>() {
+        (*s).into()
+    } else {
+        "`CatchPanic` middleware was unable to obtain panic info".into()
+    };
+    let body = format!("Service panicked: {}", message);
+
+    let mut res = Response::new(Full::from(body).map_err(|err| match err {}).boxed_unsync());
```
kosta, line 152: I dont think we should hand out the panic message by default to the user. Instead, we should probably `error!` log the exception and just return `500 Internal Server Error` with an empty response - but leave the exact behaviour up to customization as you suggested.

Rationale: I would think it's a bit unprofessional if I GET from some SaaS service and and get back a 500 with the body ~~`undefined is not a function`~~`index out of bounds: the len is 0 but the index is 1`. Also, the current approach does not leave any trace of the panic in the logs.

kosta (changes requested):
Wow, that was quick :) Thank you!

tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,202 @@
...
+    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
+    ResBody: Body<Data = Bytes> + Send + 'static,
+{
+    type Response = Response<UnsyncBoxBody<Bytes, ResBody::Error>>;
+    type Error = S::Error;
+    type Future = ResponseFuture<S::Future>;
+
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
```
kosta, line 78: Not sure, but should this also catch panics? 🤔

tower-http/src/catch_panic.rs
kosta: I had the same experience, but it seems this approach is correct:

* see https://doc.rust-lang.org/stable/src/core/panic/unwind_safe.rs.html#170
* [`catch_unwind` from `futures` basically does the same thing](https://docs.rs/futures-util/0.3.19/src/futures_util/future/future/catch_unwind.rs.html#36)

jplatte: Yeah `AssertUnwindSafe` isn't exactly intuitive. There's even a Pre-RFC about deprecating it where you can read more about what it does, how it's useful and how it's not so useful in other cases: https://internals.rust-lang.org/t/pre-rfc-deprecating-unwindsafe/15974

hawkw:
Hmm, I kind of wonder if it makes sense to have a general-purpose `CatchPanic` middleware in `tower` itself, and then a middleware in `tower-http` that turns panic errors into 500s? not a blocker.

tower-http/src/catch_panic.rs
hawkw: yeah, I think it might be desirable to have some way to configure the middleware with whether or not it actually exposes the panic message in the response. you might want to turn it on for local development, but disable it in production, both for making the app look nice, and also potentially for security reasons (you maybe don't want to expose a panic message or backtrace to the entire internet...)

davidpdrsn:
@hawkw yep that's definitely something I want to explore as well before merging this!

I would like tower-http to not have a public dependency on tower (just service and layer) but I'll see what I can come up with 😊

tower-http/src/catch_panic.rs
davidpdrsn: I've added support for custom panic handlers now. I made the default just generic an opaque "Service panicked" message and `tracing::error!` the full message. I believe this will work fine for must cases without having too much customization.

tower-http/src/catch_panic.rs
davidpdrsn: Hm so thats a good question actually. I'm actually thinking no, panics from `poll_ready` shouldn't be handled for a few reasons:

- We cannot return a response from `poll_ready` so we'd have to return an error. That would require changing the error type to `BoxError` which means a service that previously returned `Infallible` would now return `BoxError`. [This makes it more annoying to use with axum.](https://docs.rs/axum/latest/axum/error_handling/index.html)
- If `poll_ready` returns an error the `Service` contract dictates that the service must be discarded and recreated. Hyper handles that and does the same thing regardless if `poll_ready` errors or panics. So since we cannot return a response it _kinda_ doesn't matter what we do 🤷 
- In practice its rare for services to actually care about `poll_ready` so panics from here should be very rare. I would guess they're much more common from `call`.

@hawkw What do you think?

davidpdrsn:
> Hmm, I kind of wonder if it makes sense to have a general-purpose `CatchPanic` middleware in `tower` itself, and then a middleware in `tower-http` that turns panic errors into 500s? not a blocker.

I've opted not to add anything to tower since I guess such a middleware would convert panics to errors. So to make this middleware easier to use with axum I'd prefer keeping the error type and instead changing the response.

@hawkw let me know if you had something else in mind!

davidpdrsn:
Note to self: Read https://github.com/actix/actix-web/issues/1501

tower-http/src/catch_panic.rs
Nehliin: Even thought there might not be soundness issues it looks to me like it could cause potential state corruption in certain cases. I am certainly not familiar enough with catch_unwind to see if this would have the kind of problem but this feels a bit scary

tower-http/src/catch_panic.rs
LucioFranco: Yeah, I don't think we should catch panics here. Those reasons track for me.

tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+impl<S, T, ReqBody, ResBody> Service<Request<ReqBody>> for CatchPanic<S, T>
+where
+    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
+    ResBody: Body<Data = Bytes> + Send + 'static,
+    ResBody::Error: Into<BoxError>,
+    T: ResponseForPanic + Clone,
+    T::ResponseBody: Body<Data = Bytes> + Send + 'static,
+    <T::ResponseBody as Body>::Error: Into<BoxError>,
```
LucioFranco, line 186: Do we actually need to include these send bounds?

LucioFranco (approved):
LGTM jiust a few questions not blocking

tower-http/src/catch_panic.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+}
+
+/// The default `ResponseForPanic` used by `CatchPanic`.
+///
+/// It will log the panic message and return a `500 Internal Server` error response with an empty
+/// body.
+#[derive(Debug, Default, Clone, Copy)]
+#[non_exhaustive]
+pub struct DefaultResponseForPanic;
```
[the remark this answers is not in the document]
davidpdrsn: Yep 😊

tower-http/src/catch_panic.rs
davidpdrsn: Yes. They're required because we call `http_body::Body::boxed_unsync` which requires the body to be `Send`. We call that on the body from inner service and from the callback that creates the panic response.

PR #216, opened 2022-02-06
Add Vary headers for CORS preflight responses

shepmaster:
## Motivation

While migrating from Corsware to Axum, my tests detected that Corsware [applied the `Vary` header to both the preflight response and actual response](https://github.com/atorstling/corsware/blob/3467805a74145b788f8234f9f994caf045509b31/src/lib.rs#L248).

## Solution

This does the same for tower-http.

### DIFF
--- a/tower-http/src/cors.rs
+++ b/tower-http/src/cors.rs
@@ -712,7 +712,7 @@ where
             }
 
             return ResponseFuture {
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #221, opened 2022-02-14
tokio-util: 0.6.8 -> 0.7.0

teozkr:
This mirrors https://github.com/tower-rs/tower/pull/638.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -37,10 +37,6 @@ license-files = [
 multiple-versions = "deny"
 highlight = "all"
 skip-tree = [{ name = "tower", version = ">=0.3, <=0.4" }]
-skip = [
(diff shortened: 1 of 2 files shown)


davidpdrsn (approved):
Thanks!

teozkr (commented):
There still seem to be some `cargo deny` failures, but those are unrelated as far as I can tell (disagreement between `headers` and `tungstenite` on the version of `sha-1`).

davidpdrsn:
> There still seem to be some `cargo deny` failures, but those are unrelated as far as I can tell (disagreement between `headers` and `tungstenite` on the version of `sha-1`).

Thats fine. I'll resolve those later :)

PR #225, opened 2022-02-27
Bump iri-string to 0.5.0 and use shorter type names

lo48576:
This PR bumps the internal dependency to `iri-string` crate to the latest stable version.

(Disclaimer: I am the author of `iri-string` crate.)

## Motivation

* The new version performs a little better.
    + URI resolution is made roughly 3x faster (by quite simple and naive benchmark).
* The new version has no mandatory dependencies.
    + Dependencies to `nom` and its indirect deps are removed since v0.5.0.
    + `axum` still depends on `nom`, but `tokio-http` no longer does.
* Updating the dependency is happy :-).
    + The crate and its items are completely internal, so updating won't cause compatibility issues.

## Solution

Just bump the `iri-string` dependency.

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -27,7 +27,7 @@ tower-service = "0.3"
 async-compression = { version = "0.3", optional = true, features = ["tokio"] }
 base64 = { version = "0.13", optional = true }
 http-range-header = "0.3.0"
-iri-string = { version = "0.4", optional = true }
+iri-string = { version = "0.5.0", optional = true }
 mime = { version = "0.3", optional = true, default_features = false }
 mime_guess = { version = "2", optional = true, default_features = false }
 percent-encoding = { version = "2.1.0", optional = true }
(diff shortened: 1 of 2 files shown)


lo48576:
Sorry, I overlooked that MSRV of this project is 1.51...

davidpdrsn:
@lo48576 what is iri-string's MSRV? As long as we don't go above 1.54 which is axum's MSRV then its probably fine to bump it. We could even bump the MSRV for axum if necessary, as long as its not to the most recent version.

lo48576:
iri-string's MSRV is 1.58.0 (released at 2022-01-13).
I can try to lower iri-string's MSRV, but I'm not sure it is worth doing.

lo48576:
iri-string 0.5.0 uses `Vec::try_reserve` which is introduced since 1.57.0, so MSRV won't go below that...

tesaguri:
`tower-http`'s MSRV is now 1.60 (#306) so I believe this pull request can be reopened.

lo48576:
Thank you for notifying me. I created #312 for bumping `iri-string` to the newer version.

Issue #197, opened 2021-12-31
Partial Content not working on Firefox and Chrome


adumbidiot:
### Version
`tower-http v0.2.0`

### Platform
`Windows 10 64 bit`

### Description
I was trying to stream a video using the `ServeDir` service, but both Chrome and Firefox were unable to view the video. I'm fairly certain that the `Content-Length` header is being set incorrectly. The following is a sample of some of the response headers I got in Firefox while testing:
![image](https://user-images.githubusercontent.com/26289915/147805167-ea328c2a-8575-49c0-9ea3-84c923693d06.png)

I believe that the content-length header should reflect the message length, not the file length as per the [documentation on mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length). I also made a branch that fixes this issue and seems to allow me to stream the video without issue:  https://github.com/adumbidiot/tower-http/tree/partial-content-quickfix

davidpdrsn:
Thanks for the report! I don't think this should be hard to fix so people are welcome to submit PRs 😊 A test for this would also be good.

adumbidiot:
Fixed with #228

PR #228, opened 2022-03-08
fix content-length calculation on range requests

moumar:
hello,

content-length response header on range requests should represent the length of the returned range, not the length of the whole file.

more explanations here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

Best regards.

### DIFF
--- a/tower-http/src/services/fs/serve_dir.rs
+++ b/tower-http/src/services/fs/serve_dir.rs
@@ -512,8 +512,7 @@ impl Future for ResponseFuture {
                         };
                         let mut builder = Response::builder()
                             .header(header::CONTENT_TYPE, file_request.mime_header_value)
-                            .header(header::ACCEPT_RANGES, "bytes")
-                            .header(header::CONTENT_LENGTH, size.to_string());
+                            .header(header::ACCEPT_RANGES, "bytes");
                         if let Some(encoding) = file_request.maybe_encoding {
                             builder = builder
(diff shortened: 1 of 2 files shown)


davidpdrsn (changes requested):
Oh good catch! I wonder how we missed this 🤔 

Some tests for this would be great!

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -1083,7 +1085,7 @@ mod tests {
         assert_eq!(res.status(), StatusCode::PARTIAL_CONTENT);
         assert_eq!(
             res.headers()["content-length"],
-            file_contents.len().to_string()
+            (bytes_end_incl - bytes_start_incl + 1).to_string()
```
davidpdrsn, line 1088: Ah my bad. I see this is in a test which you updated. Nvm

MarcusGrass:
I see there are a few tests running on both head-requests and range-requests checking the content-length and that you fixed the test that should return something other than the full file size. It seems that I both wrote a bug and a bugged test to go with it. Really nice catch, thank you!

adumbidiot:
I think this closes #197. I’ll try to test it out later

davidpdrsn (approved):
I think this looks good. Will wait for @adumbidiot to report back before merging.

adumbidiot:
Just tested, this fixes #197

davidpdrsn:
Great! I think I'll backport this to 0.2.x tomorrow while we work on 0.3.

PR #231, opened 2022-03-15
feat(fs): make AsyncReadBody::with_capacity() pub

npmccallum:
The `AsyncReadBody` type is a generally useful type. It is already public, but it has no public constructor. This commit makes the main constructor public so that the type can be used by other projects. If the type is ever upstreamed to `http-body`, it can be reexported here to maintain compatibility.


davidpdrsn:
Sorry but I don't think we should do this. tower-http doesn't currently expose any bodies and I'm not sure we should start going down that route. I'd rather we work directly towards getting it into http-body. Perhaps as part of the hyper 1.0 effort 🤔

The implementation is also so simple that I'd honestly recommend users copy/paste the code if they need it.

PR #234, opened 2022-03-17
Add `Spa` for serving files for single page apps

davidpdrsn:
I'll write a larger description when I've written some tests.

The idea is to use this with axum like so

```rust
Router::new()
    .route("/foo", get(|| async { "GET /foo" }))
    .fallback(
        get_service(SpaService::new("./dist"))
            .handle_error(|_| async { StatusCode::INTERNAL_SERVER_ERROR }),
    );
```

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 
-- None.
+- Added `services::Spa` for serving files for single page apps.
 
 ## Changed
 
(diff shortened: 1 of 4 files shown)


davidpdrsn:
Hm actually been thinking that we can make can make something nicer in axum, that doesn't require serving assets at the root which is odd I think.

jplatte:
I kinda liked this idea, but it wasn't clear to me why it would be a separate service rather than a setting on `ServeDir` where users can configure a fallback in case no file is found.

davidpdrsn:
I was thinking that a helper giving you this setup in axum might be nicer:

```rust
Router::new()
    .nest(
        "/assets",
        get_service(ServeDir::new("dist")).handle_error(internal_error),
    )
    .fallback(any_service(ServeFile::new("dist/index.html")).handle_error(internal_error));
```

Issue #240, opened 2022-03-26
Add special support for single-page applications which does its own routing


udoprog:
### Motivation

It's fairly common for single page applications to require that all routes which do not match a known file be routed to a single entrypoint (i.e. `index.html`) and then allow it to perform its own routing. So if we go to `/awesome/path` that would be routed to the same entrypoint because `/awesome/path` is not a file being served.

This is hard to implement correctly using `ServeDir`. You could do something like this:

```rust
ServiceBuilder::new()
        .and_then(
            |response: Response<ServeFileSystemResponseBody>| async move {
                let response = if response.status() == StatusCode::NOT_FOUND {
                    match fs::read(&index_html).await {
                        Ok(bytes) => Response::builder()
                            .status(StatusCode::OK)
                            .body(Full::from(bytes).map_err(|err| match err {}).boxed())?,
                        Err(e) if e.kind() == io::ErrorKind::NotFound => {
                            Response::builder().status(StatusCode::NOT_FOUND).body(Empty::new().map_err(|err| match err {}).boxed())?
                        }
                        Err(e) => return Err(Error::from(e)),
                    }
                } else {
                    response.map(|body| body.boxed())
                };

                Ok::<_, Error>(response)
            },
        )
        .service(
            ServeDir::new(root.join("..").join("web").join("dist"))
                .append_index_html_on_directories(false),
        )
```

This however lacks all the request plumbing which is done by `ServeDir`, such as dealing with accept encoding and minor internal path rewrites.

### Proposal / Alternatives

Introduce some way to make it easier to compose a solution that can both leverage the plumbing of `ServeDir` and allow for falling back in case a file is missing and rewrite the request to a different path.

It's made somewhat complicated, because we won't know if the file is available until we try to read it well into the response future at which point we don't have access to the `ServeDir` instance that constructed it.

Unfortunately I don't have a concrete proposal for this since I'm fairly new to both tower and axum.

davidpdrsn:
In axum-extra we recently added [`SpaRouter`](https://docs.rs/axum-extra/latest/axum_extra/routing/struct.SpaRouter.html) to address this.

I decided to put it in axum, rather than tower-http, because it need more complex routing than whats possible with tower-http (such as [`Router::fallback`](https://docs.rs/axum/latest/axum/routing/struct.Router.html#method.fallback))

@udoprog does this address you use case?

udoprog:
Thanks!

[...]

PR #243, opened 2022-04-04
Add `ServeDir::fallback`

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/240

Example usage:

```rust
let service = ServeDir::new("assets")
    // respond with `index.html` for missing files
    .fallback(ServeFile::new("assets/index.html"));
```

This will call `ServeFile` if the file is missing.

You can also achieve this using combinators from `tower::ServiceExt` but several have asked for this so probably makes sense to add.

Note this is different from [`axum_extra::routing::SpaRouter`](https://docs.rs/axum-extra/latest/axum_extra/routing/struct.SpaRouter.html) which requires assets to be nested (for example under `/assets/*`). This doesn't require that and thus allows serving assets at the top level (i.e. `GET /style.css` or smth like that).

### TODO

- [x] Fix remaining todos in the code
- [x] Make sure its compatible with axum
- [x] Think about setting status codes for fallback responses
- [x] Changelog


jplatte:
This also allows showing a custom 404 page for a `ServeDir`, right? I think that really calls for a convenience API to override `ServeFile`s success status code.

jplatte:
Oops

davidpdrsn:
> This also allows showing a custom 404 page for a `ServeDir`, right?

Yep it does!

> I think that really calls for a convenience API to override `ServeFile`s success status code.

Hm what do you mean? Like not responding with `200 Ok` if the file is found and served?

davidpdrsn:
hehe not quite ready for review yet :P

jplatte:
Misclicked in mobile UI, not sure how to change back 😅

davidpdrsn:
I've fixed it :)

jplatte:
> > I think that really calls for a convenience API to override `ServeFile`s success status code.
> 
> Hm what do you mean? Like not responding with `200 Ok` if the file is found and served?

Yeah, since you'd want the custom not-found page to still have the 404 status code even when served from an on-disk file successfully.

davidpdrsn:
Ahh I see. So like `.fallback(ServeFile::new("not_found.html").success_status(StatusCode::NOT_FOUND))` or something along those lines 🤔 Have to think about this.

davidpdrsn:
Maybe

```rust
ServeDir::new("...").fallback_with_status(ServeFile::new("not_found.html"), StatusCode::NOT_FOUND)
```

That way we wouldn't need anything new in `ServeFile`, or whatever other service you might be using 🤔

jplatte:
> Maybe
> 
> ```rust
> ServeDir::new("...").fallback_with_status(ServeFile::new("not_found.html"), StatusCode::NOT_FOUND)
> ```
> 
> That way we wouldn't need anything new in `ServeFile`, or whatever other service you might be using 🤔 

I guess the question there would be when do you apply the custom status code? Probably if the status code from the inner service is anything 2xx?

Also, if this is not attached to `ServeFile`, it should probably be its own service adapter, and then maybe instead of `.fallback_with_status(inner_svc, status_code)` as an alias for `.fallback(inner_svc.layer(SetStatus::layer(status_code))` we should have `.not_found_service(inner_svc)` as an alias for `.fallback(inner_svc.layer(SetStatus::layer(StatusCode::NOT_FOUND)`? That would provide full flexibility using the adapter, and a very succinct helper method for the common use case.

davidpdrsn:
I'll try and find some time to work on this soon but might just make a breaking release next week to get https://github.com/tower-rs/tower-http/pull/237 out regardless if this has landed or not.

davidpdrsn:
Made a PR to add a `SetStatus` middleware https://github.com/tower-rs/tower-http/pull/248. When that is merged I'll update this to include helpers that also set the status from the fallback to 404. Once thats done I think we can merge this!

I don't think a more general fallback mechanism is possible since it requires cloning the request (to `call` the fallback). `ServeDir` and `ServeFile` doesn't have that issue since they don't require the body or extensions so they're a special case.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -285,26 +318,63 @@ async fn maybe_redirect_or_append_path(
...
+    F: Service<Request<ReqBody>, Response = Response<FResBody>> + Clone,
+    F::Error: Into<io::Error>,
+    F::Future: Send + 'static,
+    FResBody: http_body::Body<Data = Bytes> + Send + 'static,
+    FResBody::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+{
     type Response = Response<ResponseBody>;
     type Error = io::Error;
-    type Future = ResponseFuture;
+    type Future = ResponseFuture<ReqBody, F>;
```
davidpdrsn, line 331: These new type params makes this a breaking change.

davidpdrsn:
This is finally ready for review!

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -120,47 +83,44 @@ enum ServeVariant {
...
+                                .all(|c| matches!(c, Component::Normal(_)))
+                            {
+                                full_path.push(comp)
+                            } else {
+                                return None;
+                            }
+                        }
+                        Component::CurDir => {}
+                        Component::Prefix(_) | Component::RootDir | Component::ParentDir => {
+                            return None
```
jplatte, line 112:
```suggestion
                            return None;
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -629,31 +737,88 @@ fn handle_file_request(
...
+where
+    ReqBody: Send + 'static,
+{
+    type Response = Response<ResponseBody>;
+    type Error = io::Error;
+    type Future = ResponseFuture<ReqBody>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // contains `Infallible` so can never be called
+        unreachable!()
```
jplatte, line 768:
```suggestion
        match self.0 {}
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -629,31 +737,87 @@ fn handle_file_request(
...
+    type Error = io::Error;
+    type Future = ResponseFuture<ReqBody>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        match self.0 {}
+    }
+
+    fn call(&mut self, _req: Request<ReqBody>) -> Self::Future {
+        // contains `Infallible` so can never be called
+        unreachable!()
```
jplatte, line 772: Same here.

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -260,6 +224,75 @@ impl ServeDir {
...
+    /// # Example
+    ///
+    /// This can be used to respond with a different file:
+    ///
+    /// ```rust
+    /// use tower_http::services::{ServeDir, ServeFile};
+    ///
+    /// let service = ServeDir::new("assets")
+    ///     // respond with `404 Not Found` and the contents of `index.html` for missing files
+    ///     .not_found_service(ServeFile::new("assets/index.html"));
```
jplatte, line 280:
```suggestion
    ///     .not_found_service(ServeFile::new("assets/not_found.html"));
```

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -260,6 +224,75 @@ impl ServeDir {
...
+    /// // Run our service using `hyper`
+    /// let addr = std::net::SocketAddr::from(([127, 0, 0, 1], 3000));
+    /// hyper::Server::bind(&addr)
+    ///     .serve(tower::make::Shared::new(service))
+    ///     .await
+    ///     .expect("server error");
+    /// # };
+    /// ```
+    ///
+    /// Setups like this are often found in single page applications.
```
jplatte, line 292: This wasn't supposed to be part of both method's docs, right?

tower-http/src/services/fs/serve_dir.rs
```diff
@@ -285,26 +318,63 @@ async fn maybe_redirect_or_append_path(
...
+        if let Some(fallback) = &mut self.fallback {
+            fallback.poll_ready(cx).map_err(Into::into)
+        } else {
+            Poll::Ready(Ok(()))
+        }
     }
 
     fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
-        let mut full_path = match self.variant.full_path(&self.base, req.uri().path()) {
+        // `ServeDir` doesn't care about the request body but the fallback might. So move out the
```
jplatte, line 343: All methods that are reasonable for asset routes don't have a body anyways, right? By the way, does `ServeDir` still serve files for non-GET requests? I think it did that at some point.

jplatte (approved):
Apart from latest comments, looks good!

tower-http/src/services/fs/serve_dir.rs
davidpdrsn: > All methods that are reasonable for asset routes don't have a body anyways, right?

Yes. At least I can't think of a use case for serving a static file and having a request body.

> By the way, does ServeDir still serve files for non-GET requests? I think it did that at some point.

It still accepts all methods. Perhaps we should change that to respond with 405 for non-`GET` requests.

jplatte: There is proper handling of HEAD now, right? That should probably stay. Not sure about OPTIONS, probably okay to have it return 405 if the user doesn't have a cors middleware set up.

PR #246, opened 2022-04-14
fix: add default to trace response body

conradludgate:
## Motivation

This is an attempt to fix https://github.com/hyperium/tonic/issues/973

## Solution

Adds `Default` to `trace::ResponseBody`.

I'm not sure on the implications this may have however

### DIFF
--- a/tower-http/src/trace/body.rs
+++ b/tower-http/src/trace/body.rs
@@ -28,6 +28,25 @@ pin_project! {
     }
 }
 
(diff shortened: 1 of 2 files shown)


davidpdrsn:
I don't think this is appropriate and I'm not sure it would work in practice due to the generics. See my response [here](https://github.com/hyperium/tonic/issues/973#issuecomment-1099258099).

Issue #215, opened 2022-02-06
CompressionLayer: `accept-encoding` header parsed incorrectly


mdickopp:
### Version

tower-http v0.2.1

### Platform

Debian Linux 12 (“bookworm”)
Linux feynman 5.15.0-3-amd64 <span>#</span>1 SMP Debian 5.15.15-1 (2022-01-18) x86_64 GNU/Linux
rustc 1.58.1 (db9d1b20b 2022-01-20)

### Description

When using `CompressionLayer`, the `accept-encoding` header sent by the client is not parsed correctly (i.e., according to RFC 7231, sections 5.3.1 and 5.3.4). The following program demonstrates the issues (using axum v0.4.5):

```rust
use axum::{routing::get, Router};
use tower_http::compression::{predicate::SizeAbove, CompressionLayer};

#[tokio::main]
async fn main() {
    let app = Router::new()
        .route("/", get(|| async { "" }))
        .layer(CompressionLayer::new().compress_when(SizeAbove::new(0)));
    axum::Server::bind(&"0.0.0.0:3000".parse().unwrap())
        .serve(app.into_make_service())
        .await
        .unwrap();
}
```

#### Uppercase encodings and qvalues are not parsed

Encodings and qvalues are case-insensitive, i.e. the server should understand them whether they are lowercase, uppercase, or a mixture of both.

| Command | Expected Encoding | Observed Encoding |
| --- | --- | --- |
| `curl -I -H 'accept-encoding: GZIP' http://127.0.0.1:3000/` | gzip | (none) |
| `curl -I -H 'accept-encoding: gZiP' http://127.0.0.1:3000/` | gzip | (none) |
| `curl -I -H 'accept-encoding: gzip;q=0.5, br;Q=0.8' http://127.0.0.1:3000/` | br | gzip |

#### Spaces before and after semicolon are not parsed

Space and horizontal tab characters are allowed before and after the semicolon separating the encoding from the qvalue.

| Command | Expected Encoding | Observed Encoding |
| --- | --- | --- |
| `curl -I -H 'accept-encoding: gzip;q=0.5, br; q=0.8' http://127.0.0.1:3000/` | br | gzip |
| `curl -I -H 'accept-encoding: gzip;q=0.5, br ;q=0.8' http://127.0.0.1:3000/` | br | gzip |
| `curl -I -H 'accept-encoding: gzip;q=0.5, br ; q=0.8' http://127.0.0.1:3000/` | br | gzip |

[...]

PR #247, opened 2022-04-15
Handle `*` in `Accept-Encoding` request header correctly

mdickopp:
## Motivation

This PR fixes case 6 of bug report #215.

## Solution

`*` in the `Accept-Encoding` header means “all other encodings”, i.e. those not explicitly specified. I modified the `tower_http::content_encoding::encodings` function to deal with this case and added a unit test.

As a beneficial side effect, duplicate encodings encountered in the `Accept-Encoding` header are consolidated, so that the `encodings` function does not return duplicates a (buggy or malicious) client may have specified.

### DIFF
--- a/tower-http/src/content_encoding.rs
+++ b/tower-http/src/content_encoding.rs
@@ -1,45 +1,42 @@
 pub(crate) trait SupportedEncodings: Copy {
-    fn br(&self) -> bool;
     fn gzip(&self) -> bool;
     fn deflate(&self) -> bool;
+    fn br(&self) -> bool;
 }
 
-// The order of the variants is important: If the client has equal highest preference for multiple
-// encodings, the first of these is chosen. If the client prefers the `*` encoding, the first
-// variant not explicitly mentioned by the client is chosen.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub(crate) enum Encoding {
-    #[allow(dead_code)]
-    Identity,
-    #[cfg(any(feature = "fs", feature = "compression-br"))]
-    Brotli,
     #[cfg(any(feature = "fs", feature = "compression-gzip"))]
     Gzip,
     #[cfg(any(feature = "fs", feature = "compression-deflate"))]
     Deflate,
+    #[cfg(any(feature = "fs", feature = "compression-br"))]
+    Brotli,
+    #[allow(dead_code)]
+    Identity,
 }
 
 impl Encoding {
     #[allow(dead_code)]
     fn to_str(self) -> &'static str {
         match self {
-            Encoding::Identity => "identity",
-            #[cfg(any(feature = "fs", feature = "compression-br"))]
-            Encoding::Brotli => "br",
             #[cfg(any(feature = "fs", feature = "compression-gzip"))]
             Encoding::Gzip => "gzip",
             #[cfg(any(feature = "fs", feature = "compression-deflate"))]
             Encoding::Deflate => "deflate",
+            #[cfg(any(feature = "fs", feature = "compression-br"))]
+            Encoding::Brotli => "br",
+            Encoding::Identity => "identity",
         }
     }
 
     #[cfg(feature = "fs")]
     pub(crate) fn to_file_extension(self) -> Option<&'static std::ffi::OsStr> {
         match self {
-            Encoding::Identity => None,
-            Encoding::Brotli => Some(std::ffi::OsStr::new(".br")),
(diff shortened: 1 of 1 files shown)


Nehliin:
I'll take a look at this tomorrow!

tower-http/src/content_encoding.rs
```diff
@@ -201,32 +204,57 @@ pub(crate) fn encodings(
...
+        }
+    }
 
-            Some((encoding, qval))
-        })
-        .collect::<Vec<(Encoding, QValue)>>()
+    // The wildcard encoding (`*`) means all encodings not mentioned explicitly. If a wildcard
+    // encoding has been specified, set the qvalues of all unset encodings accordingly.
+    if wildcard_qval.is_some() {
+        let wildcard_qval = wildcard_qval.unwrap();
```
Nehliin, line 247: Use `if let Some(wildcard_qval) = ...` instead to avoid unwrapping here`

tower-http/src/content_encoding.rs
```diff
@@ -201,32 +204,57 @@ pub(crate) fn encodings(
     headers: &http::HeaderMap,
     supported_encoding: impl SupportedEncodings,
 ) -> Vec<(Encoding, QValue)> {
-    headers
+    // Mapping of encodings to corresponding qvalues.
+    let mut encodings = std::collections::BTreeMap::default();
```
Nehliin, line 208: BTreeMap is probably going to be slower here than a simple vec that you sort after pushing everything to it since the max size here is 4. Even if you also need to check for duplicates before inserting the linear search is most likely going to be faster.

tower-http/src/content_encoding.rs
```diff
@@ -1,42 +1,45 @@
...
+    fn br(&self) -> bool;
     fn gzip(&self) -> bool;
     fn deflate(&self) -> bool;
-    fn br(&self) -> bool;
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+// The order of the variants is important: If the client has equal highest preference for multiple
+// encodings, the first of these is chosen. If the client prefers the `*` encoding, the first
+// variant not explicitly mentioned by the client is chosen.
```
Nehliin, line 9: Where did you find what the order should be here (after identity)? I couldn't find it in the specification when skimming it where it stated that Br for example should have higher prio than gzip if both have equal q values

Nehliin (changes requested):
Thanks for the pr, have some small comments

tower-http/src/content_encoding.rs
mdickopp: Done!

tower-http/src/content_encoding.rs
mdickopp: I changed the code to store the data in a `Vec<(Encoding, QValue)>`.

tower-http/src/content_encoding.rs
```diff
@@ -565,4 +627,31 @@ mod tests {
...
+
+    #[test]
+    fn accept_encoding_header_with_wildcard() {
+        let mut headers = http::HeaderMap::new();
+        headers.append(
+            http::header::ACCEPT_ENCODING,
+            http::HeaderValue::from_static("*;q=0.8,gzip;q=0.5"),
+        );
+        let encoding = Encoding::from_headers(&headers, SupportedEncodingsAll::default());
+        assert_eq!(Encoding::Identity, encoding);
```
Nehliin, line 639: I'm not sure we want to use identity here as the first option since we should be able to provide another compression algorithm other than gzip and still follow the specification (if the server supports it). 

> The asterisk "*" symbol in an Accept-Encoding field matches any available content-coding not explicitly listed in the header field.

Which should mean brotli for example is acceptable in this case. Forgive me if this was what you did before my previous comments on this pr because in that case I agree we should change the order we select compression algorithm to prevent situations like this.

Nehliin (changes requested):
Thank you for your work in this, sorry for my late responses. I have been away the last two weeks so haven't had time to look at this. I had one final comment which I think you actually solved before my last pr comments.

mdickopp:
@Nehliin Sorry, I'm a bit confused now. Could you clarify what behavior you expect when the client specifies multiple encodings with same preference explicitly (e.g. `Accept-Encoding: gzip. deflate, br`), and what you expect when the client specifies multiple encodings with same preference by use of `*`?

davidpdrsn:
@mdickopp are still interested in working on this?

mdickopp:
@davidpdrsn I have no time to work on this at the moment. If anyone wants to take it over, please do.

davidpdrsn:
Okay. That's fine! Thanks for letting us know.

I'll pick it up.

PR #249, opened 2022-04-24
Check method in `ServeDir`

davidpdrsn:
## Motivation

Methods outside of `GET` and `HEAD` don't make much sense for serving static files but `ServeDir` (and therefore also `ServeFile`) would just accept all methods.

## Solution

Check the method in `ServeDir` and respond with `405 Method Not Allowed` if its not `GET` or `HEAD`. A response is immediately returned without calling the fallback.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `ServeDir::{fallback, not_found_service}` for calling another service if
   the file cannot be found.
 - Add `SetStatus` to override status codes.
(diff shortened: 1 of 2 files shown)


tower-http/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `ServeDir::{fallback, not_found_service}` for calling another service if
   the file cannot be found.
 - Add `SetStatus` to override status codes.
+- `ServeDir` and `ServeFile` now responds with `405 Method Not Allowed` to requests where the
```
jplatte, line 15:
```suggestion
- `ServeDir` and `ServeFile` now respond with `405 Method Not Allowed` to requests where the
```

jplatte (approved):

PR #250, opened 2022-04-24
Split `serve_dir.rs` into multiple files

davidpdrsn:
`serve_dir.rs` was getting quite big (1.4k lines) so thought it made sense to break up. This breaks the code into these modules:

- `future`: Contains the response future and response generation code
- `headers`: Contains some newtypes for a few headers. Was previously defined in the `fs` module but only used in `serve_dir`
- `open_file`: Contains code for opening the file or possibly returning a response early
- `tests`: Contains, well, the tests

Other than moving code around I made very few changes. The biggest one is renaming `Output` to `OpenFileOutput` and making the enum variants more specific. For example previously it contained a `StatusCode` variant which I replaced with explicit `NotFound`, `PreconditionFailed`, and `NotModified` variants.


tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -0,0 +1,637 @@
...
+use bytes::Bytes;
+use flate2::bufread::{DeflateDecoder, GzDecoder};
+use http::{header, Method, Response};
+use http::{Request, StatusCode};
+use http_body::Body as HttpBody;
+use hyper::Body;
+use std::io::{self, Read};
+use tower::ServiceExt;
+
+#[tokio::test]
```
Nehliin, line 12: Unrelated to your changes but I see now that we are missing a test for the security issue we had a while back unless I'm misstaken. Where the user could access arbitrary files outside the directory served like: `/foo/bar/c:/secret.txt`

davidpdrsn: From looking at https://github.com/tower-rs/tower-http/pull/204 it doesn't seem like we ever had a test for this. We probably should.

Nehliin: yep

davidpdrsn: I've made an issue for it https://github.com/tower-rs/tower-http/issues/251

Nehliin (approved):
Nice

PR #256, opened 2022-04-27
cors: Fix handling of multiple allowed origins

### LINKED ISSUE #255: CORS: Multiple origins for access-control-allow-origin when using `AllowOrigin::list`
### Version

0.3.0

### Description

I tried this code:

```rs
CorsLayer::new().allow_methods(Any).allow_origin(AllowOrigin::list([
    "http://localhost:3000",
    "http://192.168.1.10:3000",
]));
```

Send requests from origin `http://localhost:3000`.

I expected to see this happen:
```http
access-control-allow-origin: http://localhost:3000
```

Instead, this happened:
```http
access-control-allow-origin: http://localhost:3000,http://192.168.1.10:3000
```

According to [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin), there should not be multiple origins.

jplatte:
Fixes #255.

### DIFF
--- a/tower-http/src/cors/allow_origin.rs
+++ b/tower-http/src/cors/allow_origin.rs
@@ -5,7 +5,7 @@ use http::{
     request::Parts as RequestParts,
 };
 
-use super::{separated_by_commas, Any, WILDCARD};
+use super::{Any, WILDCARD};
 
 /// Holds configuration for how to set the [`Access-Control-Allow-Origin`][mdn] header.
 ///
@@ -24,7 +24,7 @@ impl AllowOrigin {
     ///
     /// [`CorsLayer::allow_origin`]: super::CorsLayer::allow_origin
     pub fn any() -> Self {
-        Self(OriginInner::Const(Some(WILDCARD)))
+        Self(OriginInner::Const(WILDCARD))
     }
 
     /// Set a single allowed origin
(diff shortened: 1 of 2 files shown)


tower-http/src/cors/allow_origin.rs
```diff
@@ -129,14 +129,15 @@ impl From<Vec<HeaderValue>> for AllowOrigin {
 
 #[derive(Clone)]
 enum OriginInner {
-    Const(Option<HeaderValue>),
+    Const(HeaderValue),
+    List(Vec<HeaderValue>),
```
davidpdrsn, line 133: Maybe

```suggestion
    List(Arc<[HeaderValue]>),
```

davidpdrsn (approved):

tower-http/src/cors/allow_origin.rs
jplatte: Why? Do you expect the layer to be cloned commonly?

davidpdrsn: The layer no but the service contains a layer, so yes.

jplatte: And the service will be cloned a lot? Why? Also maybe we should have a single `Arc` around the current fields of the `Layer` then? This seems unrelated in any case.

davidpdrsn: Yes services are often cloned to get an owned value that can be moved into the response future. It doesn't have to happen in this PR but is something to keep in mind.

PR #258, opened 2022-04-28
Properly clone the request for ServeDir's fallback

sandhose:
## Motivation

I tried to chain two ServeDir (one being the fallback for the other) and got confused why it could not find files from the second one.
It turns out, the request passed to the fallback service was not cloned properly from the original request.

## Solution

Renamed the fallback request variable name to properly clone the method/uri/headers/extensions from the original request, and not from itself.
I modified the fallback test to include the URI path in the response and check if it matches.


jplatte (approved):

davidpdrsn (approved):
Omg how didn't we catch this 😢 Thanks for finding it!

PR #259, opened 2022-04-29
Release 0.3.2

davidpdrsn:


tower-http/CHANGELOG.md
```diff
@@ -19,9 +19,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-## Fixes
+## Fixed
+
+- None.
+
+# 0.3.2 (April 29, 2022)
+
+## Fixed
+
+- **serve_dir**: Properly clone the request for `ServeDir`'s fallback ([#258])
```
jplatte, line 30: I think it would be good to highlight what this means for users rather than what was done internally. Something like

```suggestion
- **serve_dir**: Fix empty request parts being passed to `ServeDir`'s fallback instead of the actual ones ([#258])
```

davidpdrsn: I agree!

jplatte (approved):

PR #264, opened 2022-05-08
Support calling fallback on `405 Method Not Allowed`

davidpdrsn:
Previously `ServeDir` fallbacks would only be called if the request file wasn't found, not if it received a non `GET` or `HEAD` request.

However for serving assets at the root in a router it might be useful to delegate all requests to the fallback, regardless of the method. Something like:

```rust
ServeDir::new(".")
    .call_fallback_on_method_not_allowed(true)
    .fallback(Router::new().route("/", ...));
```

### DIFF
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -249,7 +249,7 @@ impl<F> ServeDir<F> {
 
     /// Customize whether or not to call the fallback for requests that aren't `GET` or `HEAD`.
     ///
-    /// Defaults not calling the fallback and instead returning `405 Method Not Allowed`.
+    /// Defaults to not calling the fallback and instead returning `405 Method Not Allowed`.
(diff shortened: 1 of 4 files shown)


tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -242,6 +246,14 @@ impl<F> ServeDir<F> {
     pub fn not_found_service<F2>(self, new_fallback: F2) -> ServeDir<SetStatus<F2>> {
         self.fallback(SetStatus::new(new_fallback, StatusCode::NOT_FOUND))
     }
+
+    /// Customize whether or not to call the fallback for requests that aren't `GET` or `HEAD`.
+    ///
+    /// Defaults not calling the fallback and instead returning `405 Method Not Allowed`.
```
jplatte, line 252:
```suggestion
    /// Defaults to not calling the fallback and instead returning `405 Method Not Allowed`.
```

jplatte (approved):
Otherwise LGTM.

PR #270, opened 2022-05-17
Add `Timeout` middleware

davidpdrsn:
`tower::timeout::Timeout` doesn't work well with axum since it changes the error type to `BoxError`. That requires doing this:

```rust
let app = Router::new()
    .route("/", get(|| async {}))
    .layer(
        ServiceBuilder::new()
            .layer(HandleErrorLayer::new(handle_timeout_error))
            .timeout(Duration::from_secs(30))
    );

async fn handle_timeout_error(err: BoxError) -> (StatusCode, String) {
    if err.is::<tower::timeout::error::Elapsed>() {
        (
            StatusCode::REQUEST_TIMEOUT,
            "Request took too long".to_string(),
        )
    } else {
        (
            StatusCode::INTERNAL_SERVER_ERROR,
            format!("Unhandled internal error: {}", err),
        )
    }
}
```

Thats quite a mouthful. With this middleware that becomes:

```rust
let app = Router::new()
    .route("/", get(|| async {}))
    .layer(TimeoutLayer::new(Duration::from_secs(30));
```

This works because it doesn't change the service's error type and thus keeps `Infallible` which axum requires.

### DIFF
--- a/tower-http/src/timeout.rs
+++ b/tower-http/src/timeout.rs
@@ -9,7 +9,7 @@
 //! it changes the error type to [`BoxError`](tower::BoxError). For HTTP services that is rarely
 //! what you want as returning errors will terminate the connection without sending a response.
 //!
-//! Instead this middleware wont change the error type and instead return a `408 Request Timeout`
+//! This middleware won't change the error type and instead return a `408 Request Timeout`
 //! response. That means if your service's error type is [`Infallible`] it will still be
 //! [`Infallible`] after applying this middleware.
 //!
--- a/tower-http/src/timeout.rs
+++ b/tower-http/src/timeout.rs
@@ -1,0 +1,156 @@
+//! Middleware that applies a timeout to requests.
+//!
+//! If the request does not complete within the specified timeout it will be aborted and a `408
+//! Request Timeout` response will be sent.
+//!
+//! # Differences from `tower::timeout`
+//!
+//! tower's [`Timeout`](tower::timeout::Timeout) middleware uses an error to signal timeout, i.e.
+//! it changes the error type to [`BoxError`](tower::BoxError). For HTTP services that is rarely
+//! what you want as returning errors will terminate the connection without sending a response.
+//!
+//! Instead this middleware wont change the error type and instead return a `408 Request Timeout`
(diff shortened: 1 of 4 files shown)


tower-http/src/timeout.rs
```diff
@@ -0,0 +1,156 @@
...
+//! If the request does not complete within the specified timeout it will be aborted and a `408
+//! Request Timeout` response will be sent.
+//!
+//! # Differences from `tower::timeout`
+//!
+//! tower's [`Timeout`](tower::timeout::Timeout) middleware uses an error to signal timeout, i.e.
+//! it changes the error type to [`BoxError`](tower::BoxError). For HTTP services that is rarely
+//! what you want as returning errors will terminate the connection without sending a response.
+//!
+//! Instead this middleware wont change the error type and instead return a `408 Request Timeout`
```
jplatte, line 12:
```suggestion
//! This middleware won't change the error type and instead return a `408 Request Timeout`
```

jplatte (approved):
LGTM, except one somewhat clunky sentence in the docs.

LucioFranco:
I wonder if this should just either 1) wrap the main tower timeout and just map the error (with a way to customize it, 2) we should just make tower timeout more flexible.

Also, this rases a question can you use other regular tower services or are you limited to ones just in `tower-http` because of the error shape?

davidpdrsn:
> I wonder if this should just either 1) wrap the main tower timeout and just map the error (with a way to customize it, 2) we should just make tower timeout more flexible.

I think its okay for tower-http to provide its own without wrapping tower's, since timeouts are so simple. For more complex middleware it might be better to wrap tower's.

> Also, this rases a question can you use other regular tower services or are you limited to ones just in tower-http because of the error shape?

For axum there are no limits thanks to [`HandleErrorLayer`](https://docs.rs/axum/latest/axum/error_handling/struct.HandleErrorLayer.html) and [`MethodRouter::handle_error`](https://docs.rs/axum/latest/axum/routing/struct.MethodRouter.html#method.handle_error) so its mostly a question of ergonomics.

I haven't don't often use other middleware from tower (except timeout) that changes the error type.

PR #271, opened 2022-05-20
feat: add layer that limits body size

neoeinstein:
## Motivation

To provide a tower layer that limits the allowable length of the request body to a certain amount.

## Solution

Add a new tower layer that wraps the incoming body in `http_body::Limited<B>`.

A couple of decisions:
* Do we want this layer to automatically return a 419 Payload Too Large error?
* Are we okay with the way the errors and response body are all boxed up in that event?

### DIFF
--- a/tower-http/src/limit.rs
+++ b/tower-http/src/limit.rs
@@ -1,0 +1,246 @@
+//! Imposes a length limit on request bodies.
+//!
+//! # Example
+//!
+//! ```rust
+//! use bytes::Bytes;
+//! use http::{Request, Response, StatusCode};
+//! use std::convert::Infallible;
+//! use tower::{Service, ServiceExt, ServiceBuilder, service_fn};
+//! use tower_http::limit::LengthLimitedLayer;
+//! use hyper::Body;
+//! use http_body::Limited;
+//! use tower_http::BoxError;
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), BoxError> {
+//!  async fn handle(req: Request<Limited<Body>>) -> Result<Response<Body>, BoxError>
+//! {
+//!     hyper::body::to_bytes(req.into_body()).await?;
+//!     Ok(Response::new(Body::empty()))
+//! }
+//!
+//! let mut svc = ServiceBuilder::new()
+//!     // Limit incoming requests to 4096 bytes.
+//!     .layer(LengthLimitedLayer::new(4096))
+//!     .service_fn(handle);
+//!
+//! fn test_svc<S: Service<Request<Body>>>(s: &S) {}
+//! test_svc(&svc);
+//!
+//! // Call the service.
+//! let request = Request::new(Body::empty());
+//!
+//! let response = svc.ready().await?.call(request).await?;
+//!
+//! assert_eq!(response.status(), 200);
+//!
+//! // Call the service with a body that is too large.
+//! let request = Request::new(Body::from(Bytes::from(vec![0u8; 4097])));
+//!
+//! let response = svc.ready().await?.call(request).await?;
+//!
+//! assert_eq!(response.status(), StatusCode::PAYLOAD_TOO_LARGE);
+//!
+//! #
+//! # Ok(())
+//! # }
+//! ```
+
+use crate::BoxError;
+use bytes::Bytes;
+use http::{HeaderValue, Request, Response, StatusCode};
+use http_body::combinators::UnsyncBoxBody;
+use http_body::{Body, Full, LengthLimitError, Limited};
+use pin_project_lite::pin_project;
+use std::error::Error;
+use std::future::Future;
+use std::pin::Pin;
+use std::{
+    any, fmt,
+    marker::PhantomData,
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+/// Layer that applies the [`LengthLimit`] middleware that intercepts requests
+/// with body lengths greater than the configured limit and converts them into
+/// `413 Payload Too Large` responses.
+///
+/// See the [module docs](self) for an example.
+pub struct LengthLimitedLayer<B> {
+    limit: usize,
+    _ty: PhantomData<fn() -> B>,
+}
+
+impl<B> LengthLimitedLayer<B> {
+    /// Create a new `LengthLimitedLayer` with the given body length limit.
+    pub fn new(limit: usize) -> Self {
+        Self {
+            limit,
+            _ty: PhantomData,
+        }
+    }
+}
+
+impl<B> Clone for LengthLimitedLayer<B> {
+    fn clone(&self) -> Self {
+        Self {
+            limit: self.limit,
+            _ty: PhantomData,
+        }
+    }
+}
+
+impl<B> Copy for LengthLimitedLayer<B> {}
+
+impl<B> fmt::Debug for LengthLimitedLayer<B> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LengthLimitedLayer")
+            .field("body", &any::type_name::<B>())
+            .field("limit", &self.limit)
+            .finish()
+    }
+}
+
+impl<B, S> Layer<S> for LengthLimitedLayer<B> {
+    type Service = LengthLimited<S, B>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        LengthLimited {
+            inner,
+            limit: self.limit,
+            _ty: PhantomData,
+        }
+    }
+}
+
+/// Middleware that intercepts requests with body lengths greater than the
+/// configured limit and converts them into `413 Payload Too Large` responses.
+///
+/// See the [module docs](self) for an example.
+pub struct LengthLimited<S, B> {
+    inner: S,
+    limit: usize,
+    _ty: PhantomData<fn() -> B>,
+}
+
+impl<S, B> LengthLimited<S, B> {
+    define_inner_service_accessors!();
+
+    /// Create a new `LengthLimited` with the given body length limit.
+    pub fn new(inner: S, limit: usize) -> Self {
+        Self {
+            inner,
+            limit,
+            _ty: PhantomData,
+        }
+    }
+}
+
+impl<S, B> Clone for LengthLimited<S, B>
+where
+    S: Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            limit: self.limit,
+            _ty: PhantomData,
+        }
+    }
+}
+
+impl<S, B> Copy for LengthLimited<S, B> where S: Copy {}
+
+impl<S, B> fmt::Debug for LengthLimited<S, B>
+where
+    S: fmt::Debug,
+{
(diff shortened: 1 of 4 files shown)


jplatte:
> Do we want this layer to automatically return a 419 Payload Too Large error?

Without having looked at the implementation or considering any previous disussion around this, this seems like an obvious 'Yes' to me.

tower-http/src/limit.rs
```diff
@@ -0,0 +1,246 @@
...
+            Poll::Ready(Ok(data)) => {
+                let (parts, body) = data.into_parts();
+                let body = body.map_err(|err| err.into()).boxed_unsync();
+                let resp = Response::from_parts(parts, body);
+
+                Poll::Ready(Ok(resp))
+            }
+            Poll::Ready(Err(err)) => {
+                let err = err.into();
+                if let Some(_) = err.downcast_ref::<LengthLimitError>() {
```
davidpdrsn, line 225: Hm does this actually work? This checks the error for `call`, not the `Body::poll_data` which is what yields `LengthLimitError`.

tower-http/src/limit.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let (parts, body) = req.into_parts();
+        let body = Limited::new(body, self.limit);
+        let req = Request::from_parts(parts, body);
+
+        ResponseFuture {
+            future: self.inner.call(req),
+        }
+    }
```
davidpdrsn, line 193: I think we should also check if there is `content-length` header and return `413 Payload Too Large` if its larger than the limit.

tower-http/src/limit.rs
```diff
@@ -0,0 +1,246 @@
...
+            _ty: PhantomData,
+        }
+    }
+}
+
+/// Middleware that intercepts requests with body lengths greater than the
+/// configured limit and converts them into `413 Payload Too Large` responses.
+///
+/// See the [module docs](self) for an example.
+pub struct LengthLimited<S, B> {
```
davidpdrsn, line 123: What do you think about calling it `RequestBodyLimit`? Then its more clear its limiting the request body and not the response body.

tower-http/src/limit.rs
```diff
@@ -0,0 +1,246 @@
...
+    inner: S,
+    limit: usize,
+    _ty: PhantomData<fn() -> B>,
```
davidpdrsn, line 126: Does it actually need to be generic over the request body?

davidpdrsn:
> Do we want this layer to automatically return a 419 Payload Too Large error?

Yes I think we should. Except we can't do that unless the request has a `content-length` I think. Otherwise we'd have to buffer the request body first which might be a long stream.

neoeinstein:
> Yes I think we should. Except we can't do that unless the request has a `content-length` I think. Otherwise we'd have to buffer the request body first which might be a long stream.

If you take a look at how this is set up right now, in combination with `http_body::Limited`, this will immediately cease reading any more of the incoming body once the limit has been reached and return the 419. If you want more of a guarantee of that, I can set up a test case asserting this.

tower-http/src/limit.rs
neoeinstein: It does, but there may be some restrictions that require a bit more to make this work as a layer. In particular, if the body is never read by the handler, then there will never be any error returned for the body being too large. For reference, note that the doctest for this module requires the handler to consume the body.

However, if there is an attempt to consume the body, and it is consumed beyond the given limit, the read of the body is what will trigger the error. We'd expect this error to be propagated out of the underlying service in some way.

Right now, this does expect that the top-level error be the (boxed) `LengthLimitError`, so if someone propagates the error by wrapping it inside an error enum, this wouldn't pick it out right. I can make a modification here to try to walk the source chain, which would then cover this case.

tower-http/src/limit.rs
neoeinstein: That should be relatively simple to do. It will be a bit simpler than implemented in `axum` because we won't need to worry about disallowing requests lacking a `Content-Length` header.

tower-http/src/limit.rs
```diff
@@ -0,0 +1,246 @@
...
+impl<S, B> Copy for LengthLimited<S, B> where S: Copy {}
+
+impl<S, B> fmt::Debug for LengthLimited<S, B>
+where
+    S: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LengthLimited")
+            .field("inner", &self.inner)
+            .field("service", &format_args!("{}", any::type_name::<B>()))
```
neoeinstein, line 164: This should identify the body instead.

tower-http/src/limit.rs
neoeinstein: I did try that once, and I ended up in a situation where the request body was needed as a witness somewhere. I'll try to see if I can reduce that out again.

tower-http/src/limit.rs
davidpdrsn: I see. I don't think we should change the error to `BoxError`. That makes it harder to use with axum. I imagined someone would try and consume the body, if that fails they'd downcast the error to `LengthLimitError` and build a response themselves from that. Then the error type wouldn't have to change.

tower-http/src/limit.rs
davidpdrsn: Are you saying you'll look into it?

neoeinstein: Yes. I'll incorporate this into the layer.

tower-http/src/limit.rs
davidpdrsn: I think we should remove it. I think its gonna give type inference errors when you apply multiple layers.

tower-http/src/limit.rs
neoeinstein: Consumers wouldn't need to return `BoxError` themselves, but just an error type that can be converted into a `BoxError`. Let me put my thought together into this PR and we can take a look at it from there. I'll also update the doctest to show how a custom, non-`BoxError` would look.

neoeinstein: Also, if an underlying service wanted to trap the underlying `LengthLimitError` themselves, they would still be free to do that. This would just act as a backstop to attempt to return a correct response. Without it, hyper would just sever the connection on a service returning an error. See the comments in `hyper:service::http`.

Because Axum services always require that their handlers end up returning something that can be a `Response` and have an `Infallible` error type, this backstop may not be useful in the Axum context, but for those writing handlers directly on top of hyper, this could do the right thing for the difficult chunked body case.

tower-http/src/limit.rs
neoeinstein: This has been done. I've also added some doctests that demonstrate the various cases.

neoeinstein: I added a fourth usage example for those that do not want to have automatic 413 responses, pointing to use `MapRequestBody` to wrap the inner body with `Limited`.

tower-http/src/limit.rs
davidpdrsn: > Consumers wouldn't need to return BoxError themselves, but just an error type that can be converted into a BoxError

Yes and that will make it unergonomic to use with axum as it _forces_ the error type to be `BoxError` even if your service actually never fails. It'll require doing this https://docs.rs/axum/latest/axum/error_handling/index.html#applying-fallible-middleware. Therefore I still think we should keep the error type from the inner service.

davidpdrsn: I don't think using `MapRequestBody` is a good workaround as you loose out on `content-length` checking.

davidpdrsn: I wrote the axum integration and use case I have in mind [here](https://github.com/tokio-rs/axum/pull/1048).

neoeinstein: `MapRequestBody` as a suggestion isn't meant to be a workaround, but rather a pointer for users that want to impose a body limit _without_ also imposing a content-length check and automatic response and to head off any request to make this middleware configureable in that way.

neoeinstein: Did a couple example runs with the Axum integration you provided.

One byte too long, with immedate follow-on request (_hyper_ is able to resync safely):
bildo

With a chunked body that is beyond _hyper_'s resync attempt. In this case _hyper_ severs the connection since it cannot resume (so second request is dropped):
```
HTTP/1.1 413 Payload Too Large
content-type: text/plain; charset=utf-8
content-length: 56
date: Mon, 23 May 2022 19:24:30 GMT

Failed to buffer the request body: length limit exceeded
```

And what the middleware returns when it rejects the request on `Content-Length` grounds (just a little over the limit, so _hyper_ resyncs):
```
HTTP/1.1 413 Payload Too Large
content-type: text/plain; charset=utf-8
content-length: 21
date: Mon, 23 May 2022 19:29:14 GMT

length limit exceeded
```

If the payload is significantly too large, then _hyper_ fails to resync after one attempt and severs the connection.

neoeinstein:
I made some changes based on the feedback. Now the future doesn't attempt to interpret the error returned by the underlying service. This removes the `StdError + 'static` restriction on `S::Error`. It does mean that something else (like `axum`) needs to be able to convert a failure to read from `Limited<B>` into an appropriate response, but I think that's the better option.

This now also interprets the `Content-Length` header and returns early with a 413 in the event that the payload would be too large.

Doc examples have been updated as well.

tower-http/src/limit.rs
davidpdrsn: What are the implications of that? That clients will have to open a new connection?

davidpdrsn:
Amazing! Thank you for working on this. I'll review it this week :)

tower-http/src/limit.rs
neoeinstein: If hyper can't re-sync, it will indeed sever the connection and the direct client would need to re-establish a connection. If the request is proxied by a load balancer of some sort, it may have better tolerances for re-syncing the incoming connection. This is server-dependent, however, so I point this out in the documentation without overly-alluding to _hyper_'s specific implementation.

Basically, hyper attempts to read one more chunk (with `transfer-encoding: chunked`) or one more buffer full (with `content-length: <len>`). If it hasn't reached the end of the current HTTP message frame by that point, then it will kill the connection to reset things to a known acceptable state.

davidpdrsn (commented):
@neoeinstein I think it looks good! I pushed a bunch of nit picks but had one question I'd like to hear what you think about.

tower-http/src/limit/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+//! This layer will also intercept requests with a `Content-Length` header
+//! larger than the allowable limit and return an immediate error response
+//! before reading any of the body.
+//!
+//! Note that payload length errors can be used by adversaries in an attempt
+//! to smuggle requests. When an incoming stream is dropped due to an
+//! over-sized payload, servers should close the connection or resynchronize
+//! by optimistically consuming some data in an attempt to reach the end of
+//! the current HTTP frame. If the incoming stream cannot be resynchronized,
+//! then the connection should be closed.
```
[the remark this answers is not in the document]
davidpdrsn: Alright. I'll add a note that mentions hyper.

PR #275, opened 2022-06-25
Add `NormalizePath` middleware

davidpdrsn:
In axum we're removing support for trailing slash redirects (see https://github.com/tokio-rs/axum/issues/1118 for more details). That means its a little more work for users if they want their routes to be matched both with and without a trailing slash. This middleware addresses that.

Example usage with axum:

```rust
// what you'd have to do without this middleware
Router::new()
    .route("/foo", get(handler))
    .route("/foo/", get(handler));

// using the middleware we only have to add one route
NormalizePath::trim_trailing_slash(
    Router::new().route("/foo", get(handler))
);
```

Note that axum's `Router::layer` cannot be used since routing happens before any middleware are called, thus the requests would 404. That is something we have to address separately in axum but I still think this middleware makes sense on its own.

### DIFF
--- a/tower-http/src/normalize_path.rs
+++ b/tower-http/src/normalize_path.rs
@@ -1,0 +1,192 @@
+//! Middleware that normalizes paths.
+//!
+//! Any trailing slashes from request paths will be removed. For example, a request with `/foo/`
+//! will be changed to `/foo` before reaching the inner service.
+//!
+//! # Example
+//!
+//! ```
+//! use tower_http::normalize_path::NormalizePathLayer;
+//! use http::{Request, Response, StatusCode};
+//! use hyper::Body;
+//! use std::{iter::once, convert::Infallible};
+//! use tower::{ServiceBuilder, Service, ServiceExt};
+//!
+//! # #[tokio::main]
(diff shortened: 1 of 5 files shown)


Nehliin (approved):
nice!

tower-http/src/normalize_path.rs
```diff
@@ -0,0 +1,192 @@
...
+        .parse()
+        .unwrap();
+
+        Some(new_path_and_query)
+    } else {
+        None
+    };
+
+    parts.path_and_query = new_path_and_query;
+    *uri = Uri::from_parts(parts).unwrap();
```
davidpdrsn, line 134: This is also vulnerable to https://github.com/tokio-rs/axum/issues/1122. I'll fix that before merging.

jplatte: I'm still not really clear on how that could happen. How can the server receive URIs without a scheme?

davidpdrsn: Yeah that also confuses me but it happens https://discord.com/channels/500028886025895936/990823637084934144/991007953010188308. But in http maybe 🤷

davidpdrsn:
I'll fix CI later in a separate PR.

PR #278, opened 2022-07-02
Correctly identify gRPC requests in default `on_response` callback

### LINKED ISSUE #277: on_response trace uses HTTP status code instead of gRPC code for grpc-web
## Issue

 When using `TraceLayer::new_for_grpc()`, the on_response trace prints the http status code instead of the gRPC code. 

## Cause

When determining whether the response is gRPC, the content type header is expected to be strictly equal to "application/grpc":
https://github.com/tower-rs/tower-http/blob/bab200c34e66b27b0087467d659e5692a5843846/tower-http/src/trace/on_response.rs#L270

However, grpc-web uses "application/grpc-web+proto", "application/grpc-web-text+proto", and other variants. Even "normal" gRPC-over-HTTP2 can append "+proto" or "+json" to the content type, so those would be broken as well.

See: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2

## Solution

Check that the content type header starts with "application/grpc" rather than expecting strict equality.

jdahlq:
Fixes #277

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -21,7 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
-- None.
+- **trace:** Correctly identify gRPC requests in default `on_response` callback ([#278])
+
+[#278]: https://github.com/tower-rs/tower-http/pull/278
 
 # 0.3.4 (June 06, 2022)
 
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Thanks!

Wanna add a line to changelog as well?

jdahlq:
> Thanks!
> 
> Wanna add a line to changelog as well?

Sure thing, done!

davidpdrsn (approved):
Thanks!

PR #280, opened 2022-07-04
fix example in SetStaus middleware

yanns:
The `async fn handle` was duplicated.
from https://www.youtube.com/watch?v=QtKO9f_iPhI


yanns:
The `test-msrv` is failing, as you showed in the video 😉

davidpdrsn:
Hehe yeah I haven't taken the time to fix that yet.

Issue #276, opened 2022-06-27
Decompression service decompress the request


Antti:
## Feature Request

Add request decompression to the `Decompression` service. This would be used on the server-side to decompress client requests, in addition to the existing logic to decompress server responses.

### Motivation

In our company, we are using Axum for our internal APIs and posting large JSON payloads. While we are planning to move to some more space-efficient formats, it would help a lot if we could compress the client request body using standard compression methods (gzip, deflate, br).

### Proposal

Implement decompression for the requests based on the `Content-Encoding` header

### Alternatives

Possibly create a separate service/layer for this?

davidpdrsn:
This has been requested before but the question has always been how to handle unsupported encodings.

About `content-encoding` [the spec](https://httpwg.org/specs/rfc7231.html#header.content-encoding) says

> An origin server MAY respond with a status code of 415 (Unsupported Media Type) if a representation in the request message has a content coding that is not acceptable.

So I suppose we could do just do that.

Most of machinery is already there so this is probably not hard.

Antti:
@davidpdrsn I did manage to make it work for us with the following changes (although 1 example and tests do not compile with some type errors).
https://github.com/fetlife/tower-http/commit/f2843286d31c08b755c3c983b003fa736859c666

It doesn't handle the unsupported content-encoding though.

davidpdrsn:
I think we have to add a new middleware. Changing the trait bounds like that is a breaking change.

wanderinglethe:
I have created a [https://github.com/tower-rs/tower-http/compare/master...wanderinglethe:tower-http:request-decompression](branch) with Decompression middleware for request bodies.

It still needs documentation and refactoring/naming to coexist with the other Decompression middleware.

leio:
Given that #282 is merged and released, I think this issue can be closed as well?

davidpdrsn:
Yes!

PR #282, opened 2022-07-09
Request decompression

wanderinglethe:
## Motivation

Currently compression and decompression of response bodies are supported with the `Compression` and `Decompression` services. Issue #276 proposes to also implement decompression for request bodies, this PR addresses that. To complete the compression and decompression modules, request compression could be added in another PR.

## Solution

`RequestDecompression` decompresses the request body with the algorithm defined in the `Content-Encoding` header, it does not support multiple/stacked algorithms. When an algorithm is not supported it will return a `415 Unsupported Media Type` by default, but it can also be configured to pass-through the request to the inner service.

## Questions
This is my first PR and I am not sure if I am going in the right direction, I would already like some feedback while working on the issue.

### Refactor `decompression` module
Will be done in a different PR.

### Service and Future implementation
The service and future are based on the `Decompression` service and `ResponseFuture`, but I have the feeling the constraints are too complex. Can these be simplified?

And the same for the boxing of the bodies and errors, has this been done correctly?

### API
I have copied the layer implementation from the request decompression layer to make the API uniform. The extra functionality is the pass-through of unsupported encodings, maybe later this can be changed into a single final decompression, that return the 415, and a stacked decompression that pops an encoding and pass-through to the next service.

Would this be a good first way to go?

### General code style / conventions
Since I have the basic functionalities tested I can now refactor code. Are there any structures that can be written clearer or more concise? Or are there things that do not follow Rust or Tower conventions?

## Todos
- [x] Documentation
- [x] Non-breaking `decompression::Decompression`, including docs
- [x] Refactor code structure / style

#### Thank you
Thank you for looking into this.


davidpdrsn (commented):
I haven't read through the code yet and while I appreciate that you wanna refactor I would prefer if it was done in a separate PR. Makes it easier to review and bisect in case of bugs.

Is it possible for you to split things up?

wanderinglethe:
> Is it possible for you to split things up?

Where should I locate the request decompression module and what should I name it?

davidpdrsn:
I'd call it `RequestDecompression` and put it in the `decompression` module if possible

davidpdrsn:
I think we should also consider making separate middleware for (de)compressing requests and responses rather than making one middleware do both. It's easy to apply both if that's what users want.

wanderinglethe:
> I think we should also consider making separate middleware for (de)compressing requests and responses rather than making one middleware do both. 

That is what I have done, right?

davidpdrsn:
Yes sorry. It was mostly a note to myself. Haven't actually read your code yet 😅

wanderinglethe:
I think most of the PR is completed and can be reviewed. Although my questions, in the first message, still stand.
@davidpdrsn could you run the workflow?

Thank you

davidpdrsn:
Thanks!

I'm not sure when I'll get time to review this. Going on vacation in a few days. So might not get time before I'm back in a few weeks.

wanderinglethe:
> I'm not sure when I'll get time to review this. Going on vacation in a few days. So might not get time before I'm back in a few weeks.

Have a good vacation. I will see if I can work on some other PR or maybe I will take some vacation as well. :)

reneklacan:
@davidpdrsn any chance you are back from vacation? :pray: I'd also hugely appreciate getting this merged :heart:

davidpdrsn:
I am 😊 I'll try and have a look soon.

wanderinglethe:
Thank you

reneklacan:
@davidpdrsn Thanks! Hope you had a lovely vacation

Nehliin (approved):
LGTM!

wanderinglethe:
Do you have any critique on conventions, design or anything else?

> LGTM!

davidpdrsn:
I still wanna look through. Please remain patient.

wanderinglethe:
> I still wanna look through. Please remain patient.
Thank you. That's what I expected, but maybe @Nehliin has some critique as well

frxstrem:
Any progress on getting this merged? At my company we have a need to support request decompression in our Axum/Hyper based services, and it would be much better if we get support from that in `tower-http` rather than doing it in our own codebase.

mmd-osm:
Apologies if this question is asking the obvious. Would it be possible to specify the maximum permitted request body size after decompression in a lazy way somehow, ie. without uncompressing the whole payload, as to avoid excessive memory consumption?

frxstrem:
> Apologies if this question is asking the obvious. Would it be possible to specify the maximum permitted request body size after decompression in a lazy way somehow, ie. without uncompressing the whole payload, as to avoid excessive memory consumption?

@mmd-osm As far as I can tell, this is doing streaming decompression, so you should be able to apply a [`tower_http::limit::RequestBodyLimitLayer`](https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimitLayer.html) layer to limit the number of bytes in the request body. Applying that layer after the request decompression layer should apply the limit to the decompressed bytes.

tower-http/src/decompression/mod.rs
```diff
@@ -53,15 +97,22 @@
...
-};
+pub use self::body::DecompressionBody;
+
+pub use self::future::ResponseFuture;
+pub use self::layer::DecompressionLayer;
+pub use self::service::Decompression;
+
+pub use self::request::future::RequestDecompressionFuture;
+pub use self::request::layer::RequestDecompressionLayer;
+pub use self::request::service::RequestDecompression;
```
davidpdrsn, line 115: Why did you change these imports? Wanna revert the change? It makes things easier to review.

tower-http/src/decompression/request/future.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::pin::Pin;
+use std::task::Context;
+use std::task::Poll;
+
+pin_project! {
+    #[derive(Debug)]
+    /// Response future of [`RequestDecompression`]
+    ///
+    /// Can either poll its inner future or creates a `Response`
+    /// with an Unsupported Media Type status code.
```
davidpdrsn, line 17: I'm not sure we need to document this here as its an implementation detail of the future.

It should be documented on the service or layer instead.

```suggestion
```

tower-http/src/decompression/request/future.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+}
+
+impl<F, B, E> Future for RequestDecompressionFuture<F, B, E>
+where
+    F: Future<Output = Result<Response<B>, E>>,
+    B: Body + Send + 'static,
+    <B as Body>::Data: Buf + 'static,
+    <B as Body>::Error: Into<BoxError> + 'static,
```
davidpdrsn, line 74: This should work too.

```suggestion
    B::Data: Buf + 'static,
    B::Error: Into<BoxError> + 'static,
```

tower-http/src/decompression/request/future.rs
```diff
@@ -0,0 +1,100 @@
...
+                let res = Response::builder()
+                    .header(
+                        header::ACCEPT_ENCODING,
+                        accept
+                            .to_header_value()
+                            .unwrap_or(HeaderValue::from_static("identity")),
+                    )
+                    .status(StatusCode::UNSUPPORTED_MEDIA_TYPE)
+                    .body(Empty::new().map_err(Into::into).boxed_unsync())
+                    .map_err(Into::into);
```
davidpdrsn, line 95: Its fine to unwrap here since it'll never panic.

```suggestion
                    .unwrap();
```

tower-http/src/decompression/request/future.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl<F, B, E> Future for RequestDecompressionFuture<F, B, E>
+where
+    F: Future<Output = Result<Response<B>, E>>,
+    B: Body + Send + 'static,
+    <B as Body>::Data: Buf + 'static,
+    <B as Body>::Error: Into<BoxError> + 'static,
+    E: Into<BoxError>,
+{
+    type Output = Result<Response<UnsyncBoxBody<<B as Body>::Data, BoxError>>, BoxError>;
```
davidpdrsn, line 77:
```suggestion
    type Output = Result<Response<UnsyncBoxBody<B::Data, BoxError>>, BoxError>;
```

tower-http/src/decompression/request/layer.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl RequestDecompressionLayer {
+    /// Creates a new `RequestDecompressionLayer`.
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Sets whether to support
+    /// gzip encoding.
```
davidpdrsn, line 41: Some odd line breaks going on here and in the other methods as well.

tower-http/src/decompression/request/layer.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Disables support for Brotli encoding.
+    ///
+    /// This method is available even if the `br` crate feature is disabled.
+    pub fn no_br(mut self) -> Self {
+        self.accept.set_br(false);
+        self
+    }
+
+    /// Passes through the request even when the encoding is not supported.
+    pub fn pass_through_unaccepted(mut self) -> Self {
```
davidpdrsn, line 89: I think this method should take a `bool` to similar to the `set_*` methods.

Then users can easily configure things based on things like env vars:

```rust
layer.passthrough_unaccepted(std::env::var("FOO").is_ok())
```

I believe it is also spelled `passthrough` and not `pass through`

davidpdrsn (commented):
Sorry for taking forever to review. First round of feedback is in! Generally I think this looks good. Mostly minor nitpicks here and there.

davidpdrsn:
Oh and CI should be fixed if you merge the latest `master`.

tower-http/src/decompression/request/layer.rs
wanderinglethe: In my native language it would also be `passthrough`, but in English that is a noun and the verb is `to pass through`.

tower-http/src/decompression/request/layer.rs
wanderinglethe: Not sure what I did there.

tower-http/src/decompression/mod.rs
wanderinglethe: I wanted to separate the import of the `DecompressionBody` from the response imports. I will revert the change.

tower-http/src/decompression/request/future.rs
wanderinglethe: Ok, already documented on the layer

tower-http/src/decompression/request/future.rs
wanderinglethe: Fixed, I think I needed it when I used a more restrictive signature

wanderinglethe:
Thank you for reviewing David. I have made the changes you requested.

bruwozniak:
Would be amazing to see this merged :)

davidpdrsn (approved):
Thanks for working on this! Sorry for taking so long to review it.

PR #283, opened 2022-07-12
Make `ServeDir` infallible

davidpdrsn:
## Motivation

Previously `ServeDir`'s error type was `io::Error`. That means users using `ServeDir` with axum must use `.handle_error()` to convert any IO errors into responses:

```rust
Router::new().nest(
    "/assets",
    get_service(ServeDir::new(".")).handle_error(handle_error),
);

async fn handle_error(_err: io::Error) -> impl IntoResponse {
    (StatusCode::INTERNAL_SERVER_ERROR, "Something went wrong...")
}
```

Since `ServeDir` already maps `io::ErrorKind::NotFound` and `io::ErrorKind::PermissionDenied` there isn't much you do for error besides returning `StatusCode::INTERNAL_SERVER_ERROR`.

That is tedious to do and so I think we should make `ServeDir` handle the error internally and change its error type to `Infallible`. That would make the axum integration really simple:

```rust
Router::new().nest("/", ServeDir::new("."));
```

## Solution

- Change `<ServeDir as Service<_>>::Error` to `Infallible`
- Convert any errors that happen to `StatusCode::INTERNAL_SERVER_ERROR` and log the error
- Add `ServeDir::try_call` which returns a future that maintains the error. That way users can still customize the error if they want.

The same was done for `ServeFile` as its just a wrapper around `ServeDir`.

Note that this is a breaking change. We shipped the last major 3 months ago so probably fine to do another one but we should consider if there are any other breaking changes we wanna make.


tower-http/src/macros.rs
```diff
@@ -73,3 +73,41 @@ macro_rules! opaque_body {
...
+            $actual: std::future::Future,
+        {
+            type Output = <$actual as std::future::Future>::Output;
+            #[inline]
+            fn poll(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> std::task::Poll<Self::Output> {
+                self.project().inner.poll(cx)
+            }
+        }
+    }
+}
```
davidpdrsn, line 113: This macro was copied from tower https://github.com/tower-rs/tower/blob/master/tower/src/macros.rs. We also use it a bunch in axum.

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -350,6 +401,57 @@ where
...
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let future = self.try_call(req).map(|result| -> Result<_, Infallible> {
+            match result {
+                Ok(response) => Ok(response),
+                Err(err) => {
+                    tracing::error!(error = %err, "Failed to read file");
+
+                    let body =
+                        ResponseBody::new(Empty::new().map_err(|err| match err {}).boxed_unsync());
```
Nehliin, line 432: Is it necessary to explicitly create the empty body here? Is it not possible to just omit the body field in the builder or do we get type system complaints then?

Nehliin (approved):
Nice! this should make things more consistent

tower-http/src/services/fs/serve_dir/mod.rs
davidpdrsn: It's required 😕

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -350,6 +401,57 @@ where
...
+        if let Some(fallback) = &mut self.fallback {
+            fallback.poll_ready(cx)
+        } else {
+            Poll::Ready(Ok(()))
+        }
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let future = self.try_call(req).map(|result| -> Result<_, Infallible> {
+            match result {
```
jplatte, line 426: Since we're not using any control flow inside the branches here, this `match` could be an `.unwrap_or_else()`.

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -350,6 +401,57 @@ where
...
+                    let body =
+                        ResponseBody::new(Empty::new().map_err(|err| match err {}).boxed_unsync());
+                    let response = Response::builder()
+                        .status(StatusCode::INTERNAL_SERVER_ERROR)
+                        .body(body)
+                        .unwrap();
+                    Ok(response)
+                }
+            }
+        } as _);
```
jplatte, line 440: What does this cast do?

davidpdrsn: It makes rust infer the right type for the closure. If you don't have it you get this error

```
error[E0308]: mismatched types
   --> tower-http/src/services/fs/serve_dir/mod.rs:439:39
    |
425 |         let future = self.try_call(req).map(|result: Result<_, _>| -> Result<_, Infallible> {
    |                                             ----------------------------------------------- the found closure
...
439 |         InfallibleResponseFuture::new(future)
    |         ----------------------------- ^^^^^^ expected fn pointer, found closure
    |         |
    |         arguments to this function are incorrect
    |
    = note: expected struct `futures_util::future::Map<_, fn(Result<http::Response<serve_dir::ResponseBody>, std::io::Error>) -> Result<http::Response<serve_dir::ResponseBody>, Infallible>>`
               found struct `futures_util::future::Map<_, [closure@tower-http/src/services/fs/serve_dir/mod.rs:425:45: 425:92]>
```

jplatte (approved):

Issue #284, opened 2022-07-15
cors_multiple_origins


Paullahner:
## Bug Report
 
https://github.com/Paullahner/cors_multiple_origins

using an array of origins like the function: _start_with_multiple_origins_ 
will cause the Header to have no _access-control-allow-origin_ field
But using the unction _start_with_single_origin_ the access-control-allow-origin: field exists and works as intended

davidpdrsn:
The issue seems to be that `"*"` isn't matched correctly when providing a list of origins [here](https://github.com/tower-rs/tower-http/blob/master/tower-http/src/cors/allow_origin.rs#L87). If you do this it does work correctly:

```rust
let origins = [
    "http://localhost:8080".parse().unwrap(),
    "http://localhost:3000".parse().unwrap(),
];

let app = Router::new()
    .route("/json", get(json))
    .layer(
        CorsLayer::new()
            .allow_origin(origins)
            .allow_methods([Method::GET]),
    );
```

I'll make a PR for this.

Thanks for noticing!

davidpdrsn:
I just filed https://github.com/tower-rs/tower-http/pull/285. Are you able to test it?

Paullahner:
Im sorry but on my end it doesn't matter if I use 

```
let origins = [
        "http://localhost:3000".parse().unwrap(),
        "http://localhost:8080".parse().unwrap(),
    ];
```
or 

```
    let origins = [
        "*".parse().unwrap(),
        "http://localhost:8080".parse().unwrap(),
    ];
```
in neither case can I see an _access-control-allow-origin_

davidpdrsn:
Uhm thats weird 🤔 I tested it on axum's [cors example](https://github.com/tokio-rs/axum/blob/main/examples/cors/src/main.rs) and it works there.

Paullahner:
I can confirm with the example it works! 
But why does it not work with with the code in my repository?

jplatte:
The access-control-allow-origin header can only have a single origin as its value. When you provide a list to the CORS middleware, which one is returned depends on the origin header of the request. If the origin of the request is in the lists, it is mirrored back. If not, no allow-origin header is sent.

PR #285, opened 2022-07-15
Panic if `*` is used with `AllowOrigin::list`

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/284

Otherwise https://github.com/tower-rs/tower-http/blob/master/tower-http/src/cors/allow_origin.rs#L87 would compare the origin with `"*"`.


jplatte:
I wonder whether we should instead panic. I don't see why you would legitimately provide both a wildcard which allows any origin, and other values.

jplatte (approved):

PR #290, opened 2022-08-22
Implement require_auth with validate_request

82marbag:
## Motivation

This is a follow-up to: https://github.com/tower-rs/tower-http/pull/289
`ValidateRequest` is an abstraction of a validator and `RequireAuthorization` is an instance, but the latter has yet to be implemented in terms of the former. They now are two independent types.

## Solution

`RequireAuthorization` is now implemented by `ValidateRequest`.

This is a breaking change.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Removed
 
-- None.
+- Removed `RequireAuthorization` in favor of `ValidateRequest`
 
 ## Fixed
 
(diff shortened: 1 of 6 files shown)


82marbag:
Next could be the async version

82marbag:
test-msrv fails on: Error: could not compile `hashbrown`

82marbag:
Any update on this?

davidpdrsn (commented):
Sorry the delay!

I think this looks great. Just a couple of nit picks.

If you merge latest `master` then CI should be fixed.

davidpdrsn (approved):
Thanks!

I'm currently combing through PRs and issues to figure what goes into the next breaking release but this'll be part of that for sure.

PR #303, opened 2022-10-11
Timeout on body

### LINKED ISSUE #295: Add BodyTimeout middleware
### Motivation

Besides the larger service timeouts, it's also common for people to want to add a timeout to the reading of chunks from a body, either for the request as a server, or for the response of a client.

### Proposal

- Make the `timeout` module a folder, with service and body timeout files.
- The body timeout should have a `struct TimeoutBody<B>`.
  - Implement `Body for TimeoutBody<B> where B: Body`, such that polling for data will check the inner body poll, and also a timeout.    
    - The timeout should be reset if a chunk can be returned.
- `RequestBodyTimeout` and `ResponseBodyTimeout` layers can be added (`Service` and `Layer`) which essentially do what `MapRequest` and `MapResponse` do, wrapping the `Request<B>` into `Request<TimeoutBody<B>>` (and same with response).

82marbag:
## Motivation

Explained in #295. This PR closes #295.

## Solution

Wrap the body in a `TimeoutBody`. `TimeoutBody` will poll a sleep future to check whether the body is inactive and register itself to be awoken. The sleep future is polled and checked right after creation to avoid a potential delay in execution making the executor to never poll the sleep future again. That is, if between creation and `poll` on sleep the time runs out and the sleep is done, a timeout error is immediately returned

### DIFF
--- a/tower-http/src/timeout/body.rs
+++ b/tower-http/src/timeout/body.rs
@@ -319,7 +319,7 @@ mod tests {
 
     impl Error for MockError {}
     impl Display for MockError {
-        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
             todo!()
         }
     }
--- a/tower-http/src/timeout/body.rs
+++ b/tower-http/src/timeout/body.rs
@@ -3,9 +3,9 @@
 //! Bodies must produce data at most within the specified timeout.
 //! If the body does not produce a requested data frame within the timeout period, it will return an error.
 //!
-//! # Differences from [`tower_http::timeout::service::Timeout`]
+//! # Differences from [`tower_http::timeout::Timeout`]
 //!
-//! [`tower_http::timeout::service::Timeout`] applies a timeout to the request future, not body.
+//! [`tower_http::timeout::Timeout`] applies a timeout to the request future, not body.
 //! That timeout is not reset when bytes are handled, whether the request is active or not.
 //! Bodies are handled asynchronously outside of the tower stack's future and thus needs an additional timeout.
 //!
@@ -37,7 +37,6 @@
 
 use crate::BoxError;
 use futures_core::{ready, Future};
-use http::{Request, Response};
 use http_body::Body;
 use pin_project_lite::pin_project;
 use std::{
@@ -46,8 +45,6 @@ use std::{
     time::Duration,
 };
 use tokio::time::{sleep, Sleep};
-use tower_layer::Layer;
-use tower_service::Service;
 
 pin_project! {
     /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration.
@@ -149,164 +146,6 @@ impl std::fmt::Display for TimeoutError {
         write!(f, "data was not received within the designated timeout")
     }
 }
-
-/// Applies a [`TimeoutBody`] to the request body.
-#[derive(Clone, Debug)]
-pub struct RequestBodyTimeoutLayer {
-    timeout: Duration,
-}
-
-impl RequestBodyTimeoutLayer {
-    /// Creates a new [`RequestBodyTimeoutLayer`].
-    pub fn new(timeout: Duration) -> Self {
-        Self { timeout }
-    }
-}
-
-impl<S> Layer<S> for RequestBodyTimeoutLayer {
-    type Service = RequestBodyTimeout<S>;
-
-    fn layer(&self, inner: S) -> Self::Service {
-        RequestBodyTimeout::new(inner, self.timeout)
-    }
-}
-
-/// Applies a [`TimeoutBody`] to the request body.
-#[derive(Clone, Debug)]
-pub struct RequestBodyTimeout<S> {
-    inner: S,
-    timeout: Duration,
-}
-
-impl<S> RequestBodyTimeout<S> {
-    /// Creates a new [`RequestBodyTimeout`].
-    pub fn new(service: S, timeout: Duration) -> Self {
-        Self {
-            inner: service,
-            timeout,
-        }
-    }
-
-    /// Returns a new [`Layer`] that wraps services with a [`RequestBodyTimeoutLayer`] middleware.
-    ///
-    /// [`Layer`]: tower_layer::Layer
-    pub fn layer(timeout: Duration) -> RequestBodyTimeoutLayer {
-        RequestBodyTimeoutLayer::new(timeout)
-    }
-
-    define_inner_service_accessors!();
-}
-
-impl<S, ReqBody> Service<Request<ReqBody>> for RequestBodyTimeout<S>
-where
-    S: Service<Request<TimeoutBody<ReqBody>>>,
-    S::Error: Into<Box<dyn std::error::Error>>,
-{
-    type Response = S::Response;
-    type Error = S::Error;
-    type Future = S::Future;
-
-    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.inner.poll_ready(cx)
-    }
-
-    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
-        let req = req.map(|body| TimeoutBody::new(self.timeout, body));
-        self.inner.call(req)
-    }
-}
-
-/// Applies a [`TimeoutBody`] to the response body.
-#[derive(Clone)]
-pub struct ResponseBodyTimeoutLayer {
-    timeout: Duration,
-}
-
-impl ResponseBodyTimeoutLayer {
-    /// Creates a new [`ResponseBodyTimeoutLayer`].
-    pub fn new(timeout: Duration) -> Self {
-        Self { timeout }
-    }
-}
-
-impl<S> Layer<S> for ResponseBodyTimeoutLayer {
-    type Service = ResponseBodyTimeout<S>;
-
-    fn layer(&self, inner: S) -> Self::Service {
-        ResponseBodyTimeout::new(inner, self.timeout)
-    }
-}
-
-/// Applies a [`TimeoutBody`] to the response body.
-#[derive(Clone)]
-pub struct ResponseBodyTimeout<S> {
-    inner: S,
-    timeout: Duration,
-}
-
-impl<S> ResponseBodyTimeout<S> {
-    /// Creates a new [`ResponseBodyTimeout`].
-    pub fn new(service: S, timeout: Duration) -> Self {
-        Self {
-            inner: service,
-            timeout,
-        }
-    }
-
-    /// Returns a new [`Layer`] that wraps services with a [`ResponseBodyTimeoutLayer`] middleware.
-    ///
-    /// [`Layer`]: tower_layer::Layer
-    pub fn layer(timeout: Duration) -> ResponseBodyTimeoutLayer {
-        ResponseBodyTimeoutLayer::new(timeout)
-    }
-
-    define_inner_service_accessors!();
-}
-
-impl<S, ReqBody, ResBody> Service<Request<ReqBody>> for ResponseBodyTimeout<S>
-where
-    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
-    S::Error: Into<Box<dyn std::error::Error>>,
-{
-    type Response = Response<TimeoutBody<ResBody>>;
-    type Error = S::Error;
-    type Future = ResponseBodyTimeoutFuture<S::Future>;
-
-    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.inner.poll_ready(cx)
-    }
-
-    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
-        ResponseBodyTimeoutFuture {
-            inner: self.inner.call(req),
-            timeout: self.timeout,
-        }
-    }
-}
-
-pin_project! {
-    /// Response future for [`ResponseBodyTimeout`].
-    pub struct ResponseBodyTimeoutFuture<Fut> {
-        #[pin]
-        inner: Fut,
-        timeout: Duration,
-    }
-}
-
-impl<Fut, ResBody, E> Future for ResponseBodyTimeoutFuture<Fut>
-where
-    Fut: Future<Output = Result<Response<ResBody>, E>>,
-{
-    type Output = Result<Response<TimeoutBody<ResBody>>, E>;
-
-    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let timeout = self.timeout;
-        let this = self.project();
-        let res = ready!(this.inner.poll(cx))?;
-        Poll::Ready(Ok(res.map(|body| TimeoutBody::new(timeout, body))))
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/tower-http/src/timeout/body.rs
+++ b/tower-http/src/timeout/body.rs
@@ -102,7 +102,7 @@ where
                 if let Poll::Ready(_) = this.sleep.as_pin_mut().unwrap().poll(cx) {
                     return Poll::Ready(Some(Err(Box::new(TimeoutError))))
                 }
-                Poll::Ready(data.transpose().map_err(|_| Box::new(TimeoutError) as Self::Error).transpose())
+                Poll::Ready(data.transpose().map_err(|_| Box::new(TimeoutError).into()).transpose())
             }
             Poll::Pending => Poll::Pending
         }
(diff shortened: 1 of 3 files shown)


82marbag:
cc @LucioFranco @seanmonstar @hlbarber

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
...
+
+impl std::error::Error for TimeoutError {}
+
+impl std::fmt::Display for TimeoutError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "TimeoutError")
+    }
+}
+
+/// Applies a TimeoutBody to the request body.
```
hlbarber, line 139: [`TimeoutBody`]

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
...
+impl<S> Layer<S> for RequestBodyTimeoutLayer
+{
+    type Service = RequestBodyTimeout<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        RequestBodyTimeout::new(inner, self.timeout)
+    }
+}
+
+/// Applies a TimeoutBody to the request body.
```
hlbarber, line 161: [`TimeoutBody`]

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
...
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let req = req.map(|body| TimeoutBody::new(self.timeout, body));
+        self.inner.call(req)
+    }
+}
+
+/// Applies a TimeoutBody to the response body.
```
hlbarber, line 203: [`TimeoutBody`]

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
...
+            Poll::Ready(()) => return Poll::Ready(Err(Box::new(TimeoutError))),
+        }
+
+        this.body.poll_trailers(cx).map_err(|_| Box::new(TimeoutError).into())
+    }
+}
+
+/// Error for [`TimeoutBody`].
+#[derive(Debug)]
+pub struct TimeoutError;
```
hlbarber, line 129: If this is going to be erased then maybe we should not `pub` it?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
+//! Middleware that applies a timeout to bodies.
+//!
+//! Bodies must produce data at most within the specified timeout.
+//! If they are inactive, an error will be generated.
+//!
+//! # Differences from [`tower_http::timeout::service::Timeout`]
```
hlbarber, line 6: This section is a nice touch.

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        // Check for body data.
+        match this.body.poll_data(cx) {
+            Poll::Ready(data) => {
+                // Some data is ready. Reset the `Sleep`...
+                this.sleep.set(Some(sleep(*this.timeout)));
+
+                // ...then `poll` it to get awoken.
+                if let Poll::Ready(_) = this.sleep.as_pin_mut().unwrap().poll(cx) {
```
hlbarber, line 102: Do we dare `.unwrap_unchecked()`?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,347 @@
...
+use futures_core::Future;
+use pin_project_lite::pin_project;
+use tokio::time::{Sleep, sleep};
+use http::{Request, Response};
+use tower_layer::Layer;
+use tower_service::Service;
+
+pin_project! {
+    /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration.
+    pub struct TimeoutBody<B> {
```
hlbarber, line 48: Taking a big step back, this _could_ live in `http_body` crate if there was a standard interface for sleep futures.

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+        }
+
+        // Check for body data.
+        match this.body.poll_data(cx) {
+            Poll::Ready(data) => {
+                // Some data is ready. Reset the `Sleep`...
+                this.sleep.set(Some(sleep(*this.timeout)));
+
+                // ...then `poll` it to get awoken.
+                if let Poll::Ready(_) = this.sleep.as_pin_mut().unwrap().poll(cx) {
```
hlbarber, line 106: Adding some more detail.

We register our waker with the reactor backing `Sleep`, this provides us with the `.wake()` when the timeout has elapsed.

Suppose we don't `if let Poll::Ready(_)` then `return Err` here, then in the unlikely but possible(?) situation where it did return `Poll::Ready` then it should not call `.wake()` again. As a result we stay waiting for the `.wake()` from the `this.body.poll_data(cx)` - which is not the desired behavior.

tower-http/src/timeout/body.rs
hlbarber: Or....maybe we should keep it `pub` so people can check for it via downcasting?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        let mut this = self.project();
+
+        // Start the `Sleep` if not active.
+        let sleep_pinned = if let Some(some) = this.sleep.as_mut().as_pin_mut() {
+            some
+        } else {
+            this.sleep.set(Some(sleep(*this.timeout)));
+            this.sleep.as_mut().as_pin_mut().unwrap()
```
LucioFranco, line 90: I don't understand this line? If we set it above I don't think we need this check?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+        };
+
+        // Error if the timeout has expired.
+        match sleep_pinned.poll(cx) {
+            Poll::Pending => (),
+            Poll::Ready(()) => return Poll::Ready(Some(Err(Box::new(TimeoutError)))),
+        }
```
LucioFranco, line 97: Would read better as an `if let Poll::Ready(())`

tower-http/src/timeout/body.rs
LucioFranco, line 106: Why do we set the sleep above then repoll it again? If we have data available we should return that right away and then on the next poll check if there is a timeout. That is if this timeout is a data read frame timeout aka we timeout if we've been pending for a data frame for too long which I think this is what you are trying to achieve.

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+                // Some data is ready. Reset the `Sleep`...
+                this.sleep.set(Some(sleep(*this.timeout)));
+
+                // ...then `poll` it to get awoken.
+                if let Poll::Ready(_) = this.sleep.as_pin_mut().unwrap().poll(cx) {
+                    return Poll::Ready(Some(Err(Box::new(TimeoutError))));
+                }
+                Poll::Ready(
+                    data.transpose()
+                        .map_err(|_| Box::new(TimeoutError).into())
```
LucioFranco, line 111: What happens to the error from the inner body poll_data? Seems like we just throw it away here? In this case, I think we can just have a bound that is `B::Error: Into<BoxError>` and then the map here becomes `map_err(Into::into)`.

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+                        .transpose(),
+                )
+            }
+            Poll::Pending => Poll::Pending,
```
LucioFranco, line 115: Might want to consider using `futures_core::ready!` in the above poll so that the match becomes a match over `Option<Result<...>>` instead of `Poll<Option<Result<...>>>`. Its basically the try macro but for poll fn. https://docs.rs/futures-core/latest/futures_core/macro.ready.html

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+        }
+    }
+
+    fn poll_trailers(
```
LucioFranco, line 119: Same comments for the above apply here

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+
+        this.body
+            .poll_trailers(cx)
+            .map_err(|_| Box::new(TimeoutError).into())
+    }
+}
+
+/// Error for [`TimeoutBody`].
+#[derive(Debug)]
+struct TimeoutError;
```
LucioFranco, line 144: This struct should be pub with a `(())` type inside so it can't be constructed outside the crate. We should look at how other middleware in this crate handle errors.

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub fn layer(timeout: Duration) -> RequestBodyTimeoutLayer {
+        RequestBodyTimeoutLayer::new(timeout)
+    }
+
+    define_inner_service_accessors!();
+}
+
+impl<S, ReqBody, ResBody> Service<Request<ReqBody>> for RequestBodyTimeout<S>
+where
+    S: Service<Request<TimeoutBody<ReqBody>>, Response = Response<ResBody>>,
```
LucioFranco, line 203: Why do you need the `Response = ...` part, I think this can be omitted?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub fn layer(timeout: Duration) -> ResponseBodyTimeoutLayer {
+        ResponseBodyTimeoutLayer::new(timeout)
+    }
+
+    define_inner_service_accessors!();
+}
+
+impl<S, ReqBody, ResBody> Service<Request<ReqBody>> for ResponseBodyTimeout<S>
+where
+    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
```
LucioFranco, line 269: Same here I think we can omit the response part?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+pin_project! {
+    /// Response future for [`ResponseBodyTimeout`].
+    pub struct ResponseBodyTimeoutFuture<Fut> {
+        #[pin]
+        inner: Fut,
+        timeout: Duration,
+    }
+}
+
+use futures_core::ready;
```
LucioFranco, line 297: imports should be at the top of the file

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,373 @@
...
+impl<Fut, ResBody, E> Future for ResponseBodyTimeoutFuture<Fut>
+where
+    Fut: Future<Output = Result<Response<ResBody>, E>>,
+{
+    type Output = Result<Response<TimeoutBody<ResBody>>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let timeout = self.timeout;
+        let this = self.project();
+        let res = ready!(this.inner.poll(cx)?);
```
LucioFranco, line 307: This try operator is confusing to me and how it interacts with ready. I would imagine it would go on the outside?

LucioFranco (changes requested):
This is a great start! Left a few comments/questions, lmk if you want to discuss any of it offline.

tower-http/src/timeout/body.rs
hlbarber: The rationale behind `poll`ing it as soon as it's set is that we want to register `cx` with the sleep reactor so that when the timeout is complete this task is woken. The fear is: if we don't `poll` here this task could sit indefinitely waiting for the data to `.wake()` up the task.

It seems that, if we don't care about the sleep reactor calling `.wake()` then we might as well use `Instant::now()`?

LucioFranco: Well I guess what is the timeout for? Is it for each data frame? So lets say I pull one data frame out then does the timer start then or when I want to pull the next one out. If I want to do something in between data frames this breaks that logic a bit, no?

hlbarber (approved):

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,369 @@
...
+    use bytes::Bytes;
+    use pin_project_lite::pin_project;
+    use std::{error::Error, fmt::Display};
+
+    #[derive(Debug)]
+    struct MockError;
+
+    impl Error for MockError {}
+    impl Display for MockError {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
LucioFranco, line 312: There is a warning here we should resolve before merge. Idk why CI isn't failing for it.

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,369 @@
...
+use tokio::time::{sleep, Sleep};
+use tower_layer::Layer;
+use tower_service::Service;
+
+pin_project! {
+    /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration.
+    pub struct TimeoutBody<B> {
+        timeout: Duration,
+        #[pin]
+        sleep: Option<Sleep>,
```
LucioFranco, line 56: So I think there is a small issue here in that body 0.4 doesn't define what happens if when poll_data returns Pending that if you should be able to poll_trailers. So in that case, this single timeout could fire early in some awkward situations where someone is polling both data and trailers. So the solution here would be to have two sleep states in the body wrapper. I think we need to do this now for the current release but I think we should also add a note about how in `http-body` 1.0 these two poll_fn will be merged into a single `poll_frame` and that we can go back to a single sleep impl. So a comment about that would be helpful.

LucioFranco (changes requested):

tower-http/src/timeout/body.rs
82marbag: Added, thanks!

tower-http/src/timeout/body.rs
82marbag: CI seems to be green, which warning?

LucioFranco: image

when I look at the review page I see that

82marbag: Oh ok. Updated

tower-http/src/timeout/body.rs
LucioFranco: What happens when you do this?

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,379 @@
+//! Middleware that applies a timeout to bodies.
```
LucioFranco, line 1:
```suggestion
//! Middleware that applies a timeout to request and response bodies.
```

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,379 @@
...
+//!
+//! Bodies must produce data at most within the specified timeout.
+//! If they are inactive, an error will be generated.
```
LucioFranco, line 4: Not sure what this means, how about 
```suggestion
//! If the body does not produce a requested data frame within the timeout period it will return an error.
```

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,379 @@
...
+//!
+//! # Differences from `tower_http::timeout::service::Timeout`
```
LucioFranco, line 6: We should use inline links
```suggestion
//! # Differences from [`tower_http::timeout::service::Timeout`]
```

tower-http/src/timeout/body.rs
```diff
@@ -0,0 +1,379 @@
...
+//!
+//! `tower_http::timeout::service::Timeout` applies a timeout on the full request.
```
LucioFranco, line 8:
```suggestion
//! [`tower_http::timeout::service::Timeout`] applies a timeout on the full request.
```

LucioFranco, line 8: I would maybe say something like it applies the timeout to the request future and not the body. Bodies are handled asynchronously outside of the tower stack's future and thus needs an additional timeout.

LucioFranco (changes requested):

tower-http/src/timeout/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+mod service;
+
+pub use body::TimeoutBody;
+pub use body::TimeoutError;
+pub use service::RequestBodyTimeout;
+pub use service::RequestBodyTimeoutLayer;
+pub use service::ResponseBodyTimeout;
+pub use service::ResponseBodyTimeoutLayer;
+pub use service::Timeout;
+pub use service::TimeoutLayer;
```
LucioFranco, line 13: These can be combined into two use statements

LucioFranco (approved):
LGTM just one last thing and we merge

LucioFranco (approved):
SHIP IT

PR #313, opened 2022-12-02
Support configurable backend for `ServeDir`

### LINKED ISSUE #309: Support configurable backend for ServeDir
We've heard from several users who want to use `ServeDir` (or `ServeFile`) with files that are embedded in the binary, perhaps using [rust-embed](https://crates.io/crates/rust-embed). By using `ServeDir` they'd get a lot for free such as finding the file from the path, range requests, pre-compressed files, etc.

That isn't possible today since `ServeDir` is hardcoded to always fetch files from the filesystem. I've been thinking that it could be solved by making a "backend" trait that tells `ServeDir` how to open files and get metadata. Something like:

```rust
#[async_trait]
pub trait Backend {
    type File: AsyncRead + AsyncSeek;
    type Metadata: FileMetadata;

    async fn open_file(&self) -> io::Result<Self::File>;

    async fn metadata(&self, path: Path) -> io::Result<Self::Metadata>;
}

// we need a trait here because users cannot construct `std::fs::Metadata`
pub trait Metadata {
    // whatever methods we need
}
```

Then tower-http could provide a default backend that uses the file system but allow others to write their own.

davidpdrsn:
Still work-in-progress. I wanna actually implement a backend using [rust-embed](https://crates.io/crates/rust-embed) to make sure it works and there are some TODOs with unanswered questions in the code.

Fixes https://github.com/tower-rs/tower-http/issues/309

### DIFF
--- a/test-files/subdir/foo.txt
+++ b/test-files/subdir/foo.txt
@@ -1,0 +1,1 @@
+Hi from foo.txt
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -50,6 +50,7 @@ tower = { version = "0.4.10", features = ["buffer", "util", "retry", "make", "ti
 tracing-subscriber = "0.3"
 uuid = { version = "1.0", features = ["v4"] }
 serde_json = "1.0"
+rust-embed = "6.4"
 
 [features]
 default = []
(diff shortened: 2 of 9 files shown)


kolektiv:
This will be hugely useful when it's done, is there any movement planned? (I appreciate people have actual lives, just curiosity!) I'd love to see a "virtual filesystem" set of traits actually move to a more core area of Rust (maybe not `std` but... it would be nice if `std::fs` had a set of traits there)

jlizen:
Superseded by #684

PR #315, opened 2022-12-12
Add support for private network preflights

### LINKED ISSUE #314: Support for Access-Control-Allow-Private-Network
### Motivation

Chrome is removing/has removed the ability to send requests to private network endpoints from public websites:
https://developer.chrome.com/blog/private-network-access-preflight/

There is a pre-flight header it sends now (`Access-Control-Request-Private-Network`) and expects in the response to allow those requests through.

### Proposal

Perhaps a new method alongside `allow_headers`/`allow_methods`/etc on `CorsLayer` (`allow_private_network_access`?) that provides the correct response header for the preflight request?

jplatte:
Fixes #314.

Needs a test, but I don't want to put in more time. Would be happy for somebody else to take over.

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -17,7 +17,7 @@ bitflags = "1.3.1"
 bytes = "1"
 futures-core = "0.3"
 futures-util = { version = "0.3.14", default_features = false, features = [] }
-http = "0.2.2"
+http = "0.2.7"
 http-body = "0.4.5"
 pin-project-lite = "0.2.7"
 tower-layer = "0.3"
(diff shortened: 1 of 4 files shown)


Nehliin (changes requested):
Some unit tests would be nice for this as well just as a sanity check

jplatte:
I don't really want to dedicate more time to this beyond fixing trivial things. Should we add help-wanted somewhere so somebody else can take this over?

Nehliin:
Fair enough, sure that could make sense

jplatte:
Done in #373.

PR #317, opened 2022-12-24
Don't include identity in Content-Encoding header

Bobo1239:
According to the [spec][1] `identity` should only be used for the Accept-Encoding header and not be included in Content-Encoding.

In practice this change fixes chunked video loading in Firefox when using `ServeDir`. (making seeking possible without downloading the whole video file)

[1]: https://httpwg.org/specs/rfc9110.html#field.content-encoding

### DIFF
--- a/tower-http/src/services/fs/serve_dir/future.rs
+++ b/tower-http/src/services/fs/serve_dir/future.rs
@@ -2,7 +2,7 @@ use super::{
     open_file::{FileOpened, FileRequestExtent, OpenFileOutput},
     DefaultServeDirFallback, ResponseBody,
 };
-use crate::{services::fs::AsyncReadBody, BoxError, content_encoding::Encoding};
+use crate::{content_encoding::Encoding, services::fs::AsyncReadBody, BoxError};
 use bytes::Bytes;
 use futures_util::{
     future::{BoxFuture, FutureExt, TryFutureExt},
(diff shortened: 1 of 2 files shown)


tower-http/src/services/fs/serve_dir/future.rs
```diff
@@ -215,7 +215,10 @@ fn build_response(output: FileOpened) -> Response<ResponseBody> {
         .header(header::ACCEPT_RANGES, "bytes");
 
     if let Some(encoding) = output.maybe_encoding {
```
Nehliin, line 217: (You will need to add `PartialEq` as well to the enum for this to work)
```suggestion
if let Some(encoding) = output
        .maybe_encoding
        .filter(|encoding| encoding != Encoding::Identity)
```

Nehliin (commented):
Thanks for the PR! Do you have an example when this actually is set as an header value? I updated some of our tests to check if we return the encoding when serving uncompressed files and it doesn't look like we set the header value in the basic scenarios.

Bobo1239:
Thanks for the review. I've added a test which shows would've returned `Content-Encoding: identity` previously.

Also for posterity's sake: The original problem (video seeking) can be reproduced by using the simplest `ServeDir` example (see below) and navigating to a large video file in Firefox. In my case this was a 1.2GB .webm video. Without this PR seeking inside the video had a noticeable delay as the whole video would be downloaded and no range requests would be used. (not sure whether Firefox should do better here or not)
<details>
<summary>Code</summary>

```rust
use std::net::{Ipv4Addr, SocketAddr};

use axum::Server;
use tower::make::Shared;

#[tokio::main]
async fn main() {
    let addr = SocketAddr::from((Ipv4Addr::UNSPECIFIED, 3000));

    let service = ServeDir::new(".");
    Server::bind(&addr)
        .serve(Shared::new(service))
        .await
        .expect("server error");
}
```
</details>

Nehliin (approved):
Thanks!

Issue #319, opened 2023-01-08
CompressionLayer add zstd support


beckend:
### Motivation
It's new, but it's starting to pick up.
Here is a curl request:
```shell
curl --compressed http://whatever.com/my/endpoint
> GET /my/endpoint HTTP/1.1
> Host: whatever.com
> User-Agent: curl/7.87.0
> Accept: */*
> Accept-Encoding: deflate, gzip, br, zstd

```

It's also available in envoy proxy serving such requests.
https://github.com/envoyproxy/envoy/tree/main/examples/zstd

davidpdrsn:
That should be quite straightforward seeing as [async-compression already supports it](https://docs.rs/async-compression/latest/async_compression/#compression-algorithm).

PR #322, opened 2023-01-20
Support zstd (de)compression

davidpdrsn:
Fixes https://github.com/tower-rs/tower-http/issues/319


tower-http/src/compression_utils.rs
```diff
@@ -20,20 +20,29 @@ pub(crate) struct AcceptEncoding {
...
+            (false, false, true, false) => "br",
+            (true, true, true, true) => "gzip,deflate,br,zstd",
+            (true, true, false, true) => "gzip,deflate,zstd",
+            (true, false, true, true) => "gzip,br,zstd",
+            (true, false, false, true) => "gzip,zstd",
+            (false, true, true, true) => "deflate,br,zstd",
+            (false, true, false, true) => "deflate,zstd",
+            (false, false, true, true) => "br,zstd",
+            (false, false, false, true) => "zstd",
+            (false, false, false, false) => return None,
```
davidpdrsn, line 45: Not really sure what the priorities should be here.

tower-http/src/content_encoding.rs
```diff
@@ -26,6 +29,8 @@ impl Encoding {
             Encoding::Deflate => "deflate",
             #[cfg(any(feature = "fs", feature = "compression-br"))]
             Encoding::Brotli => "br",
+            #[cfg(any(feature = "fs", feature = "compression-zstd"))]
+            Encoding::Zstd => "br",
```
Nehliin, line 33: Copy paste error :)

tower-http/src/compression_utils.rs
jplatte: zstd is great, zstd first x)

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -180,6 +180,23 @@ impl<F> ServeDir<F> {
         self
     }
 
+    /// Informs the service that it should also look for a precompressed zstd
+    /// version of _any_ file in the directory.
+    ///
+    /// Assuming the `dir` directory is being served and `dir/foo.txt` is requested,
+    /// a client with an `Accept-Encoding` header that allows the zstd encoding
+    /// will receive the file `dir/foo.txt.zz` instead of `dir/foo.txt`.
```
Nehliin, line 188: .zstd here instead

Nehliin (changes requested):
Nice, just found some small copy paste errors

tower-http/src/compression_utils.rs
Nehliin: Br is supposedly pretty good as well, and I guess more available than zstd?

tower-http/src/services/fs/serve_dir/mod.rs
zamazan4ik: If we are talking about files, I guess `.zst` is the right choice: https://www.iana.org/assignments/media-types/application/zstd

tower-http/src/compression_utils.rs
jplatte: I wasn't really serious, by the way. But I agree with @Nehliin, if we want to sort by quality then `br` should also be near the start.

Nehliin (approved):
LGTM

PR #325, opened 2023-01-26
Prefer compression encodings: brotli > gzip > deflate

casey:
This PR changes the preference of compression encodings when multiple encodings are acceptable to the client at the same quality value. It prefers brotli over gzip over deflate, since that seems to be the general consensus as to which compression schemes are preferred when multiple are available.

I think this is a good default, since if the browser accepts brotli and gzip, and both the brotli and gzip features are enabled, the user would probably prefer brotli.


Nehliin (approved):

casey:
Nice, thank you for reviewing!

PR #329, opened 2023-01-30
support to string display for RequestId

songwei163:

### DIFF
--- a/tower-http/src/request_id.rs
+++ b/tower-http/src/request_id.rs
@@ -172,7 +172,7 @@ use http::{
     Request, Response,
 };
 use pin_project_lite::pin_project;
-use std::task::{Context, Poll};
+use std::{task::{Context, Poll}, fmt};
 use std::{future::Future, pin::Pin};
 use tower_layer::Layer;
 use tower_service::Service;
@@ -212,6 +212,12 @@ impl RequestId {
 impl From<HeaderValue> for RequestId {
     fn from(value: HeaderValue) -> Self {
         Self::new(value)
+    }
+}
+
+impl fmt::Display for RequestId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.to_str().unwrap_or_default())
     }
 }
 


songwei163:
support to string display for RequestId

tower-http/src/request_id.rs
```diff
@@ -215,6 +215,12 @@ impl From<HeaderValue> for RequestId {
     }
 }
 
+impl fmt::Display for RequestId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.to_str().unwrap_or_default())
```
Nehliin, line 220: This will be very misleading if the value isn't convertible to string so this approach won't work

Nehliin (changes requested):

tower-http/src/request_id.rs
songwei163: Can uuid not be converted into a string?

davidpdrsn: Sure `Uuid` can, but `RequestId` can be [created from any `HeaderValue`](https://github.com/tower-rs/tower-http/blob/master/tower-http/src/request_id.rs#L197) which might not be valid utf-8.

songwei163: Considering business attributes, request-id is a string in most cases. Otherwise the meaning of the display cannot be found

songwei163: If the request-id is not composed of ascii characters, it is not a request-id.because it doesn't make any sense

davidpdrsn: I don't agree with that. Non-ascii request ids are fine to uniquely identify a request. You just need the id to be unique.

davidpdrsn:
Per the discussion in https://github.com/tower-rs/tower-http/pull/329#pullrequestreview-1274681172 I don't think it makes sense to implement `Display` for `RequestId`. We could add a `RequestId::to_str` helper that calls `HeaderValue::to_str` and returns a `Result<&str, _>`. You're welcome to open a PR for that.

But for now I think I'll close this.

PR #335, opened 2023-03-11
Implement Default for MakeRequestUuid

tottoto:
## Motivation

Hides the details of MakeRequestUuid in instantiating it.

## Solution

Implements Default for MakeRequestUuid.

### DIFF
--- a/tower-http/src/request_id.rs
+++ b/tower-http/src/request_id.rs
@@ -591,7 +591,7 @@ mod tests {
     #[tokio::test]
     async fn uuid() {
         let svc = ServiceBuilder::new()
-            .set_x_request_id(MakeRequestUuid::default())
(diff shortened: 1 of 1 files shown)


tower-http/src/request_id.rs
```diff
@@ -591,7 +591,7 @@ mod tests {
     #[tokio::test]
     async fn uuid() {
         let svc = ServiceBuilder::new()
-            .set_x_request_id(MakeRequestUuid)
+            .set_x_request_id(MakeRequestUuid::default())
```
davidpdrsn, line 594: I'm okay with deriving default but don't think we need to use it here. It's just more typing.

tottoto: Fixed.

davidpdrsn (approved):

PR #341, opened 2023-03-20
Update to bitflags 2

tottoto:
## Motivation

Updates to [bitflags version 2](https://github.com/bitflags/bitflags/releases/tag/2.0.2).

## Solution

Updates the manifest.


Nehliin (approved):
Doesn't look this this will be a breaking change so LGTM

PR #347, opened 2023-03-24
Fix path normalization of index route

### LINKED ISSUE #318: Index route returns 404 with `NormalizePath` middleware
## Bug Report
Versions:
`tower-http = "0.3.5"`
`axum = "0.6.1"`

`Axum` "/" route doesn't work (returns 404) with `NormalizePath` middleware

``` rust
#[tokio::main]
async fn main() {
    let app = NormalizePath::trim_trailing_slash(
        Router::new()
            .route("/", get(|| async { "index" }))
            .route("/other", get(|| async { "other" }))
    );

    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    axum::Server::bind(&addr).serve(app.into_make_service()).await.expect("Failed to start axum server");
}
```

cassaundra:
Previously, `normalize_path` on `/` would result in an empty string, causing routers to return 404 on the index route. Fix this by defaulting to `/` if the resultant path would be empty.

Fixes #318.

### DIFF
--- a/tower-http/src/normalize_path.rs
+++ b/tower-http/src/normalize_path.rs
@@ -102,12 +102,12 @@ where
     }
 
     fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
-        remove_trailing_slash(req.uri_mut());
+        normalize_trailing_slash(req.uri_mut());
         self.inner.call(req)
     }
 }
 
-fn remove_trailing_slash(uri: &mut Uri) {
+fn normalize_trailing_slash(uri: &mut Uri) {
     if !uri.path().ends_with('/') {
         return;
     }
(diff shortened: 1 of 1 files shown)


cassaundra:
I just added an improved implementation which also removes multiple trailing slashes in the case of an index (e.g. `////` to `/`).

Nehliin (commented):
I don't mind the solution of adding a "/" but it feels wrong to do this within a function called `remove_trailing_slash`. Could this logic be extracted to before calling into the method? If not we should at least rename the function

cassaundra:
How about `normalize_trailing_slash`?

Nehliin (approved):
Thanks

Issue #353, opened 2023-04-10
Support concatenated GZIP streams


cmackenzie1:
### Motivation

GZIP files have a property where multiple GZIP files concatenated together are still considered a valid GZIP file. Currently, the GZIP decompression does not use the async-compression [`multiple_members`](https://docs.rs/async-compression/latest/async_compression/tokio/bufread/struct.GzipDecoder.html#method.multiple_members), which allows for streams of multiple gzip files, one after another. 

### Proposal

Allow configuring the `multiple_members` attribute for the decoders so users can enable it. 

### Alternatives

- Expose a new Multi*Decoders that allow for reading streams of files.

davidpdrsn:
Any reason we wouldn't enable that all the time?

cmackenzie1:
> Any reason we wouldn't enable that all the time?

Nothing that I can think of. These are technically valid gzips and most libraries handle this automatically. At least that's the case for Go and the bash command.

cmackenzie1:
@davidpdrsn created #354.

PR #354, opened 2023-04-10
Enable `multiple_members` for `GzipDecoder`

cmackenzie1:
This commit enables support for `multiple_members` which allows the `GzipDecoder` to support gzip streams consisting of multiple, concatenated gzip's - which in turn is a valid gzip.

Before this commit, when using the decompression middleware, the input stream would be closed after reading the first gzip chunk, ignoring the rest of the gzip - meaning the resulting decompressed body could be missing data but still considered "OK".

An example of a concatenated gzip:

```bash
echo "hello" | gzip > hello.txt.gz
echo "world" | gzip > world.txt.gz

cat hello.txt.gz world.txt.gz > hello-world.txt.gz

gzip -d < hello-world.txt.gz
hello
world
```

Fixes: #353

## Motivation

#353

## Solution

Set `multiple_members(true)` on the `GzipDecoder`.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 
-- None.
+- **decompression:** Enable `multiple_members` for `GzipDecoder` ([#354])
 
 ## Removed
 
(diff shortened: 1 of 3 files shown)


Nehliin (approved):

davidpdrsn (commented):
Thanks! I think this looks good. Great that that you also added tests.

Do you wanna update the changelog as well? Then this is good to merge.

cmackenzie1:
Sure, just pushed the changes for the CHANGELOG.md!

davidpdrsn (approved):
Thanks!

Issue #287, opened 2022-08-05
Compression encoding preference


daxpedda:
### Motivation

Currently, when using Firefox for example, the encoding used by `CompressionLayer` is GZIP, which is unfortunate, considering that Brotli produces far smaller files and Firefox does support it.

### Proposal

Currently the encoding used by `CompressionLayer`, is the first one listed in `Accept-Encoding`, which, as far as I can tell, is not mandated by the specification, see [RFC 9110 Section 12.5.3](https://httpwg.org/specs/rfc9110#field.accept-encoding) and [8.4.1](https://httpwg.org/specs/rfc9110#content.codings).

Furthermore, most servers don't handle it this way, and just pick the one from the list they prefer the most. When clients want to specify which one they prefer, they have to use quality values, which is already supported by `CompressionLayer` (as far as I can tell).

So my proposal is:
- Specify a default server preference for each encoding instead of picking the first one.
- Expose an interface to let the user specify a preference.

Happy to make a PR of course.

### Alternatives

For example, is Brotli is the encoding preferred, users could always just only enable that crate feature, considering Brotli is widely supported. Obviously this isn't ideal.

Otherwise implementing `CompressionLayer` by hand always works.

davidpdrsn:
Adding a method for specifying the preferred compression (if the client doesn't have a preference) sounds good to me!

casey:
This would be very nice! I enabled brotli and gzip, and although it wasn't specified in the docs, I assumed that if brotli and gzip were both acceptable to the client, brotli would be used.

casey:
With #325, tower-http will now prefer brotli > gzip > deflate, when more than one encoding is acceptable to the client. There still isn't a way for users of tower-axum to express a preference between compression schemes, but the new defaults might mean there isn't as much of a need to do so, so that might be enough to close this issue.

dirkjonker:
Brotli decreases performance for large json files by an order of magnitude for me, roughly 10 times slower than gzip. The file size may be a lot smaller but the huge latency penalty wasn't worth it.

daxpedda:
This is a known problem that keeps popping up when using `async-compression`.
See https://github.com/Nemo157/async-compression/issues/174.

[...]

PR #356, opened 2023-04-14
Override default brotli compression level 11 -> 4

Ameobea:
Additional discussion: https://github.com/tower-rs/tower-http/issues/287

## Motivation

The `brotli` crate used by `async-compression` has a default compression level of 11, which is the maximum for brotli.  This causes extremely slow compression performance for many response bodies and is definitely an inappropriate compression level for dynamic content.

There's currently an open issue (https://github.com/dropbox/rust-brotli/issues/93) on the `brotli` crate's repo to change this default, but it hasn't happened at this time.

## Solution

This change adds a special case to convert a provided compression level of default to a compression level of 4, which is a reasonable default for dynamic content.

### DIFF
--- a/tower-http/src/compression/body.rs
+++ b/tower-http/src/compression/body.rs
@@ -332,7 +332,16 @@ where
     type Output = BrotliEncoder<Self::Input>;
 
     fn apply(input: Self::Input, quality: CompressionLevel) -> Self::Output {
-        BrotliEncoder::with_quality(input, quality.into_async_compression())
+        // The brotli crate used under the hood here has a default compression level of 11,
+        // which is the max for brotli. This causes extremely slow compression times, so we
+        // manually set a default of 4 here.
+        //
+        // This is the same default used by NGINX for on-the-fly brotli compression.
+        let level = match quality {
+            CompressionLevel::Default => async_compression::Level::Precise(4),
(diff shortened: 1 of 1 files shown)


Ameobea:
An alternative way of implementing this that I just considered is altering the `into_async_compression` function.  We could create a configurable default level for each compression algorithm and then pass the algorithm type as an argument to the function.

It's a bit less hacky and allows us to close the loop for all the other algorithms as well.  Will have a bigger impact though. 
 Lmk if you'd prefer that and I can change over to doing that instead.

Logarithmus:
@davidpdrsn Recently we discovered a sudden performance drop in our web API and after troubleshooting pin-pointed it to clients switching from gzip to brotli, and further investigation led us here. For now we've overriden `CompressionLayer::quality` to `fastest`, but it would be nice to merge this so other people won't suddenly encounter this issue like us.

3 months ago the https://github.com/dropbox/rust-brotli/issues/93 has been created, but no reply at all. Also latest commit has been back in January. Looks like fixing this upstream won't happen. So I think you should accept this PR and create a new patch release 0.4.2, so the whole rust web ecosystem could benefit from it.

Nehliin (approved):

PR #360, opened 2023-04-17
feat(ValidateRequestHeaderLayer): add has_header("...").with_value("...") function

Oliboy50:
## Motivation

At my company, we use a CDN (AWS Cloudfront) in front of our Rust applications... 

Then, the AWS recommendation to make sure that all requests come from our CDN (and not from a "hacker" that somehow managed to find the direct IP address of our Rust application) is to make the CDN set a "secret header" with a "secret value" and to assert, in the Rust application, that this "secret header" has been correctly set with the expected "secret value".

AWS documented this behavior [here for classical Cloudfront origins](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/restrict-access-to-load-balancer.html#restrict-alb-add-custom-header) and [here for Lambda URLs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/DownloadDistS3AndCustomOrigins.html#concept_lambda_function_url).

However, since `tower-http` currently does not offer this kind of "header checker", instead of duplicating a `ValidateRequestHeaderLayer::custom()` closure in all our Rust applications, we had to make our own internal Rust crate containing our own "tower http layer" that uses `ValidateRequestHeaderLayer` to do this check... It works fine, but I think it would be best for everyone to have this simple layer directly offered by `tower-http` (under the `validate-request` feature flag).

## Solution

Add a ~`ValidateRequestHeaderLayer::assert()`~ `ValidateRequestHeaderLayer::has_header("...").with_value("...").with_status_code(StatusCode::FORBIDDEN)` function that acts like `ValidateRequestHeaderLayer::accept()` but for any header.

### DIFF
--- a/tower-http/src/validate_request.rs
+++ b/tower-http/src/validate_request.rs
@@ -53,7 +53,7 @@
 //! # }
 //! ```
 //!
-//! Validation of a custom header can be made by using [`ValidateRequestHeaderLayer::has_header()`]:
+//! Validation of a custom header can be made by using [`ValidateRequestHeaderLayer::has_header_value()`]:
 //!
 //! ```
 //! use tower_http::validate_request::ValidateRequestHeaderLayer;
@@ -70,11 +70,10 @@
 //! # async fn main() -> Result<(), BoxError> {
 //! let mut service = ServiceBuilder::new()
 //!     // Require a `X-Custom-Header` header to have the value `random-value-1234567890` or reject with a `403 Forbidden` response
-//!     .layer(
-//!         ValidateRequestHeaderLayer::has_header("x-custom-header")
-//!             .with_value("random-value-1234567890")
-//!             .with_status_code(StatusCode::FORBIDDEN),
-//!     )
+//!     .layer(ValidateRequestHeaderLayer::has_header_value(
+//!         "x-custom-header",
+//!         "random-value-1234567890",
+//!     ))
 //!     .service_fn(handle);
 //!
 //! // Requests with the correct value are allowed through
@@ -122,6 +121,62 @@
 //! # }
 //! ```
 //!
+//! To require only that a header is present, use [`ValidateRequestHeaderLayer::custom()`]:
+//!
+//! ```
+//! use tower_http::validate_request::ValidateRequestHeaderLayer;
+//! use http::{Request, Response, StatusCode};
+//! use http_body_util::Full;
+//! use bytes::Bytes;
+//! use tower::{ServiceBuilder, service_fn, BoxError};
+//!
+//! async fn handle(request: Request<Full<Bytes>>) -> Result<Response<Full<Bytes>>, BoxError> {
+//!     Ok(Response::new(Full::default()))
+//! }
+//!
+//! # fn main() {
+//! let service = ServiceBuilder::new()
+//!     .layer(ValidateRequestHeaderLayer::custom(|req: &mut Request<Full<Bytes>>| {
+//!         if req.headers().contains_key("x-custom-header") {
+//!             Ok(())
+//!         } else {
+//!             let mut res = Response::new(Full::<Bytes>::default());
+//!             *res.status_mut() = StatusCode::FORBIDDEN;
+//!             Err(res)
+//!         }
+//!     }))
+//!     .service_fn(handle);
+//! # }
+//! ```
+//!
+//! To serve a custom response when validation fails, also use [`ValidateRequestHeaderLayer::custom()`]:
+//!
+//! ```
+//! use tower_http::validate_request::ValidateRequestHeaderLayer;
+//! use http::{Request, Response, StatusCode};
+//! use http_body_util::Full;
+//! use bytes::Bytes;
+//! use tower::{ServiceBuilder, service_fn, BoxError};
+//!
+//! async fn handle(request: Request<Full<Bytes>>) -> Result<Response<Full<Bytes>>, BoxError> {
+//!     Ok(Response::new(Full::default()))
+//! }
+//!
+//! # fn main() {
+//! let service = ServiceBuilder::new()
+//!     .layer(ValidateRequestHeaderLayer::custom(|req: &mut Request<Full<Bytes>>| {
+//!         match req.headers().get("x-custom-header").map(|v| v.as_bytes()) {
+//!             Some(b"random-value-1234567890") => Ok(()),
+//!             _ => Err(Response::builder()
+//!                 .status(StatusCode::FORBIDDEN)
+//!                 .body(Full::<Bytes>::default())
+//!                 .unwrap()),
+//!         }
+//!     }))
+//!     .service_fn(handle);
+//! # }
+//! ```
+//!
 //! Custom validation can be made by implementing [`ValidateRequest`]:
 //!
 //! ```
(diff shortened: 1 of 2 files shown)


tower-http/src/validate_request.rs
```diff
@@ -50,6 +52,70 @@
...
+//! #    .unwrap();
+//! #
+//! # let response = service
+//! #    .ready()
+//! #    .await?
+//! #    .call(request)
+//! #    .await?;
+//! #
+//! # assert_eq!(StatusCode::FORBIDDEN, response.status());
+//! #
```
Oliboy50, line 114: ℹ️ this has been hidden from the documentation because it feels redundant... but it has been kept because this remains a valuable test

tower-http/src/validate_request.rs
```diff
@@ -112,6 +178,8 @@
 //! # Ok(())
 //! # }
 //! ```
+//!
+//! [`Accept`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
```
Oliboy50, line 182: not too sure about this one... 

I think that it is mostly useless because I didn't find any link to this in the doc... but I copy/paste it from the existing `accept()` function documentation 🤷

tower-http/src/validate_request.rs
```diff
@@ -50,6 +52,70 @@
...
+//!
+//! async fn handle(request: Request<Body>) -> Result<Response<Body>, Error> {
+//!     Ok(Response::new(Body::empty()))
+//! }
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! let mut service = ServiceBuilder::new()
+//!     // Require a `X-Custom-Header` header to have the value `random-value-1234567890` or reject with a `403 Forbidden` response
+//!     .layer(ValidateRequestHeaderLayer::assert("x-custom-header", "random-value-1234567890", StatusCode::FORBIDDEN))
```
Oliboy50, line 71: I also though about using a Builder pattern here

so instead of the `assert` function, the API would look like:

```rs
.layer(
    ValidateRequestHeaderLayer::has("x-custom-header").with_value("random-value-1234567890").or_reject_with(StatusCode::FORBIDDEN)
)
```

where `or_reject_with` would be the "build" method...
and we could say that `with_value` is optional, so this new layer could also be used by those who just want to make sure that a request has a specific header but we don't care about its value

Oliboy50:
ℹ️ I don't think that the CI is failing because of this PR
Capture d’écran 2023-05-09 à 10 17 36

tower-http/src/validate_request.rs
jlizen: @Oliboy50 extremely belated, apologies... I like the looks of this proposed API.

I agree that this is a fairly common use case that we might as well support cleanly.

I'd suggest a name tweak:
```
.layer(
    ValidateRequestHeaderLayer::has_header("x-custom-header")
        .with_value("random-value-1234567890")
        .with_status_code(StatusCode::FORBIDDEN)
)
```

It also looks like we could apply `with_status_code` to all of the different validation filters. Anyway most would like custom validation responses, yes?

Or were you trying to set up specific validation policy failure to response mappings?

jlizen: Mild preference for `with_status_code` to match timeout layer's new API:
https://docs.rs/tower-http/latest/tower_http/timeout/struct.TimeoutLayer.html#method.with_status_code

Oliboy50: 👋 your tweaks sounds very good to me 👍

jlizen: Cool! Glad to review it whenever you have time, just ping me.

Oliboy50: 👋 @jlizen sorry for the long delay, I finished to apply the corresponding changes, I hope it fits what you had in mind 🤞

jlizen (changes requested):
Deep apologies, I gave this a quick read the first time around as part of bulk triage, but my suggestion of `with_status_code()` is quite misleading since this is triggered on rejection, and semantics are unclear.

I'm also not convinced that this extra builder API pulls its weight.

Why not just:
```
  ValidateRequestHeaderLayer::has_header_value("x-custom-header", "expected-value")
```

that returns a `403` when not met?

We're going to wind up with a fairly complex typestate builder API otherwise that saves us effectively 3 lines of code using `custom`:
```
  ValidateRequestHeaderLayer::custom(|req: &mut Request<_>| {
      match req.headers().get("x-custom-header").map(|v| v.as_bytes()) {
          Some(b"random-value-1234567890") => Ok(()),
          _ => Err(Response::builder().status(403).body(Body::default()).unwrap()),
      }
  })
```

We could have a doc example for "to roll your own" that shows doing it only looking for a header, not a value, or serving a custom resposne besides the default 403.

My sense is the simple case (header/value => 403 on no match) covers 95% of the use cases with a much smaller API surface.

WDYT @Oliboy50 ?

Oliboy50:
@jlizen no problem, I updated the PR one more time ✅

tower-http/src/validate_request.rs
jlizen: Is there a reason it couldn't just be a unit test?

tower-http/src/validate_request.rs
```diff
@@ -168,6 +270,53 @@ impl<ResBody> ValidateRequestHeaderLayer<AcceptHeader<ResBody>> {
...
+    /// let request = Request::builder()
+    ///     .header("x-custom-header", "random-value-1234567890")
+    ///     .body(Full::default())
+    ///     .unwrap();
+    ///
+    /// let response = service.ready().await.unwrap().call(request).await.unwrap();
+    /// assert_eq!(response.status(), StatusCode::OK);
+    /// # }
+    /// ```
+    pub fn has_header_value(expected_header_name: &str, expected_header_value: &str) -> Self
```
jlizen: Can we parse to `HeaderName` and store it internally as that instead of a String?

That way we could return a result and reject invalid header names rather than silently rejecting every request. We probably should note that it expects a non-empty string with <= 32,768 bytes, with only valid characters per RFC 7230.

Also simplifies our internals a bit.

tower-http/src/validate_request.rs
```diff
@@ -168,6 +270,53 @@ impl<ResBody> ValidateRequestHeaderLayer<AcceptHeader<ResBody>> {
     }
 }
 
+impl<ResBody> ValidateRequestHeaderLayer<AssertHeaderOrReject<ResBody>> {
+    /// Validate requests have a required header with a specific value.
```
jlizen, line 261: We should also document the behavior if the same header is set multiple times (uses the first)

tower-http/src/validate_request.rs
```diff
@@ -411,6 +560,67 @@ where
...
+            let mut res = Response::new(ResBody::default());
+            *res.status_mut() = StatusCode::FORBIDDEN;
+            return Err(res);
+        }
+
+        Ok(())
+    }
+}
+
 #[cfg(test)]
```
jlizen, line 623: We should test the corner cases as well:
- multiple header values
- invalid header name
- invalid (non-utf8) header value
- header name that is different casing (should still match)
- header value with different casing (would not match)

jlizen (changes requested):
Core impl looks solid, some smaller tweaks

jlizen (approved):
Thanks for the contribution @Oliboy50 !

PR #362, opened 2023-04-20
Fix broken links in doc for #334

### LINKED ISSUE #334: Broken links and missing examples for `TimeoutLayer`
### Version

0.4.0

### Platform

https://docs.rs/tower-http/0.4.0/tower_http/

### Crates

`tower-http`

### Description

On <https://docs.rs/tower-http/latest/tower_http/timeout/struct.TimeoutLayer.html> and other types within this module the "See the [module docs](https://docs.rs/tower-http/latest/tower_http/timeout/self) for an example." link is broken (and the module docs don't contain any example).

There are some warnings in [the build logs](https://docs.rs/crate/tower-http/0.4.0/builds/751816) about `self` being private, because this should likely be `[module docs](super)` as the types for this module are actually defined in sub-modules. (This is a new `rustdoc` warning that might make sense to add to your CI job).

erebe:
Fix broken documentation as reported in issue #334 

The documentation is not generated for the mentioned modules because they are declared as private.
This PR make them public to allow the generation of the doc for the modules. (I took as example other modules in tower-http)

Making those modules pub does not allow visibility of anything more than what is already re-exported

edit(jplatte): Closes #334.

### DIFF
--- a/tower-http/src/timeout/body.rs
+++ b/tower-http/src/timeout/body.rs
@@ -11,6 +11,7 @@ use tokio::time::{sleep, Sleep};
 
 pin_project! {
     /// Middleware that applies a timeout to request and response bodies.
+    ///
     /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration.
     ///
     /// Bodies must produce data at most within the specified timeout.
--- a/tower-http/src/timeout/body.rs
+++ b/tower-http/src/timeout/body.rs
@@ -1,40 +1,3 @@
-//! Middleware that applies a timeout to request and response bodies.
-//!
-//! Bodies must produce data at most within the specified timeout.
-//! If the body does not produce a requested data frame within the timeout period, it will return an error.
-//!
-//! # Differences from [`crate::timeout::Timeout`]
-//!
-//! [`crate::timeout::Timeout`] applies a timeout to the request future, not body.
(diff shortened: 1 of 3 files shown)


tower-http/src/timeout/mod.rs
```diff
@@ -1,9 +1,11 @@
 //! Middleware for setting timeouts on requests and responses.
 
-mod body;
-mod service;
+pub mod body;
+pub mod service;
```
davidpdrsn, line 4: I think I'd rather move the body docs to `TimeoutBody` and the service docs to this module. Rather than making the separate modules public.

erebe: I have updated the PR with your preferences

tower-http/src/timeout/body.rs
```diff
@@ -47,7 +10,43 @@ use std::{
 use tokio::time::{sleep, Sleep};
 
 pin_project! {
+    /// Middleware that applies a timeout to request and response bodies.
     /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration.
```
jplatte, line 14:
```suggestion
    /// Middleware that applies a timeout to request and response bodies.
    ///
    /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration.
```

jplatte (approved):

PR #368, opened 2023-05-17
Preserve service error type in RequestDecompression

rtimush:
There seems to be no reason for `RequestDecompression` to change the service error type to `BoxError`, so it would be nice to preserve the underlying error type. This is especially useful if the inner service is infallible.
 
## Motivation

When I add `RequestDecompressionLayer` to a stack of infallible layers, the resulting service has `BoxError` as an error type, but it never actually fails.

## Solution

Make `RequestDecompression` preserve the error type of the inner service.


davidpdrsn (approved):
Thanks! I'll fix CI and update the changelog.

PR #372, opened 2023-05-23
Move examples to its own workspace

tottoto:
## Motivation

- Fixes the ci and makes it more robust.
- Releases examples from `tower-http`'s msrv.

## Solution

Moves examples to its own workspace. As `tower-http` is relatively lower layer library and it's real world examples tend to use a lot of crates as dependencies, the advantages which treating examples and `tower-http` as a single workspace seem small.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,6 +36,21 @@ jobs:
         RUSTDOCFLAGS: "-D rustdoc::broken_intra_doc_links"
       run: cargo doc --all-features --no-deps
 
+  check-examples:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


jplatte:
I don't think this is a good idea. We had this setup in axum for a while, but reverted ultimately because it meant that rust-analyzer wouldn't check the examples still compiled when you changed something in one of the main crates.

We could still separate the MSRV in CI though: Just declare `default-members = ["tower-http"]` in the workspace manifest, and remove `--workspace` from all the CI jobs not meant to check examples.

tottoto:
I wasn't aware of the `rust-analyzer` aspect. I agree that this is not a good approach. I will try to use `default-members`.

PR #373, opened 2023-05-29
Add support for private network preflights - continue

Ptrskay3:
Continued from #315, original issue is #314.

Added some basic sanity checks. If anything needs changes, let me know.

### DIFF
--- a/tower-http/src/cors/allow_private_network.rs
+++ b/tower-http/src/cors/allow_private_network.rs
@@ -84,15 +84,6 @@ impl From<bool> for AllowPrivateNetwork {
     }
 }
 
-impl<F> From<F> for AllowPrivateNetwork
-where
-    F: Fn(&HeaderValue, &RequestParts) -> bool + Send + Sync + 'static,
-{
-    fn from(f: F) -> Self {
-        AllowPrivateNetwork::predicate(f)
-    }
-}
-
 impl fmt::Debug for AllowPrivateNetwork {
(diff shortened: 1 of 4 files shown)


tower-http/src/cors/allow_private_network.rs
```diff
@@ -84,6 +84,15 @@ impl From<bool> for AllowPrivateNetwork {
     }
 }
 
+impl<F> From<F> for AllowPrivateNetwork
+where
+    F: Fn(&HeaderValue, &RequestParts) -> bool + Send + Sync + 'static,
+{
+    fn from(f: F) -> Self {
+        AllowPrivateNetwork::predicate(f)
```
jplatte, line 92: The other `Allow` types that have a `predicate` constructor don't have such `From` impls. I don't think we should add one for `AllowPrivateNetwork`. I'm pretty sure it will degrade the error message when a non-matching type is passed to a function expecting `impl Into<AllowPrivateNetwork>`.

Ptrskay3: Yes, I see that with this `From` impl the error message gets worse and for example no longer recommends that _`From<bool>` is implemented for `AllowPrivateNetwork`_. Removed it.

jplatte (approved):

Nehliin (approved):

jplatte:
CI errors are unrelated to this PR.. The cargo-public-api-crates issue has already been fixed upstream AFAIK, the other one should be fixed by #385.

Issue #374, opened 2023-06-05
Exclude examples from tower-http dependencies and MSRV inspection


tottoto:
### Motivation

The motivation to exclude examples from `tower-http` dependencies and MSRV inspection is to improve ci robustness and allow us to free from `tower-http`'s limitations such as MSRV in examples. As `tower-http` is relatively lower layer library and it's real world examples tend to use a lot of crates as dependencies, treating their dependencies together impose extra limitations. I think these limitations might not be worthy for `tower-http`.

### Proposal

This proposal and the alternative's content are thanks to https://github.com/tower-rs/tower-http/pull/372#issuecomment-1577088925.

Declares `default-members = ["tower-http"]` in the workspace manifest, and remove `--workspace` from all the CI jobs.

### Alternatives

Moves examples to its own workspace. But this is not a good idea as decreases development experience with `rust-analyzer`.

tottoto:
Declares `default-menbers` at `Cargo.toml` as

```diff
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+default-members = ["tower-http"]
 members = [
   "tower-http",
   "examples/*",
```

and disable `env-filters` of `tracing-subscriber` at `tower-http/Cargo.toml` to remove duplicate `regex-syntax` dependency detected by `cargo-deny`.

```diff
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -47,7 +47,7 @@ hyper = { version = "0.14", features = ["full"] }
 once_cell = "1"
 tokio = { version = "1", features = ["full"] }
 tower = { version = "0.4.10", features = ["buffer", "util", "retry", "make", "timeout"] }
-tracing-subscriber = "0.3"
+tracing-subscriber = { version = "0.3", default_features = false, features = ["fmt"] }
 uuid = { version = "1.0", features = ["v4"] }
 serde_json = "1.0"
 zstd = "0.11"
```

`cargo-deny` still reports duplicate `regex-syntax` as follows though it is not depended by `tower-http`.

```
$ cargo deny --all-features --manifest-path ./tower-http/Cargo.toml check
error[duplicate]: found 2 duplicate entries for crate 'regex-syntax'
   ┌─ /home/test/tower-http/Cargo.lock:74:1
   │
74 │ ╭ regex-syntax 0.6.29 registry+https://github.com/rust-lang/crates.io-index
[...]

PR #375, opened 2023-06-06
Update deny config

tottoto:
## Motivation

Reflects the current state to `cargo-deny` config to fix `cargo-deny` test.

## Solution

Updates it. Refs #374.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -2,12 +2,7 @@
 vulnerability = "deny"
 unmaintained = "warn"
 notice = "warn"
-ignore = [
-  # Potential Unix segfault with time crate, used by chrono.
-  # if setting env vars in threaded context, we are unlikely affected
-  # https://github.com/time-rs/time/issues/293
-  "RUSTSEC-2020-0159",
-]
+ignore = []
 
 [licenses]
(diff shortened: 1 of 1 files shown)


Nehliin (approved):
Thanks!

Nehliin:
Do you mind trying to get the CI api check fix in here as well? otherwise I can do it a bit later but you'll have to rebase afterwards

deny.toml
```diff
@@ -36,6 +31,11 @@ license-files = [
 [bans]
 multiple-versions = "deny"
 highlight = "all"
+skip = [
+  # tracing-subscriber depends on multiple versions of regex-syntax.
+  # This is not actualy depended by tower-http as this is caused by examples.
```
jplatte, line 36:
```suggestion
  # This is not actually depended by tower-http as this is caused by examples.
```

tottoto: Thanks!

tottoto:
@Nehliin Sure. I'll take a look it.

Nehliin:
No I don't think we should remove it but it doesn't need to block this pr for now

PR #385, opened 2023-07-19
Unbreak CI

jplatte:
## Motivation

CI is failing in other PRs, with no fault of the PR authors.

## Solution

Stop erroring on duplicate dependencies in the `cargo-deny` check.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -29,14 +29,8 @@ license-files = [
 ]
 
 [bans]
-multiple-versions = "deny"
+multiple-versions = "warn"
 highlight = "all"
-skip = [
-  # tracing-subscriber depends on multiple versions of regex-syntax.
-  # This is not actually depended by tower-http as this is caused by examples.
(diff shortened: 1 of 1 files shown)


Nehliin:
I'd rather have specific exceptions than accepting all duplicate dependencies

davidpdrsn (approved):

davidpdrsn:
I'm okay either way 😅

jplatte:
> I'd rather have specific exceptions than accepting all duplicate dependencies

The problem with this, as mentioned in the commit message, is that CI frequently breaks from it from dependencies just doing their regular maintenance and upgrading dependencies. I also like avoiding duplicate dependencies, very much (recently went to upgrade syn in a bunch of places that hadn't upgraded to 2.0 yet). But more often than not, it's non-trivial to fix these and unrelated contributions get stuck on it, especially when the project isn't receiving frequent updates (like tower-http, currently).

Nehliin:
Fair enough

Nehliin (approved):

Issue #384, opened 2023-07-19
ServeDir: requests that ask for a range that extends past the end of a document return a 416 response


jfaust:
### Version

`0.4.1`

### Description

This is... not necessarily a bug report so much as a compatibility report? I was just working on getting some SEO things working on my site, only to discover that LinkedIn's post inspector fails (https://www.linkedin.com/post-inspector/). This is because their scraper send a `range: bytes=0-3145727` header no matter what. I'm assuming this is so they only pull up to 3MB of the page... however, it causes `ServeDir` to return a 416 (Range Not Satisfiable). On the face of it this seems like the correct behavior for `ServeDir`. However, I've just tested a few http servers (using `curl`), and so far they're all happy just returning the full document in this case.

Thoughts?

jfaust:
FYI, not an urgent question, as I've written some middleware to strip the range header in this case. Just thought it'd be worth bringing up here.

davidpdrsn:
Yes I think so

jfaust:
This turns out a little more complicated than I was expecting, because the actual error is getting raised by `http-range-header::ParsedRanges::validate()`. 

Does this change belong there? I didn't originally think so, but now I'm not sure. Reading https://www.ietf.org/rfc/rfc2616.txt, section `10.4.17`:
```
10.4.17 416 Requested Range Not Satisfiable

   A server SHOULD return a response with this status code if a request
   included a Range request-header field (section 14.35), and none of
   the range-specifier values in this field overlap the current extent
   of the selected resource, and the request did not include an If-Range
   request-header field. (For byte-ranges, this means that the first-
   byte-pos of all of the byte-range-spec values were greater than the
   current length of the selected resource.)

   When this status code is returned for a byte-range request, the
   response SHOULD include a Content-Range entity-header field
   specifying the current length of the selected resource (see section
   14.16). This response MUST NOT use the multipart/byteranges content-
   type.
```

`none of the range-specifier values in this field overlap` makes me think that the behavior of `http-range-header` is wrong in this case, but I could be parsing it wrong. I also haven't read the entire RFC, so I may be missing other context.

If `http-range-header` is right, and we just want to be lax in `ServeDir`, then I think I need to basically duplicate `ParsedRanges::validate()` into `tower-http`. I'll file a ticket with https://github.com/MarcusGrass/http-range-header and see what he thinks.

seanmonstar:
[...]

PR #391, opened 2023-07-21
Http range headers 0.4

MarcusGrass:
## Motivation
Fixes https://github.com/tower-rs/tower-http/issues/384

## Solution
Updates http-range-header to `0.4`, it broke a test rejecting a spec-valid header so that's fixed here as well. 
Added a bunch of tests in `0.4` re-ran the fuzzing and made a few optimizations. I think the performance increase was something like 15% across the board. Set the msrv on that project to the same as this one, 1.60.0. Added msrv to the `package.metadata` for `tower-http`, so I could run `cargo msrv verify` on it, it checks out. 

The actual fix came from @jfaust in this PR https://github.com/MarcusGrass/http-range-header/pull/3

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 
-- Msrv 1.60.0 in `tower-http` package metadata
+- None.
 
 ## Changed
 
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
-- Parses range headers with ranges where the end of range goes past the end of the document by bumping 
+- Accepts range headers with ranges where the end of range goes past the end of the document by bumping 
 http-range-header to `0.4`
 
 # 0.4.2 (July 19, 2023)
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 
-- None.
+- Msrv 1.60.0 in `tower-http` package metadata
 
 ## Changed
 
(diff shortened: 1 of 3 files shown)


tower-http/CHANGELOG.md
```diff
@@ -21,7 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
-- None.
+- Parses range headers with ranges where the end of range goes past the end of the document by bumping
```
jplatte, line 24:
```suggestion
- Accept range headers with ranges where the end of range goes past the end of the document by bumping 
```

tower-http/Cargo.toml
```diff
@@ -12,6 +12,9 @@ categories = ["asynchronous", "network-programming", "web-programming"]
 keywords = ["io", "async", "futures", "service", "http"]
 rust-version = "1.60"
 
+[package.metadata]
+msrv = "1.60.0"
```
jplatte, line 16: What tooling uses this? A few lines above, `rust-version` declares the same thing.

MarcusGrass: Oh wow, can't believe I missed that right above, https://github.com/foresterre/cargo-msrv on `cargo msrv verify` checks the [package.metadata] maybe it's only that.

tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -496,11 +496,16 @@ async fn read_partial_rejects_out_of_bounds_range() {
         .unwrap();
     let res = svc.oneshot(req).await.unwrap();
 
-    assert_eq!(res.status(), StatusCode::RANGE_NOT_SATISFIABLE);
+    assert_eq!(res.status(), StatusCode::PARTIAL_CONTENT);
```
jplatte, line 499: So if all bytes get returned, does it still make sense for the status code to be 'partial content'?

MarcusGrass: I did actualy consider that and if I understand the spec correctly, it does https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206
> The HTTP 206 Partial Content success status response code indicates that the request has succeeded and the body contains the requested ranges of data, as described in the [Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) header of the request.

jplatte (approved):

PR #398, opened 2023-08-17
cors: Don't overwrite vary header set by the inner service

### LINKED ISSUE #397: CORS middleware blocks other Vary headers from being set
### Version

0.4.3

### Platform

Windows 10 x64

### Description

The CORS middleware blocks all other Vary headers, only allowing its own to be set in responses.

Cargo.toml:

```toml
[package]
name = "bug-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
axum = "0.6.20"
tokio = { version = "1.29.1", features = ["macros", "rt", "rt-multi-thread"] }
tower-http = { version = "0.4.3", features = ["cors"] }
```

main.rs:
```rust
use axum::http::StatusCode;
use axum::response::{IntoResponse, Response};
use axum::routing::get;
use axum::Router;
use tower_http::cors::CorsLayer;

#[tokio::main]
async fn main() {
    let router = Router::new().route("/", get(hello)).layer(CorsLayer::permissive());
    axum::Server::bind(&"127.0.0.1:6000".parse().unwrap())
        .serve(router.into_make_service())
        .await
        .unwrap()
}

async fn hello() -> Response {
    (StatusCode::OK, [("vary", "accept, accept-encoding")], "Hello world").into_response()
}
```

This code clearly adds a Vary header. However, by sending a request to this app using curl, the only Vary header values that appear are:

```
< vary: origin
< vary: access-control-request-method
< vary: access-control-request-headers
```

Removing the CorsLayer allows `vary: accept, accept-encoding` to appear in responses.

jplatte:
## Motivation

Fixes #397.

## Solution

For GET requests where we call an inner service and compute headers for it (as opposed to assembling the whole response to be sent to the client), append those computed headers instead of overwriting existing headers set by the inner service.

@AlyoshaVasilieva: can you confirm that two `vary` headers in the response are okay for this case? With the first commit of this PR, we will no longer create multiple `vary` headers in the middleware, but with the second commit we will just keep all existing headers from the inner service so the end result in your example will be

```
< vary: accept, accept-encoding
< vary: origin, access-control-request-method, access-control-request-headers
```

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Accepts range headers with ranges where the end of range goes past the end of the document by bumping 
 http-range-header to `0.4`
+- cors: Keep Vary headers set by the inner service when setting response headers ([#398])
+
+[#398]:  https://github.com/tower-rs/tower-http/pull/398
 
 # 0.4.2 (July 19, 2023)
 
--- a/tower-http/src/cors/mod.rs
+++ b/tower-http/src/cors/mod.rs
@@ -618,24 +618,10 @@ where
 
         // These headers are applied to both preflight and subsequent regular CORS requests:
         // https://fetch.spec.whatwg.org/#http-responses
-
         headers.extend(self.layer.allow_origin.to_header(origin, &parts));
         headers.extend(self.layer.allow_credentials.to_header(origin, &parts));
         headers.extend(self.layer.allow_private_network.to_header(origin, &parts));
-
-        let mut vary_headers = self.layer.vary.values();
-        if let Some(first) = vary_headers.next() {
-            let mut header = match headers.entry(header::VARY) {
-                header::Entry::Occupied(_) => {
-                    unreachable!("no vary header inserted up to this point")
-                }
-                header::Entry::Vacant(v) => v.insert_entry(first),
-            };
-
-            for val in vary_headers {
-                header.append(val);
-            }
-        }
+        headers.extend(self.layer.vary.to_header());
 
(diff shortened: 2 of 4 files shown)


AlyoshaVasilieva:
My understanding is that it's permitted, RFC 7230 says https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2 :

>   A sender MUST NOT generate multiple header fields with the same field
>   name in a message unless either the entire field value for that
>   header field is defined as a comma-separated list [i.e., #(values)]
>   or the header field is a well-known exception (as noted below).
>
>   A recipient MAY combine multiple header fields with the same field
>   name into one "field-name: field-value" pair, without changing the
>   semantics of the message, by appending each subsequent field value to
>   the combined field value in order, separated by a comma.  The order
>   in which header fields with the same field name are received is
>   therefore significant to the interpretation of the combined field
>   value; a proxy MUST NOT change the order of these field values when
>   forwarding a message.

Since Vary is a list, the multiple Vary headers should be treated as a single list.

Random StackOverflow answers appear to agree that caches can handle multiple Vary headers.

I *think* the exception is that `Vary: *` can't appear with other Vary headers, as that value isn't a list, but I'm not experienced in this area. I've personally never seen `Vary: *`.

jplatte:
Right. The first bit is making me question the append for the other headers though. I think we might have to overwrite some (maybe while logging an error), and only append where multiple values are explicitly allowed.

jplatte:
CI failure to be fixed by #418.

jplatte:
Updated to only append vary headers, and overwrite others since they're not allowed to occur multiple times in the output. Maybe it would make sense to also deduplicate `vary` values, but that can be a separate change (it's basically just an optimization).

davidpdrsn (approved):

PR #399, opened 2023-08-20
Vary header in compression

Dragonink:
## Motivation

From [the "Compression in HTTP" MDN page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Compression#end-to-end_compression):
> As content negotiation has been used to choose a representation based on its encoding, the server must send a [Vary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary) header containing at least [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) alongside this header in the response; that way, caches will be able to cache the different representations of the resource.

## Solution

We just append `Accept-Encoding` to the `Vary` header when constructing the compression response.

### DIFF
--- a/tower-http/src/compression/future.rs
+++ b/tower-http/src/compression/future.rs
@@ -46,6 +46,12 @@ where
 
         let (mut parts, body) = res.into_parts();
 
+        if should_compress {
+            parts
+                .headers
+                .append(header::VARY, header::ACCEPT_ENCODING.into());
+        }
+
         let body = match (should_compress, self.encoding) {
             // if compression is _not_ support or the client doesn't accept it
             (false, _) | (_, Encoding::Identity) => {


jplatte:
I wonder if we shouldn't also send the header in some of the conditions wherewe early return without a content-encoding, to indicate tp caches that a different response body is available when accept-encoding is set. Wdyt?

Dragonink:
So we always send the Vary header whatever the body will be? You're right, we should do this.
I will fix that.

I have a question about branches: why has the `v0.4.x` branch (and the `0.4.3` version) diverged from `master`?
So what base branch should I target with this PR? `master` or `v0.4.x`?

jplatte:
> So we always send the Vary header whatever the body will be? You're right, we should do this. I will fix that.

No, not in every case. If `should_compress` is `false`, the body will never be compressed no matter which headers the client sends AFAIU. However, for `Encoding::Identity`, the body was not compressed specifically of what headers the client sent (again, AFAIU).

> I have a question about branches: why has the `v0.4.x` branch (and the `0.4.3` version) diverged from `master`? So what base branch should I target with this PR? `master` or `v0.4.x`?

Dragonink:
So we should always send the Vary *except* when `should_compress` is false?

jplatte:
No, I think there's another early-return branch (body already compressed, i.e. `content-encoding` already set) in which the middleware should essentially do nothing.

cc @davidpdrsn @Nehliin wdyt about when we should set `vary`?

Dragonink:
The Content-Encoding already set is tested in `should_compress`:
```rust
let should_compress = !res.headers().contains_key(header::CONTENT_ENCODING)
    && self.predicate.should_compress(&res);
```

Thus if I guard the Vary header with an `if should_compress`, the header will be added unless the response is already compressed or the response is never be compressed.

jplatte:
Ping @davidpdrsn @Nehliin any opinions about this PR / the question of when `vary` should be sent?

Also @Dragonink have you looked into the test failure? (MSRV job failing was fixed in #418)

Nehliin:
sorry won't have time to review in the near future

jplatte (approved):

PR #403, opened 2023-09-01
chore: Prepare `v0.4.4` release

LucioFranco:


Oliboy50:
👋 ℹ️ the default branch (`master`) is still on `0.4.2`, would it be possible to update the changelog file there too and not only on the `0.4.x` branch?

PR #407, opened 2023-09-10
enable clippy for tower-http and fix current issues

GlenDC:
## Motivation

I maintain a fork of `tower`, including also `tower-http` in <https://github.com/plabayo/tower-async>.
Only now did I notice that tower-http does not yet enable clippy.

Nothing bad was spotted by enabling it, but still,
seems by now that the Rust community agrees that enabling clippy everywhere is useful?
So as such, here my first contribution upstream.

## Solution

N/A

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -94,7 +94,7 @@ jobs:
     needs: check
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v6
     - uses: dtolnay/rust-toolchain@stable
       with:
         components: clippy
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -90,6 +90,20 @@ jobs:
         components: rustfmt
     - run: cargo fmt --all --check
 
+  clippy:
+    needs: check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        components: clippy
+    - name: Install protoc
+      uses: taiki-e/install-action@v2
+      with:
+        tool: protoc@3.20.3
+    - run: cargo clippy --workspace --all-features --all-targets
+
   deny-check:
     name: cargo-deny check
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jplatte:
Not sure why CI isn't running. I'll try closing and re-opening.

jplatte (approved):
Changes look good to me, but I want to also wait for David's opinion of including Clippy in CI.

Oliboy50:
this PR should be rebased

GlenDC:
> this PR should be rebased

Probably, but the bigger issue is first to ensure maintainers want to merge this in to begin with.
If this MR Is desired then I'll get the PR ready with pleasure.
But until there is consensus on what to do with this PR there is no sense to put more effort into it.

tower-http/src/lib.rs
```diff
@@ -211,7 +210,7 @@
     nonstandard_style,
     missing_docs
 )]
-#![deny(unreachable_pub, private_in_public)]
```
jplatte, line 214: I think this lint got replaced by smaller individual lints? Probably worth enabling those / some of them.

tower-http/src/lib.rs
```diff
@@ -183,7 +183,6 @@
     clippy::todo,
     clippy::empty_enum,
     clippy::enum_glob_use,
-    clippy::pub_enum_variant_names,
```
jplatte, line 186: Lint deprecated?

Cargo.toml
```diff
@@ -3,3 +3,4 @@ members = [
   "tower-http",
   "examples/*",
 ]
+resolver = "2"
```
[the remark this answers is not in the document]
Oliboy50: > this PR should be rebased

> Probably, but the bigger issue is first to ensure maintainers want to merge this in to begin with

at least because [resolver = "2" has been added to Cargo.toml meanwhile](https://github.com/tower-rs/tower-http/blob/main/Cargo.toml#L2)

so it would be easier to review for David when he will come here... if he comes here, indeed

jplatte (commented):
The removal of `dbg!` and some `&`s as well as using a nested or pattern in a match arm and switching the `allow()`ed clippy lint name where a previous lint was deprecated/replaced are all things I consider uncontroversial enough to merge. If you want to split the PR in two, I'm happy to merge that part.

jplatte:
Took another look, happy to merge once feedback that has been posted so far is adressed. David hasn't been active in quite a while.

jlizen:
Freshening this up with a rebase and making sure we can get `-DWarnings` running in CI happily. I skipped denying warnings on test code since it would have been a much more infectious set of refactors, that I didn't have the appetite for.

PR #408, opened 2023-09-11
Do compress grpc-web responses

bouk:
Related: https://github.com/tower-rs/tower-http/issues/215

## Motivation

grpc-web responses aren't compressed right now, but they should be.

## Solution

Add exception for `application/grpc-web` and change exception filter to use starts_with so we match on any encoding format.


bouk:
@jplatte can you take a look at this one?

jlizen (approved):
Change looks good to me, thank you @bouk !

@jplatte To answer your earlier question: native gRPC (`application/grpc`) has [its own compression spec](https://github.com/grpc/grpc/blob/master/doc/compression.md) so applying HTTP content-encoding on top would double-compress or break the framing. That's why we skip it in general.

 gRPC-web (`application/grpc-web`, `application/grpc-web+proto`, etc.) is a different protocol that runs over HTTP/1.1 and thus does not use gRPC's compression framing. It's designed to be treated as a normal HTTP response by intermediaries. [Its spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) doesn't explicitly specify anything about compression, unfortunately.

But, I found that frameworks like envoy DO respect compression headers for grpc-web. Here you can see both `grpc-web` and `grpc-web+proto` explicitly allowlisted in envoy's compression module with values that would match this PR's change:

https://github.com/envoyproxy/envoy/blob/97f4cec040185a6588e3f79a655f2685ae43fcda/source/extensions/filters/http/compressor/compressor_filter.cc#L68-L69

Good enough for me :)

PR #421, opened 2023-10-06
Do not leak directory info if append_index_html_on_directories = false

georgmu:
## Motivation

Currently it is possible to gain information about directories even if `append_index_html_on_directories = false`.

This is caused by the mechanism to redirect to trailing slash for directories (like testcase `redirect_to_trailing_slash_on_dir()`).

This way, an attacker could search for directories by testing URLs without trailing slash and then could continue the
search within such a subdirectory using the same mechanism.

## Solution

This pull request prevents the redirect and directly returns `404 Not Found` if `append_index_html_on_directories = false`.


jplatte (approved):
Thanks!

PR #422, opened 2023-10-07
refactor!: include grpc error message in tracing

xiaoyawei:
## Motivation

Currently the tracing middleware only captures the gRPC status code when extracting the error out of a http response. Besides the code, gRPC also requires a human-readable message for error handling, and it'll be useful if this message is extracted from the response as well.

## Solution

Per [gRPC over HTTP/2 Protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses), a percent-encoded message is included in the `grpc-message` field in the response header/trailer. This PR extracts this value and includes it in a new `GrpcStatus` struct within `GrpcFailureClass`, such that downstream usages can use this message for tracing.

Breaking changes:
- `GrpcFailureClass::Code(NonZeroI32)` → `GrpcFailureClass::Status(GrpcStatus)`
- `GrpcCode` and `GrpcFailureClass` are now `#[non_exhaustive]`
- `GrpcCode` is now `#[repr(i32)]`
- `GrpcFailureClass` implements `std::error::Error`

### DIFF
--- a/tower-http/src/classify/grpc_errors_as_failures.rs
+++ b/tower-http/src/classify/grpc_errors_as_failures.rs
@@ -1,8 +1,9 @@
 use super::{ClassifiedResponse, ClassifyEos, ClassifyResponse, SharedClassifier};
 use bitflags::bitflags;
 use http::{HeaderMap, Response};
+#[cfg(feature = "trace")]
 use percent_encoding::percent_decode;
-use std::fmt;
+use std::{fmt, num::NonZeroI32};
 
 /// gRPC status codes.
 ///
@@ -10,6 +11,8 @@ use std::fmt;
 ///
 /// [gRPC status codes]: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(i32)]
+#[non_exhaustive]
 pub enum GrpcCode {
     /// The operation completed successfully.
     Ok = 0,
@@ -70,31 +73,25 @@ impl GrpcCode {
         }
     }
 
-    fn from_bytes(bytes: &[u8]) -> Option<GrpcCode> {
-        match bytes.len() {
-            1 => match bytes[0] {
-                b'0' => Some(GrpcCode::Ok),
-                b'1' => Some(GrpcCode::Cancelled),
-                b'2' => Some(GrpcCode::Unknown),
-                b'3' => Some(GrpcCode::InvalidArgument),
-                b'4' => Some(GrpcCode::DeadlineExceeded),
-                b'5' => Some(GrpcCode::NotFound),
-                b'6' => Some(GrpcCode::AlreadyExists),
-                b'7' => Some(GrpcCode::PermissionDenied),
-                b'8' => Some(GrpcCode::ResourceExhausted),
-                b'9' => Some(GrpcCode::FailedPrecondition),
-                _ => None,
-            },
-            2 => match (bytes[0], bytes[1]) {
-                (b'1', b'0') => Some(GrpcCode::Aborted),
-                (b'1', b'1') => Some(GrpcCode::OutOfRange),
-                (b'1', b'2') => Some(GrpcCode::Unimplemented),
-                (b'1', b'3') => Some(GrpcCode::Internal),
-                (b'1', b'4') => Some(GrpcCode::Unavailable),
-                (b'1', b'5') => Some(GrpcCode::DataLoss),
-                (b'1', b'6') => Some(GrpcCode::Unauthenticated),
-                _ => None,
-            },
+    fn from_i32(code: i32) -> Option<GrpcCode> {
+        match code {
+            0 => Some(GrpcCode::Ok),
+            1 => Some(GrpcCode::Cancelled),
+            2 => Some(GrpcCode::Unknown),
+            3 => Some(GrpcCode::InvalidArgument),
+            4 => Some(GrpcCode::DeadlineExceeded),
+            5 => Some(GrpcCode::NotFound),
+            6 => Some(GrpcCode::AlreadyExists),
+            7 => Some(GrpcCode::PermissionDenied),
+            8 => Some(GrpcCode::ResourceExhausted),
+            9 => Some(GrpcCode::FailedPrecondition),
+            10 => Some(GrpcCode::Aborted),
+            11 => Some(GrpcCode::OutOfRange),
+            12 => Some(GrpcCode::Unimplemented),
+            13 => Some(GrpcCode::Internal),
+            14 => Some(GrpcCode::Unavailable),
+            15 => Some(GrpcCode::DataLoss),
+            16 => Some(GrpcCode::Unauthenticated),
             _ => None,
         }
     }
@@ -308,6 +305,7 @@ impl ClassifyEos for GrpcEosErrorsAsFailures {
 
 /// The failure class for [`GrpcErrorsAsFailures`].
 #[derive(Debug)]
+#[non_exhaustive]
 pub enum GrpcFailureClass {
     /// A gRPC response was classified as a failure with the corresponding status.
     Status(GrpcStatus),
@@ -326,6 +324,8 @@ impl fmt::Display for GrpcFailureClass {
     }
 }
 
+impl std::error::Error for GrpcFailureClass {}
+
 pub(crate) fn classify_grpc_metadata(
     headers: &HeaderMap,
     success_codes: GrpcCodeBitmask,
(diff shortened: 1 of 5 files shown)


tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -1,49 +1,50 @@
...
+use percent_encoding::percent_decode;
+use std::fmt;
 
 /// gRPC status codes. Used in [`GrpcErrorsAsFailures::success_codes`].
 ///
 /// These variants match the [gRPC status codes].
 ///
 /// [gRPC status codes]: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
jlizen, line 12: We should add `#[repr(i32)]` rather than relying on the default `isize` repr while casting

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -277,28 +307,53 @@ pub(crate) fn classify_grpc_metadata(
         };
     }
 
-    let status = or_else!(headers.get("grpc-status"), GrpcStatusHeaderMissing);
-    let status = or_else!(status.to_str().ok(), HeaderNotString);
-    let status = or_else!(status.parse::<i32>().ok(), HeaderNotInt);
+    let code = or_else!(headers.get("grpc-status"), GrpcStatusHeaderMissing);
+    let code = or_else!(GrpcCode::from_bytes(code.as_ref()), HeaderNotGrpcCode);
```
jlizen: This is a behavioral regression. Previously, >16 would be propagated as non-success, now they are converted to `HeaderNotGrpcCode`.

The spec states: 
> gRPC libraries that encounter values outside this range must either propagate them directly or convert them to UNKNOWN.

[ref](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md)

jlizen: Let's make sure we have a test for this after fixing.

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -277,28 +307,53 @@ pub(crate) fn classify_grpc_metadata(
...
         ParsedGrpcStatus::Success
     } else {
-        ParsedGrpcStatus::NonSuccess(NonZeroI32::new(status).unwrap())
+        ParsedGrpcStatus::NonSuccess(GrpcStatus {
+            code,
+            message: headers.get("grpc-message").and_then(|header| {
+                percent_decode(header.as_bytes())
+                    .decode_utf8()
+                    .map(|cow| cow.to_string())
+                    .ok()
```
jlizen: Currently we are throwing away the full value in case of non-utf8. It would be more in the spirit of the spec to do a lossy decode:
```
percent_decode(header.as_bytes()).decode_utf8_lossy()
```

> Implementations must not error due to receiving an invalid ASCII-Value that's a valid field-value in HTTP, but the precise behavior is not strictly defined: they may throw the value away or accept the value. If accepted, care must be taken to make sure that the application is permitted to echo the value back as metadata. For example, if the metadata is provided to the application as a list in a request, the application should not trigger an error by providing that same list as the metadata in the response.

[ref](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses)

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -1,49 +1,50 @@
...
+use std::fmt;
 
 /// gRPC status codes. Used in [`GrpcErrorsAsFailures::success_codes`].
 ///
 /// These variants match the [gRPC status codes].
 ///
 /// [gRPC status codes]: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum GrpcCode {
```
jlizen, line 15: Let's tag this as `#[non_exhaustive]` while we are anyway breaking semver

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -248,15 +276,17 @@ impl ClassifyEos for GrpcEosErrorsAsFailures {
 #[derive(Debug)]
 pub enum GrpcFailureClass {
```
jlizen, line 308: While we are anyway breaking semver, let's make this `#[non_exhaustive]`.

jlizen (changes requested):
@xiaoyawei I'm on board with adding this functionality in, it does match the protocol spec, but some concerns with the specific implementation, see comments.

Please let me know if you are interested in working on this within ~1 month, I know it's a very old PR. Else I might push my suggestions on top since I'm anyway planning a breaking release soon.

----

General feedback:

This is a breaking change to a public type (which I see no good way to avoid), so let's make the commit message and pr description something like:
```
refactor!: include grpc error message in tracing
```

We do need test coverage as well:
- happy path (percent-encoded value)
- invalid percent-encoding
- invalid utf-8
- unknown status code
- empty grpc-message header value

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -248,15 +276,17 @@ impl ClassifyEos for GrpcEosErrorsAsFailures {
...
     /// A gRPC response was classified as a failure with the corresponding status.
-    Code(std::num::NonZeroI32),
+    Status(GrpcStatus),
     /// A gRPC response was classified as an error with the corresponding error description.
     Error(String),
 }
 
 impl fmt::Display for GrpcFailureClass {
```
jlizen, line 315: Let's also impl `std::error::Error` for `GrpcFailureClass` while we are anyway touching it

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -277,28 +307,53 @@ pub(crate) fn classify_grpc_metadata(
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct GrpcStatus {
+    code: GrpcCode,
+    message: Option<String>,
+}
+
+impl GrpcStatus {
+    pub(crate) fn code(&self) -> GrpcCode {
```
jlizen: Let's promote this from `pub(crate)` to `pub`, it seems useful

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -277,28 +307,53 @@ pub(crate) fn classify_grpc_metadata(
...
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct GrpcStatus {
+    code: GrpcCode,
+    message: Option<String>,
+}
+
+impl GrpcStatus {
```
jlizen, line 376: Let's also add a `pub fn message()-> Option<&str>` to avoid forcing usage of the format machinery to peek the message

xiaoyawei:
@jlizen Thanks for the review. I will try to address your comments and push something in the next a couple of days

xiaoyawei:
Hey @jlizen , would you take another look when you have time?

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -310,28 +340,82 @@ pub(crate) fn classify_grpc_metadata(
...
+    #[cfg(feature = "trace")]
+    let message = headers.get("grpc-message").map(|header| {
+        percent_decode(header.as_bytes())
+            .decode_utf8_lossy()
+            .into_owned()
+    });
+    #[cfg(not(feature = "trace"))]
+    let message = headers
+        .get("grpc-message")
+        .and_then(|header| header.to_str().ok().map(|s| s.to_owned()));
```
jlizen: this path also needs the `.decode_utf_lossy()` handling

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -386,8 +492,78 @@ mod tests {
...
+            code: Some(GrpcCode::Unknown),
+            code_raw: 2,
+            message: Some("hello world".to_string()),
+        }),
+    }
+
+    classify_grpc_metadata_test! {
+        name: invalid_percent_encoding,
+        status: "13",
+        message: "bad%2Gencode",
```
jlizen, line 513: this is good, but let's also have a similar test that is valid utf-8

tower-http/src/classify/grpc_errors_as_failures.rs
```diff
@@ -310,28 +340,82 @@ pub(crate) fn classify_grpc_metadata(
...
-        ParsedGrpcStatus::Success
-    } else {
-        ParsedGrpcStatus::NonSuccess(NonZeroI32::new(status).unwrap())
+    if let Some(code) = grpc_code {
+        if success_codes.contains(GrpcCodeBitmask::from(code)) {
+            return ParsedGrpcStatus::Success;
+        }
+    }
+
+    #[cfg(feature = "trace")]
```
jlizen: As best as i can tell this split only exists to avoid depending on `percent-encoding` for non-trace, but it's a tiny no-dependencies dependency, I probably wouldn't bother with it in favor of simplifying.

tower-http/src/classify/grpc_errors_as_failures.rs
jlizen, line 416: Let's also have a test probing that `GrpcCode::Ok` (code 0) is classified as a success

jlizen (changes requested):
The refactor looks great, a few more small things, then it's ready to merge.

xiaoyawei:
@jlizen See how your comments are addressed in the 3rd commit of this PR

jlizen (approved):
Looks great! Thanks for the contribution.

Aiming to get this released within the next ~2 weeks.

Issue #423, opened 2023-10-17
A means to propagate state between the layers that run before the service function, and those that run after.


bassmanitram:
## Feature Request
A middleware to be used as a way of sharing state between "pre service" layers/filters etc and those that run "post service", particularly when the service itself is unaware of that state.

### Motivation
The concrete motivation in my case is the "as lite-a-touch-as-possible" approach to wrapping AWS Lambda functions (eventually in any language, but initially in Rust) with layers that enforce our enterprise policy and aid in homogenizing the interface that the Lambda function is written to - specifically for AWS API Gateway API invocations.

These layers examine the request and need to set various state values that are then used by the layers that run after the Lambda function has returned. The problem is, of course, that the Lambda function consumes the request and produces a response.

The `propagate_header` middleware is addressing a very similar use case for headers.

### Proposal
Effectively, copy the `propagate_header` middleware and change its semantics to be those of http::Request  and Response extensions!

### Alternatives
There are of course many ways that Layer developers might be able to propagate such state when necessary - but they all rely on the same need - to save the state off "somewhere" when invoking the request/response service and retrieve when the response is ready. This scheme of exploiting the request and response extensions mechanism within a propagation middleware is an extremely convenient way to achieve this in a generic fashion.

I am in the process of creating this middleware - see [PR 424](https://github.com/tower-rs/tower-http/pull/424).

bassmanitram:
There appears to be no way for me to change one of my PRs from Draft to Ready, so I am closing this and opening another.

PR #424, opened 2023-10-17
Add propagate_extensions middleware

bassmanitram:
(See issue [423](https://github.com/tower-rs/tower-http/issues/423))

This middleware is intended to be used as a way of sharing state between "pre service" layers/filters etc and those that run "post service", particularly when the service itself is unaware of that state.

This is in fact, almost a verbatim copy of propagate_header, replacing the Header semantics with those required by http::Extensions.

For ease of implementation it _removes_ the state from the request before the inner service is invoked, and inserts the state into the response when the response is ready.

It does not concern itself with the initial insert of the state into the request since that is covered by an existing middleware, and may also be achieved by other means in one of the "pre service" components.

Because Extensions are managed by _type_, both the layer and the underlying service have to be generic over the type of the user's state object - and that type has to be cloneable as well as Send and Sync. In order to keep the compiler happy, both the service and the layer have a phantom data field. I am not 100% sure that this is necessary, but it does seem to be one of the few safe code instances where it probably is.

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -68,6 +68,7 @@ full = [
     "map-response-body",
     "metrics",
     "normalize-path",
+    "propagate-extension",
     "propagate-header",
     "redirect",
     "request-id",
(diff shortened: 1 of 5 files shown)


bassmanitram:
Bad branching practices makes two independent middlewares that I am working one interdependent.

PR #348, opened 2023-03-24
Upgrade to http-body 1.0

davidpdrsn:
Still some things to do but should allow https://github.com/tokio-rs/axum/pull/1882 to move forward.

## TODO

- [x] Publish new version of http-body that has `Frame::map_data`
- [x] Are we okay with having http-body-util in our public API? We have things like https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimit.html#impl-Service%3CRequest%3CReqBody%3E%3E-for-RequestBodyLimit%3CS%3E which includes `Limited`. Should we create wrappers?
- [x] Update (de)compression, but how?
- [x] Update docs
- [x] Update examples. This is blocked on axum, tonic, and warp supporting hyper 1.0. Examples are commented out for now until axum, tonic, etc have been updated.
- [x] Changelog

PR #434, opened 2023-11-11
Refactor use of futures crate

tottoto:
## Motivation

Refactoring.

## Solution

- Replaces `futures_{util,core}` with `std`'s APIs when possible.
- Uses `futures_core`'s API via `futures_util` as already used at the other places.
- Replaces `futures` with `futures_util`.
- Makes `futures_util` dependency optional.

### DIFF
--- a/tower-http/Cargo.toml
+++ b/tower-http/Cargo.toml
@@ -15,8 +15,6 @@ rust-version = "1.66"
 [dependencies]
 bitflags = "2.0.2"
 bytes = "1"
-futures-core = "0.3"
-futures-util = { version = "0.3.14", default_features = false, features = [] }
 http = "0.2.7"
 http-body = "0.4.5"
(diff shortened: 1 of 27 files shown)


jplatte:
Hm, I'm unsure whether we should really remove the `futures` crate usage from examples. Could you split that change into a separate PR or open an issue for discussion? Then I'd be happy to merge the remaining changes.

tottoto:
Fixed.

getong:
http 1.0 is released, how about upgrading to http 1.0?

jplatte:
> http 1.0 is released, how about upgrading to http 1.0?

That's what #348 does. Why are you asking on this PR?

getong:
sorry.

jplatte (approved):
Thanks!

jplatte:
@getong no worries, I was just a bit confused by it.

PR #435, opened 2023-11-18
Replace futures with futures-util in examples dependency

tottoto:
## Motivation

Simplifies examples dependency.

## Solution

Replaces `futures` with `futures-util` in the examples dependency.


jplatte:
I wouldn't really say it's simpler, but it might improve CI times a little bit.

@davidpdrsn wdyt?

jplatte (approved):

PR #437, opened 2023-11-21
Release 0.5.0

davidpdrsn:
🚀


davidpdrsn:
> but maybe that's not that easy anymore

It's not super easy unfortunately but I can port it. Have written the necessary adapters elsewhere so mostly just copying those over.

davidpdrsn:
Done https://github.com/jplatte/caniuse.rs/pull/78!

jplatte:
Thanks a bunch, it works so I would say we're definitely good to go now! 🚢

PR #441, opened 2023-11-22
Reduce futures-util dependency

### LINKED ISSUE #436: Minimize futures crates dependency
### Motivation

Allows users to depend on the minimum dependencies which are actually used.

### Proposal

Changes `tower-http` to depends on `futures-core` and `futures-util` when they are actually needed, especially about the latter. This includes porting `BoxFuture` to `tower-http`'s code. This proposal's drawback is increasing the maintenance cost to control the crates condition, and the profit can be affected by the other dependencies.

### Alternatives

Keeps depending on `futures-util`.

tottoto:
Resolves #436.


jplatte (approved):
Nice, thanks!

Issue #416, opened 2023-09-28
CompressionLayer breaks range requests, violating RFC 7233


and-reas-se:
### Version

tower-http v0.4.3

### Platform

Linux 6.5.2 x86_64

### Description

Say you have a large file served by a ServeDir and with a CompressionLayer thrown on top. ServeDir supports range requests. If the client requests to get a file starting at from bytes x to bytes y, and the response is compressed, this request should be interpreted as compressed bytes from x to y. But these two in combination will interpret it as uncompressed bytes x to y, so the wrong set of bytes are returned.

Furthermore the content-range header is set incorrectly in the response. It's also based on uncompressed bytes rather than compressed.

I ran in to this problem in a real life scenario. It intermittently breaks Microsofts Azure CDN (Front Door) which will sometimes use range requests when requesting files from the origin. 

The workaround that solved the problem for me was to write a small axum middleware that strips the range header from requests and accept-ranges header from the response (reproduced below).

```rust
use axum::{http::Request, middleware::Next, response::Response};

pub async fn remove_range<B: std::fmt::Debug>(mut req: Request<B>, next: Next<B>) -> Response {
    req.headers_mut().remove("range");
    let mut response = next.run(req).await;
    response.headers_mut().remove("accept-ranges");
    response
}
```

jplatte:
Right so the (de)compression middlewares need to either strip, or if possible adjust these headers. Makes sense.

jplatte:
Actually maybe the better solution is for the middlewares to disable themselves on such requests. Otherwise requesting the end of a huge file could lead to a full re-transmission for no reason.

jplatte:
Given David's approval of my previous comment, a PR implementing it would be appreciated.

and-reas-se:
What if a client starts downloading a large file using a non-range request, the connection drops partway trough, and then the client tries to resume the download using a range request? Wouldn't you get a corrupted file if the compression middlewares are enabled for the first request and disabled for the second?

jplatte:
Hm, I see what you mean. Really the whole range request for compressed data thing critically relies on the server having the full compressed content cached. That is annoying. Maybe the only solution is really what you wrote then. But there's another problem, if `ServeDir` supported compression itself, wrapping it in a compression layer should be a no-op, but as written will also filter out those headers 😕

seanmonstar:
I doubt this is the first time this has been figured out. Perhaps other server frameworks in other languages could show what to do.

PR #446, opened 2023-12-03
Skip compression for range requests

ryanfowler:
This PR  fixes broken range requests when using the compression middleware by skipping response body compression when the `Content-Range` header is present in the response. It also strips the `Accept-Ranges` response header when performing compression.

## Motivation

Fixes https://github.com/tower-rs/tower-http/issues/416

## Solution

I believe the correct behaviour for this library is to pass along the `Range` request header, and skip compression if the `Content-Range` response header is present. This will allow range requests if the wrapped services support them, and skip compression in that circumstance to avoid corruption. If the user would rather compress everything, they can disallow range requests by stripping out the `Range` request header in an earlier middleware.

Notably, this is how a prominent Go library handles range requests and compression:
- [skip compression when `Content-Range` header is present](https://github.com/klauspost/compress/blob/v1.17.4/gzhttp/compress.go#L122)
- [strip the `Accept-Ranges` header when compressing](https://github.com/klauspost/compress/blob/v1.17.4/gzhttp/compress.go#L190)


jplatte (commented):
Thanks! Can you add a changelog entry?

jplatte (approved):

PR #447, opened 2023-12-10
Add concurrency_limit middleware

zdenek-crha:
## Motivation

The tower::limit::ConcurrencyLimit middleware does not work properly when used with services that return response with streaming body. The middleware considers concurrency only from call to service to response future resolution.

But response with streaming body can't be considered finished until body has been consumed.

## Solution

Add concurrency_limit module with ConcurrencyLimit middleware implementation that holds on the semaphore permit until response and its body has been consumed.

It uses original middleware from `tower` crate and `tower-http::metrics::InFlightRequests` middleware as inspiration.

## Questions/Considerations

As far as module organization goes, it would be nicer to add this as `tower-http::limit::concurrency`, similar to the module layout in `tower` crate. But the `tower-http::limit` module is fully dedicated to `RequestBody` middleware now and moving it into sub-module would probably be breaking change.

I have considered leaving `RequestBody` middleware where it is and just add `tower-http::limit::concurrency` sub-module, but it felt confusing. Adding another top level module felt as cleanest alternative I had.

### DIFF
--- a/tower-http/src/concurrency_limit.rs
+++ b/tower-http/src/concurrency_limit.rs
@@ -1,0 +1,312 @@
+//! Limit the max number of concurrently processed requests.
+//!
+//! The service sets a maximum limit to the number of concurrently processed requests. The
+//! processing of a request starts when it is received by the service (`tower::Service::call` is
+//! called) and is considered complete when the response body is consumed, dropped, or an error
+//! happens.
+//!
+//! Internally, it uses semaphore to track and limit number of in-flight requests
+//!
+//! # Relation to `ConcurrencyLimit` from `tower` crate
+//!
+//! The `tower::limit::concurrency::ConcurrencyLimit` service uses a different definition of
+//! 'request processing'. It starts when request is received by `tower::Service::call`, and ends
+//! immediatelly after response is produced.
+//!
+//! In some cases it may not work properly with [`http::Response`], as it does not account for
+//! process of consuming response body.
+//!
+//! When stream is used as response body, the process of consumig it (ie streaming to called) may
+//! take longer and use more resources than just producing the response itself. And often it the
+//! number of streams we are processing concurrently we want to limit.
+//!
+//! The service version from [`tower-http`](crate) takes response body consumption into
+//! consideration and *will* limit number of concurrent streams correctly.
+//!
+//! ```
+//! use std::convert::Infallible;
+//! use bytes::Bytes;
+//! use http::{Request, Response};
+//! use http_body_util::Full;
+//! use tower::{Service, ServiceExt, ServiceBuilder};
+//! use tower_http::concurrency_limit::ConcurrencyLimitLayer;
+//!
+//! async fn handle(req: Request<Full<Bytes>>) -> Result<Response<Full<Bytes>>, Infallible> {
+//!     // ...
+//!     # Ok(Response::new(Full::default()))
+//! }
+//!
+//! # #[tokio::main]
+//! # async fn main() -> Result<(), Box<dyn std::error::Error>> {
+//! let mut service = ServiceBuilder::new()
+//!     // limit number of concurrent requests to 3
+//!     .layer(ConcurrencyLimitLayer::new(3))
+//!     .service_fn(handle);
+//!
+//! // Call the service.
+//! let response = service
+//!     .ready()
+//!     .await?
+//!     .call(Request::new(Full::default()))
+//!     .await?;
+//! # Ok(())
+//! # }
+//! ```
+//!
+
+use http::{Request, Response};
+use pin_project_lite::pin_project;
+use std::{
+    future::Future,
+    pin::Pin,
+    sync::Arc,
+    task::{ready, Context, Poll},
+};
+use tokio::sync::{OwnedSemaphorePermit, Semaphore};
+use tokio_util::sync::PollSemaphore;
+
+/// Limit max number of concurrent requests (per service)
+///
+/// The layer enforces a same concurrency limit for each inner service separately. In other words,
+/// [`ConcurrencyLimit`] middleware constructed from this layer for each service will have its own
+/// semaphore and will track and limit requests separately.
+///
+/// To track and limit multiple services together, see [`SharedConcurrencyLimitLayer`]
+///
+/// See the [module docs](crate::concurrency_limit) for more details.
+#[derive(Clone, Debug)]
+pub struct ConcurrencyLimitLayer {
+    max: usize,
+}
+
+impl ConcurrencyLimitLayer {
+    /// Create new [`ConcurrencyLimitLayer`] with semaphore size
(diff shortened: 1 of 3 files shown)


tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
+//! Limit the max number of concurrently processed requests.
```
jlizen, line 1: (totally optional): might be nice to mention that this works well with the new [OnEarlyDrop layer](https://docs.rs/tower-http/latest/tower_http/on_early_drop/index.html) in case the request is waiting for a while and the client gives up.

tower-http/src/lib.rs
jlizen, line 312: Is there a reason we couldn't locate this as a submodule within `limit` and just re-export its types? We also could reorganize the existing code using the `#[path]` annotation without breaking changes.

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+//! called) and is considered complete when the response body is consumed, dropped, or an error
+//! happens.
+//!
+//! Internally, it uses semaphore to track and limit number of in-flight requests
+//!
+//! # Relation to `ConcurrencyLimit` from `tower` crate
+//!
+//! The `tower::limit::concurrency::ConcurrencyLimit` service uses a different definition of
+//! 'request processing'. It starts when request is received by `tower::Service::call`, and ends
+//! immediatelly after response is produced.
```
jlizen, line 14:
```suggestion
//! immediately after response is produced.
```

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+//!
+//! In some cases it may not work properly with [`http::Response`], as it does not account for
+//! process of consuming response body.
+//!
+//! When stream is used as response body, the process of consumig it (ie streaming to called) may
```
jlizen, line 19:
```suggestion
//! When stream is used as response body, the process of consuming it (ie streaming to caller) may
```

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+//! take longer and use more resources than just producing the response itself. And often it the
```
jlizen, line 20:
```suggestion
//! take longer and use more resources than just producing the response itself. And, often, it is the
```

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+//! number of streams we are processing concurrently we want to limit.
```
jlizen, line 21:
```suggestion
//! number of streams we are processing concurrently that we want to limit.
```

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+
+impl<S> tower_layer::Layer<S> for SharedConcurrencyLimitLayer {
+    type Service = ConcurrencyLimit<S>;
+
+    fn layer(&self, service: S) -> Self::Service {
+        ConcurrencyLimit::new(service, self.semaphore.clone())
+    }
+}
+
+/// Middleware that limits max number fo concurrent in-flight requests.
```
jlizen, line 133:
```suggestion
/// Middleware that limits max number of concurrent in-flight requests.
```

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+/// To track and limit each service separately, see [`ConcurrencyLimitLayer`].
+///
+/// See the [module docs](crate::concurrency_limit) for more details.
+#[derive(Clone, Debug)]
+pub struct SharedConcurrencyLimitLayer {
+    semaphore: Arc<Semaphore>,
+}
+
+impl SharedConcurrencyLimitLayer {
+    /// Create new [`ConcurrencyLimitLayer`] with shared semaphore
```
jlizen, line 112:
```suggestion
    /// Create new [`SharedConcurrencyLimitLayer`] with a shared semaphore.
```

tower-http/src/concurrency_limit.rs
```diff
@@ -0,0 +1,312 @@
...
+        self.inner.is_end_stream()
+    }
+
+    #[inline]
+    fn size_hint(&self) -> http_body::SizeHint {
+        self.inner.size_hint()
+    }
+}
+
+#[cfg(test)]
```
jlizen, line 271: We should have tests for:
- `ConcurrencyLimitLayer` (on its own), probably including probing the unique semaphore management per cloned layer
- Probe the actual concurrency limiting behavior using eg 2 requests but a limit of 1
- Request that returns an error, validate that the permit is still released
- Body drop without consumption

jlizen (changes requested):
@zdenek-crha this is great! Sorry it took so long to get a review.

I'm on board with the approach but had some small tweaks.

Let me know if you no longer have time to work on this, no problem, I don't mind adding a revision on top.

zdenek-crha:
@jlizen Wow, I have completely forgot I authored this one :-)

I will be able to find some time to push it forward next week.

I don't use github PRs much, do you prefer another commit with fixes on top or adding the fixes into existing commit and force-push?

Since the PR is quite old, maybe even rebase and re-do it on current master?

jlizen:
Any of them is fine with me. If it were me, I would probably start fresh from main, and manually pull the changes back over, then force push on top. But whatever feels least annoying.

tower-http/src/lib.rs
zdenek-crha: I can do the work no problem, but I would appreciate guidance on to which shape to steer the code. My approach probably would be:

* move docs and private modules from `limit.rs` to public `limit/body_limit.rs`
* keep current re-exports in `limit.rs` for backward compatibility
* have same re-exports in `limit/body_limit.rs` to be consistent with to-be-added module

Then, in second commit:

* add `limit/concurrency_limit.rs`
* mimic the `limit/body_limit.rs` structure with private modules and re-exports for consistency

I like `limit/body.rs` and `limit/concurrency.rs` better because they are shorter and don't duplicate the 'limit' word in path. But having two, nested `body` modules is not nice, so would probably go with the longer variant.

With the original re-exports remaining in `limit.rs`, there should not be breaking change (but I may be missing something).

I have not used the `#[path]` attribute before. I've read the docs, but I'm not sure how it could be used here. (but I'm willing to learn :-D)

jlizen: Seems fine to me.

> move docs and private modules from limit.rs to public limit/body_limit.rs

Specifically we should maintain the same visibility of types from a consumer perspective.

> I like limit/body.rs and limit/concurrency.rs better because they are shorter and don't duplicate the 'limit' word in path. But having two, nested body modules is not nice, so would probably go with the longer variant.

No strong opinion here.

PR #448, opened 2023-12-11
docs(example): migrate axum-key-value-store to axum 0.7

Oliboy50:
## Motivation
https://github.com/tower-rs/tower-http/pull/348#discussion_r1409212369

## Solution
Restore axum-key-value-store example with axum 0.7


jplatte (approved):
Thanks!

PR #450, opened 2023-12-20
support servefile zstd

randomairborne:
## Motivation
ServeDir supports precompressed zstd files, but ServeFile does not

## Solution
Copy-and-paste enabling code to zstd from deflate


Nehliin (commented):
Thanks, please add a test similar to the other precompressed algorithms and then we can merge this

randomairborne:
alright this should be ready now. Thanks.

Nehliin (approved):

Issue #420, opened 2023-10-06
CompressionLayer causes event streams to buffer, breaking them


ekzhang:
## Bug Report

When using [Server-Sent Events (SSE)](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events), a streaming body is sent in real time with type `text/event-stream`. This is a standard MIME type for many websites, such as notification feeds, or streaming chat completion APIs.

However, if you apply the default `CompressionLayer` on a site with SSE, it will cause all of the server-sent events to buffer until reaching a given size, before they are able to reach the HTTP client.

I `strace`d a server running `hyper` over HTTP/1.1, with an event stream outputting messages every 1 second, and it does still write chunks of data to the socket every second. However, the web browser (as well as `curl --no-buffer --compressed`) is not able to decode the messages individually.

I also confirmed this behavior with other streaming middleware implementations, such as FastAPI's [`GZipMiddleware`](https://fastapi.tiangolo.com/advanced/middleware/#gzipmiddleware). It seems that compression encoding generally doesn't work nicely with `text/event-stream`.

### Version

`0.4.4`

### Platform

```
Linux ip-172-31-47-18 5.15.0-1045-aws #50~20.04.1-Ubuntu SMP Wed Sep 6 17:29:11 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

### Crates

`compression` module with `compression-gzip` feature

### Description

I wanted to ask if it's possible to include

```rust
NotForContentType::const_new("text/event-stream")
```

in the `DefaultPredicate` for `tower_http::compression::CompressionLayer`. This would reduce the chance of an issue like this happening for application developers, who commonly use SSE for live event feeds.

Thank you so much for your work on this library.

BenJeau:
Addressed as part of https://github.com/tower-rs/tower-http/pull/465

jplatte:
And shippped in 0.5.3.

PR #465, opened 2024-02-11
feat: add SSE compression exclusion within default predicate

BenJeau:
## Motivation

Discussed in this Github discussion from Axum's repo:
* https://github.com/tokio-rs/axum/discussions/2034#discussioncomment-8432073

And had opened a docs PR within Axum's repo to add a note about compression for SSEs:
* https://github.com/tokio-rs/axum/pull/2585

If not careful or properly handled, general compression with SSE will result in missing some or all events.

## Solution

Exclude SSEs from being compressed by checking the content type of `text/event-stream` within the `DefaultPredicate`

### DIFF
--- a/tower-http/src/compression/predicate.rs
+++ b/tower-http/src/compression/predicate.rs
@@ -90,6 +90,7 @@ where
 ///
 /// - They're gRPC, which has its own protocol specific compression scheme.
 /// - It's an image as determined by the `content-type` starting with `image/`.
+/// - They're Server-Sent Events (SSE) as determined by the `content-type` being `text/event-stream`.
 /// - The response is less than 32 bytes.
 ///
 /// # Configuring the defaults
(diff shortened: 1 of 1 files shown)


FrankReh:
Could the problem have been that the compression buffer isn't flushed for the SSE case?

BenJeau:
> Could the problem have been that the compression buffer isn't flushed for the SSE case?

Potentially, I didn't had the time to investigate but saw that disabling compression resolved all my issues.

---

I can't close the linked issue https://github.com/tower-rs/tower-http/issues/420 - if someone could, it would be great to cleanup the issues

jplatte:
I'll try to remember to close it after the fix has shipped in a crates.io release.

ekzhang:
Thanks for fixing this!

Besides text/event-stream, another candidate content type that involves dynamically pushing data is multipart/x-mixed-replace -- which is used sometimes in image tags: https://wiki.tcl-lang.org/page/multipart%2Fx-mixed-replace#:~:text=multipart%2Fx%2Dmixed%2Dreplace%20is%20an%20HTTP%20content%20type,content%20to%20the%20web%20browser.

(I think that one is a lot less common though)

PR #468, opened 2024-02-14
Fix Deny Action failures in CI

privettoli:
- Fixes deny action CI failures like in PR #467.

## Motivation
Fixes deny action CI failures like in PR #467.

## Solution
- Specify manifest-path as a separate input property instead of in `arguments`
- Removes `arguments` because the new value [is the default](https://github.com/EmbarkStudios/cargo-deny-action/blob/68cd9c5e3e16328a430a37c743167572e3243e7e/action.yml#L17)


jplatte (approved):
Thanks!

PR #473, opened 2024-02-22
enable cors/tests.rs and make it pass

GlenDC:
## Motivation

`cors/tests.rs` was added somewhere between now and the `0.5` release.
However the file wasn't enabled, and this seemed to also hide a bug,
where we did not in fact preserve the already set `Vary` (http response) header.

Found while porting over the updated/added logic since 0.5 to <https://github.com/plabayo/rama>

(as that codebase uses ported code from `tower-async`, which is ported from `tower`, 
  since then I have abandoned `tower-async` as it did its job to proof that it was possible, but more drastic changes
  were desired on my side for `rama`. I do keep however `rama` in sync with improvements made to codebases such as `tower-http`, which is one of many reasons why I love to contribute to `tower` where I can and where it is desired)

## Solution

1. enable the test by adding the mod to `cors/mod.rs`
2. fix the bug in `cors/mod.rs` where we remove from `headers` instead of  `response_headers`

As flagged on Discord it does also mean we need a third change where the order is not preserved,
and instead the custom (already set) Vary header will be at the end.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -14,7 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 
 - **compression:** Skip compression for range requests ([#446])
-- **cors:** don't overwrite previously-set (Vary) value ([#473])
+- **cors:** *Actually* keep Vary headers set by the inner service when setting response headers ([#473])
+  - Version 0.5.1 intended to ship this, but the implementation was buggy and didn't actually do anything
 
 [#399]: https://github.com/tower-rs/tower-http/pull/399
 [#446]: https://github.com/tower-rs/tower-http/pull/446
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -14,9 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 
 - **compression:** Skip compression for range requests ([#446])
+- **cors:** don't overwrite previously-set (Vary) value ([#473])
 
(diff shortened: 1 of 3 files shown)


tower-http/src/cors/mod.rs
```diff
@@ -682,13 +685,15 @@ where
             KindProj::CorsCall { future, headers } => {
                 let mut response: Response<B> = ready!(future.poll(cx))?;
 
+                let response_headers = response.headers_mut();
+
                 // vary header can have multiple values, don't overwrite
                 // previously-set value(s).
-                if let Some(vary) = headers.remove(header::VARY) {
+                if let Some(vary) = response_headers.remove(header::VARY) {
                     headers.append(header::VARY, vary);
```
jplatte, line 693: Aha, I see the problem now! This was supposed to be the actual response's headers. Can you revert the other changes in this fn and in the test and change this to append to `response`s headers?

jplatte (commented):
Thanks for the PR! In addition to the change requested below, can you also add a changelog entry?

tower-http/src/cors/mod.rs
GlenDC: Done ✅

tower-http/CHANGELOG.md
```diff
@@ -14,9 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Fixed
 
 - **compression:** Skip compression for range requests ([#446])
+- **cors:** don't overwrite previously-set (Vary) value ([#473])
```
jplatte, line 17:
```suggestion
- **cors:** *Actually* keep Vary headers set by the inner service when setting response headers ([#473])
  - Version 0.5.1 intended to ship this, but the implementation was buggy and didn't actually do anything
```

PR #477, opened 2024-03-05
Make http-body and http-body-util dependencies optional

tottoto:
## Motivation

Some features do not seem to use `http-body` crate, and some of them also not using `http-body-util` crate as the dependencies.

## Solution

Makes `http-body` and `http-body-util` dependencies optional.

This is a breaking change as this disables `body` module conditionally,


jplatte:
Thanks for the PR! Can you add a `# Unreleased` section to `tower-http/CHANGELOG.md` and describe the changes (esp. the breaking one) there?

tottoto:
Added the changelog.

jplatte (approved):
Great, thanks!

PR #479, opened 2024-03-09
Relax body trait bound

tottoto:
## Motivation

`Body` trait bound does not seem to be needed in these response body type.

## Solution

Removes `Body` trait requirement from these type parameters.


jplatte (approved):
Thanks!

Issue #365, opened 2023-05-03
Rewrite tracing middleware


jplatte:
There are a few problems with the current tracing middleware:

- It is overly generic, making it hard to grok
- The body type it uses doesn't (ever) implement `Default`, but other middleware requires this bound

It would be nice if we could come up with a different tracing middleware that fulfills most or all of the current ones' use cases without sharing these problems.

Oliboy50:
> The body type it uses doesn't (ever) implement Default, but other middleware requires this bound

Can you elaborate on how it impacts users, please?
Because we use this layer along with other ones from this repository and have no issue so far.

jplatte:
I've taken a first step in #380, which only simplifies the implementation, not the interface.

Having looked at most of the code though, one piece in particular that stands out as really complex / abstract is the `classify` API. I have never used it and wonder if it's worth keeping it at all. @davidpdrsn how much of this are you using, and are you aware of others using it?

PR #483, opened 2024-04-02
trace: adds back call to classify_eos on trailers

markdingram:
- was dropped in the http-body 1.0 upgrade

## Motivation

We have some tracing middleware that started as a copy of <https://github.com/EmbarkStudios/server-framework/blob/main/src/middleware/trace.rs>

After upgrading to tower-http 0.5.X one of the tests is failing -

https://github.com/EmbarkStudios/server-framework/blob/1bae7fe7a417443b365d2452007adfe46910037b/src/middleware/trace.rs#L205-L215

```
json atoms at path ".span.otel.status_code" are not equal:
    expected:
        "ERROR"
    actual:
        "OK"
```

## Solution

Adds back call to classify_eos on trailers. 

I've proven with this change the downstream tests pass once more. Notably I haven't attempted to write any tests in this repo - there was a historic desire to rewrite this middleware expessed in <https://github.com/tower-rs/tower-http/issues/365> - I would be amenable to spend some time to upstream some of the tests from <https://github.com/EmbarkStudios/server-framework/blob/main/src/middleware/trace.rs> if considered worthwhile by the maintainers.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The implicit `async-compression` feature is removed (BREAKING) ([#642])
 - The implicit `tokio` feature is removed (BREAKING) ([#628])
 
+## Fixed
+
+- `trace`: restore failure classification at end-of-stream ([#483])
+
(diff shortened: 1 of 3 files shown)


jlizen:
Thanks. Also restored `classify_eos(None)` for the trailer-less end-of-stream case.

jlizen (approved):

PR #486, opened 2024-04-27
Add `redirect_path_prefix` option

### LINKED ISSUE #413: `ServeDir` does not redirect correctly for directories without trailing `/` when nested
### Version
`tower-http v0.4.4`

### Platform
All

### Description
This function redirects paths that don't end with `/` and is a directory to the URI with `/` appended
https://github.com/tower-rs/tower-http/blob/e8eb54966604ea7fa574a2a25e55232f5cfe675b/tower-http/src/services/fs/serve_dir/open_file.rs#L253

However it does not take into account where the service could be nested, therefore producing an invalid redirect.

Example directory:
```
src/
├ main.rs
└ hello/
  └ index.html
```
main.rs:
```rust
use axum::{Router, Server};

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt::init();

    let serve_dir = ServeDir::new("src");
    let router = Router::new().nest_service("/src", serve_dir);

    let addr = "0.0.0.0:3000".parse().unwrap();

    Server::bind(&addr)
        .serve(router.into_make_service())
        .await
        .unwrap();

}

```

Going to `localhost:3000/src/hello` to see the issue
Expected: Redirects to `localhost:3000/src/hello/` and returns content at `src/hello/index.html`
Actual: Redirects to `localhost:3000/hello/` and returns Not Found

claraphyll:
## Motivation

Currently `ServeDir` assumes that the files are served from `/`, which breaks trailing slash redirects in some configurations like a reverse proxy that redirects `/somePath/*` to the `ServeDir` server or a scenario where `axum::Router::nest_service` is used.

## Solution

This allows the user to pass in a string that is prepended to the redirect path.

edit(jplatte): Closes #413

### DIFF
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -52,6 +52,7 @@ const DEFAULT_CAPACITY: usize = 65536;
 #[derive(Clone, Debug)]
 pub struct ServeDir<F = DefaultServeDirFallback> {
     base: PathBuf,
+    redirect_path_prefix: String,
     buf_chunk_size: usize,
     precompressed_variants: Option<PrecompressedVariants>,
     // This is used to specialize implementation for
@@ -72,6 +73,7 @@ impl ServeDir<DefaultServeDirFallback> {
 
         Self {
             base,
+            redirect_path_prefix: String::new(),
             buf_chunk_size: DEFAULT_CAPACITY,
             precompressed_variants: None,
             variant: ServeVariant::Directory {
@@ -89,6 +91,7 @@ impl ServeDir<DefaultServeDirFallback> {
     {
         Self {
             base: path.as_ref().to_owned(),
+            redirect_path_prefix: String::new(),
             buf_chunk_size: DEFAULT_CAPACITY,
             precompressed_variants: None,
             variant: ServeVariant::SingleFile { mime },
@@ -131,6 +134,23 @@ impl<F> ServeDir<F> {
             }
             ServeVariant::SingleFile { mime: _ } => self,
         }
+    }
+
+    /// Sets a path to be prepended when performing a trailing slash redirect.
+    ///
+    /// This is useful when you want to serve the files at another location than `/`, for example
+    /// when you are using multiple services and want this instance to handle `/static/<path>`.
+    /// In that example, you should pass in `/static` so that a trailing slash redirect does not
+    /// redirect to `/<path>/` but instead to `/static/<path>/`
+    ///
+    /// The default is the empty string.
+    ///
+    /// A trailing slash is removed from the prefix.
+    pub fn redirect_path_prefix(mut self, prefix: impl Into<String>) -> Self {
+        let mut prefix = prefix.into();
+        prefix = prefix.trim_end_matches('/').to_owned();
+        self.redirect_path_prefix = prefix;
+        self
     }
 
     /// Set a specific read buffer chunk size.
@@ -229,6 +249,7 @@ impl<F> ServeDir<F> {
     /// ```
     pub fn fallback<F2>(self, new_fallback: F2) -> ServeDir<F2> {
         ServeDir {
+            redirect_path_prefix: self.redirect_path_prefix,
             base: self.base,
             buf_chunk_size: self.buf_chunk_size,
             precompressed_variants: self.precompressed_variants,
@@ -377,6 +398,8 @@ impl<F> ServeDir<F> {
             }
         };
 
+        let redirect_path_prefix = self.redirect_path_prefix.clone();
+
         let buf_chunk_size = self.buf_chunk_size;
         let range_header = req
             .headers()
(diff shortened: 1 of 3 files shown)


jplatte:
Hey, thanks for the PR! I think we should have something like this (as indicated in my last comment on #413), but.. naming is hard. I've been considering something like `.prefix("/static", PrefixMode::PrependOnRedirect)` (where the other `PrefixMode` would be to strip the prefix on incoming requests before filesystem lookup _and_ prepend it on redirects), but this would require users to import that enum, and I still don't know what name to give the other prefix mode 😄

I guess since axum `.nest()` is the most important use case and it's easy enough to implement something like it elsewhere, we could start with just this "mode", i.e. a function like what you added but with a clearer name.

So if you don't care much about the naming or what exactly the API looks like, I would recommend you rename `prepend_path` to `redirect_path_prefix`. Otherwise, I'm curious to hear your thoughts!

maxi0604:
Thanks for the review

> I guess since axum `.nest()` is the most important use case and it's easy enough to implement something like it elsewhere, we could start with just this "mode", i.e. a function like what you added but with a clearer name.
I think this one solves the more pressing issue with the redirects.

You could use a workaround like putting the files at `base-folder/prefix/...` if you are (somehow) getting requests for paths with the prefix; but as suggested in the issue, fixing the trailing slash redirects would require intercepting the response. 

A separate option to strip an arbitrary prefix from the file path could also be added if that workaround doesn't work for someone. (Maybe `strip_path_prefix`) This would have to be explained in the documentation but ultimately provides more flexibility (With the two options combined, you could replace a prefix with another one.)

> So if you don't care much about the naming or what exactly the API looks like, I would recommend you rename `prepend_path` to `redirect_path_prefix`. Otherwise, I'm curious to hear your thoughts!

I've renamed it to `redirect_path_prefix`, that does sound like a more reasonable name. I think this already solves an immediate issue while more flexibility could be added later.

maxi0604:
I don't know why the cargo-hack test is failing, it seems unrelated to me.

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -211,6 +227,7 @@ impl<F> ServeDir<F> {
     /// ```
     pub fn fallback<F2>(self, new_fallback: F2) -> ServeDir<F2> {
         ServeDir {
+            redirect_path_prefix: "".to_string(),
```
jlizen: This should forward the existing value rather than discarding it

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -267,8 +270,10 @@ async fn maybe_redirect_or_append_path(
         path_to_file.push("index.html");
         None
     } else {
-        let location =
-            HeaderValue::from_str(&append_slash_on_path(uri.clone()).to_string()).unwrap();
+        let location_string = format!("{prepend_path}{}", append_slash_on_path(uri.clone()));
+
+        let location = HeaderValue::from_str(&location_string).unwrap();
```
jlizen: We shouldn't panic here. The uri unwrapped in existing code but was already validated, we should instead propagate the error. Since this PR main has refactored to use `OpenFileOutput::InvalidRedirectUri` if you rebase.

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -115,6 +118,19 @@ impl<F> ServeDir<F> {
...
 
+    /// Sets a path to be prepended when performing a trailing slash redirect.
+    ///
+    /// This is useful when you want to serve the files at another location than "/", for example
+    /// when you are using multiple services and want this instance to handle `/static/<path>`.
+    /// In that example, you should pass in "/static" so that a trailing slash redirect does not
+    /// redirect to `/<path>/` but instead to `/static/<path>/`
+    ///
+    /// The default is the empty string.
+    pub fn redirect_path_prefix(mut self, path: String) -> Self {
```
jlizen: This should accept `impl Into<String>`

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -115,6 +118,19 @@ impl<F> ServeDir<F> {
         }
     }
 
+    /// Sets a path to be prepended when performing a trailing slash redirect.
+    ///
+    /// This is useful when you want to serve the files at another location than "/", for example
+    /// when you are using multiple services and want this instance to handle `/static/<path>`.
+    /// In that example, you should pass in "/static" so that a trailing slash redirect does not
```
jlizen:
```suggestion
    /// In that example, you should pass in `/static` so that a trailing slash redirect does not
```

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -115,6 +118,19 @@ impl<F> ServeDir<F> {
         }
     }
 
+    /// Sets a path to be prepended when performing a trailing slash redirect.
+    ///
+    /// This is useful when you want to serve the files at another location than "/", for example
```
jlizen:
```suggestion
    /// This is useful when you want to serve the files at another location than `/`, for example
```

tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -384,6 +384,19 @@ async fn redirect_to_trailing_slash_on_dir() {
     assert_eq!(location, "/src/");
 }
 
+#[tokio::test]
```
jlizen, line 450: We should also test:
- `redirect_path_prefix` + `fallback()`
- prefix with trailing slash
- empty prefix (or should we reject that during validation?)
- combination with query parameters
- combination with percent-encoded path

jlizen (changes requested):
@claraphyll sorry for the extended delay in the review...

The new naming and API seems fine to me. Glad to approve this with some small changes, and it needs a rebase.

Or if you no longer have time to work on it, no problem, I am happy to finish it up.

jlizen (approved):
Thanks!

samanpa:
Can a new release be made with this feature

jlizen:
Preparing a release, just waiting on a few more PRs.

PR #492, opened 2024-05-25
Make Sealed trait pass cargo-hack CI

AlyoshaVasilieva:
This fixes the cargo-hack CI step by allowing the `Sealed` trait to be unused.

Alternative solution: Feature-gating `mod sealed` to only compile when the feature `util` is enabled.

I did it this way to avoid needing to add more feature gates if other code wants to rely on `Sealed`, but I can change it to `#[cfg(feature = "util")]` if that's preferred.


jplatte (approved):
Oh, really that simple? Nice!

PR #493, opened 2024-06-11
remove unnecessary sync constraint from async_predicate

PoOnesNerfect:
This `Sync` constraint on the future made it really hard to use for the real use cases. Plus it doesn't seem necessary.


jplatte (approved):
Thanks!

PR #496, opened 2024-07-05
Fix possible typos in comments

cpick:
I believe "both" is extraneous in these two comments and it tripped me up a bit the first time I was trying to understand the different layers.

## Motivation

Make the comments more clear.

## Solution

Remove the words.


jplatte (approved):
Yeah, looks like a fail from copy-pasting `fn sensitive_headers`' docs. Thanks for the fix!

PR #499, opened 2024-07-22
Upgrade base64 and brotli

xmakro:


nyurik:
Friendly ping to @davidpdrsn and @jplatte - Brotli dep before 6 has some conflicts with prior versions, please merge and release when you have a sec. Thx :)

jplatte:
What do you mean regarding conflicts? The brotli dep being upgraded here is just a dev-dependency (and base64 is not part of the public API). I'll merge anyways, but not sure when it will be released.

jplatte (approved):

nyurik:
thx! I guess it is not that big of a deal with dev dependency, although it could still fail:  if some of your dependencies use brotli internally, and they use a different version, cargo will try to build with both brotli versions together - which will fail because brotli uses FFI and makes them public - so linker would see a conflict of two different functions being exported under the same name.  Brotli v6 does not automatically publish ffi, so it is pure rust only, and won't conflict with previous version

PR #503, opened 2024-08-13
Bump tower to v0.5

paolobarbolini:
I see `tower` 0.5 came out but `tower-http` hadn't upgraded yet. I've fixed all compilation issues and now it seems to work.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -5,13 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-# Unreleased
+# 0.6.0
 
 ## Changed:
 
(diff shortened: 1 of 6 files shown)


jplatte (approved):
Thanks! Looks like this is a breaking change, as tower is included in `package.metadata.cargo-public-api-crates.allowed` in `Cargo.toml`.

Mind adding an entry to `tower-http/CHANGELOG.md`?

paolobarbolini:
Done

tower-http/CHANGELOG.md
```diff
@@ -5,13 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-# Unreleased
+# 0.6.0
```
jplatte, line 8: I don't think this is usually changed before release. I guess it's a good reminder that there's breaking changes, but the fact it's not out yet should also still be included. How about this?

```suggestion
# 0.6.0 (unreleased)
```

PR #506, opened 2024-08-30
add From<i32> impl for GrpcCode

### LINKED ISSUE #505: Implement `From<NonZeroI32>` for `GrpcCode`

## Feature Request

`GrpcCode` does not provide the required `From<T>` implementations to easily be used over a `tonic` server that returns a `tonic::Status` with a `tonic::Code`. These logically are the same concept, but cannot be converted between these enums (and the numerical representaions) except by implementing your own matching logic.

### Motivation

We are implementing a `GrpcErrorsAsFailures` for a tracing layer, and this returns a `GrpcFailureClass::Code(NonZeroI32)`. It would be nice to have a simple path to convert this to a `GrpcCode` so we can leverage the 

### Proposal

Similarly, if not identical, to [`impl From<i32> for tonic::Code`](https://docs.rs/tonic/latest/src/tonic/status.rs.html#844-868).

### Alternatives

The alternative is that users have to implement `from_i32` or similar as methods themselves, which is potentially more error-prone than using a standard and common implementation, especially if gRPC status codes are changed in the future (such as to add new enumerated values for additional statuses).

gshipilov:
This change adds a `From<i32>` impl for the GrpcCode enum. This allows users to easily convert the result of the `GrpcErrorsAsFailures` back into an enum.

Fixes: #505

## Motivation

I recently had to implement a custom `TraceLayer` on-failure struct, and it made sense to reuse the `GrpcErrorsAsFailures` class to extract the status. Unfortunately, it returns the error code as a `NoneZeroI32` but tower_http doesn't have a convenient way to map this back to the num.

Implementing this is trivial, but still feels like a reasonable addition to this library.

## Solution

I feel like this solution is pretty self-explanatory :).

### DIFF
--- a/tower-http/src/classify/grpc_errors_as_failures.rs
+++ b/tower-http/src/classify/grpc_errors_as_failures.rs
@@ -70,6 +70,32 @@ impl GrpcCode {
     }
 }
 
+impl From<i32> for GrpcCode {
+    fn from(value: i32) -> Self {
+        match value {
+            0 => GrpcCode::Ok,
+            1 => GrpcCode::Cancelled,
+            2 => GrpcCode::Unknown,
+            3 => GrpcCode::InvalidArgument,
+            4 => GrpcCode::DeadlineExceeded,
+            5 => GrpcCode::NotFound,
+            6 => GrpcCode::AlreadyExists,
+            7 => GrpcCode::PermissionDenied,
+            8 => GrpcCode::ResourceExhausted,
+            9 => GrpcCode::FailedPrecondition,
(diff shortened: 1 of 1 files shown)


jlizen (approved):
This seems fine to me! It matches `tonic`'s approach:   
-  https://github.com/hyperium/tonic/blob/3ab377b881b068a45995abab622f86b24588d81d/tonic/src/status.rs#L849-L871

(including the lossy fallback to Unknown)

Let me throw in a quick test and comment, and get this merged.

PR #507, opened 2024-09-03
fs: Fix serving precompressed files without an extension

### LINKED ISSUE #504: ServeDir doesn't serve files without extentions if pre-compression is enabled
### Version

```
tower-http v0.5.2
```

### Platform

```
Linux turing-machine 6.8.0-41-generic #41-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug  2 20:41:06 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

### Description

When using `ServeDir::precompressed_gzip` or similar methods, `ServeDir` won't serve files without a file extention if the client supports the relevant encoding.

Consider the following `axum` server:

```
use axum::Router;
use std::net::SocketAddr;

#[tokio::main]
async fn main() {
    let app = Router::new().fallback_service(ServeDir::new("assets/").precompressed_gzip());

    let addr = SocketAddr::from(([127, 0, 0, 1], 8000));
    let listener = tokio::net::TcpListener::bind(addr).await.unwrap();

    axum::serve(listener, app).await.unwrap();
}
```

I created the `assets/hello` file and tried to access it using curl:

```
$ curl -i http://localhost:8000/hello 
HTTP/1.1 200 OK
content-type: application/octet-stream
accept-ranges: bytes
last-modified: Thu, 29 Aug 2024 14:51:50 GMT
content-length: 14
date: Thu, 29 Aug 2024 15:04:58 GMT

Hello, world!
$ curl -i -H "Accept-Encoding: gzip" http://localhost:8000/hello 
HTTP/1.1 404 Not Found
content-length: 0
date: Thu, 29 Aug 2024 15:05:06 GMT

```

This happens whether `assets/hello.gz` exists or not.

### Speculation

There seems to be a bug in the `tower_http::services::fs::serve_dir::open_file::preferred_encoding` method.
I tried adding this test:

```
#[test]
fn preferred_encoding_without_extension() {
    let mut path = PathBuf::from("hello");
    preferred_encoding(&mut path, &[(Encoding::Gzip, QValue::one())]);
    assert_eq!(path, PathBuf::from("hello.gz"));
}
```

This test fails with the following panic:

```
thread 'services::fs::serve_dir::open_file::tests::preferred_encoding_without_extension' panicked at tower-http/src/services/fs/serve_dir/open_file.rs:342:9:
assertion `left == right` failed
  left: "hello..gz"
 right: "hello.gz"
```

SvizelPritula:
## Motivation

When trying to locate the precompressed version of a file that doesn't have a file extension (say `/downloads/foobar-cli`), the `ServeDir` service passes the relevant extension with a leading dot to `PathBuf::set_extension`, which doesn't expect a leading dot. This means it will end up with an filename that looks like this: `/downloads/foobar-cli..gz`, which probably doesn't exist. Worse still, it also fails to reconstruct the original path when it doesn't find anything at the corrupted path, which leads to a 404 error. 

## Solution

This change makes the function responsible for constructing the modified path directly modify the filename, instead of using the extension API, since an extension might not exist.

Fixes #504

I feel that #504 is clearly a bug, but someone might for some reason depend on the broken behavior somewhere.


tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -252,6 +252,50 @@ async fn missing_precompressed_variant_fallbacks_to_uncompressed_for_head_reques
...
+    assert_eq!(res.status(), StatusCode::OK);
+
+    assert_eq!(res.headers()["content-type"], "application/octet-stream");
+    assert_eq!(res.headers()["content-encoding"], "gzip");
+
+    let body = res.into_body().collect().await.unwrap().to_bytes();
+    let mut decoder = GzDecoder::new(&body[..]);
+    let mut decompressed = String::new();
+    decoder.read_to_string(&mut decompressed).unwrap();
+    assert!(decompressed.starts_with("Content."));
```
jplatte, line 275: Why assert it _starts with_ that? This seems like it should be an `assert_eq!` - maybe the string literal needs an extra `\n` in that case, or we could call `decompressed.trim()`.

SvizelPritula: I agree that this is a weird way to do this. I did it for consistency with the existing precompression tests, which do the same.

jplatte: Weird. No need to be consistent with that. I would welcome another PR to change the existing tests too, if you feel like it.

jplatte (commented):
One more thing, could you please add a changelog entry to `tower-http/CHANGELOG.md`?

tower-http/src/services/fs/serve_dir/tests.rs
SvizelPritula: I changed it to check the entire response body.

I'm afraid it will break on Windows if git is configured to change line endings to CRLF. This conversion will also only affect the uncompressed test files, which is annoying.

jplatte (approved):
Thanks!

tower-http/src/services/fs/serve_dir/tests.rs
SvizelPritula: I realized that I can just remove the trailing newline from the test file, so I did that. That makes it not a valid Unix text file, but there is no reason it should be.

SvizelPritula:
It seems I managed to add a commit to the PR between you approving it and you merging it. I didn't know that's possible and it seems dangerous.

jplatte:
I saw the new commit before pressing merge. I know this can be disabled, we do it at work, but I haven't seen this be a problem in all of the open source projects I work on. Especially because the GitHub web UI frequently craps out when you try to merge if a commit is pushed between opening the page and pressing merge 😂

PR #512, opened 2024-09-05
Add new SetTaskLocal layer / service

jplatte:
## Motivation

Allows easy state sharing via `tokio::task_local!`. Somewhat similar to `AddExtensionLayer`.

## Solution

Add `SetTaskLocalLayer` and `SetTaskLocal` service.


ikrivosheev:
@jplatte maybe can you move this layer to `tower`? I don't see specific things for http.

jplatte:
You're right. Though I don't currently have the motivation to figure out the detail on why my initial implementation didn't work out. Anyways, if that changes I'll make a PR to tower.

PR #514, opened 2024-09-08
Fix -Zminimal-versions build

paolobarbolini:
## Motivation

Building with `-Zminimal-versions` fails because `tower_layer::Identity::new` is not const.

## Solution

Bumps the minimum version requirement for `tower-layer` to the first version that had it const.


jplatte (approved):

jplatte:
Looks like we have another MSRV issue..

PR #517, opened 2024-09-11
Fix CI

jplatte:
## Motivation

tokio-util v0.7.12 has bumped its MSRV, and CI is failing.
`cargo public-api-crates` is also failing due to rustdoc JSON changes.

## Solution

Have the MSRV job downgrade tokio-util, pin a nightly for public-api-crates. Also upgrade the checkout action, bc. why not.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -114,7 +114,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@nightly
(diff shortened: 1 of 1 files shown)


seanmonstar:
Just noting that what I've started to do in other projects is just to pin the version in the MSRV CI job. It's what anyone who really needed that support could do, and I still know when the direct crate finally needs a newer version.

jplatte:
I feel like 1.70 is old enough that nobody is likely relying on something older than it (and still pulling in updates like the upcoming `tower-http 0.6.0`). Do you think there would be value in keeping a lower MSRV here?

LucioFranco (approved):

PR #519, opened 2024-09-22
Add `html_as_default_extension` option to `ServeDir`

its-the-shrimp:
## Motivation

It's common for file servers to add an `.html` file extension to the requested path if no file extension is provided, to allow for more concise & human-readable links, e.g. for a request of `/dashboard`, return `/dashboard.html`

## Solution

Add a `html_as_default_extension` option to the `ServeDir` service, which will be `true` by default, to be in tune with [`append_index_html_on_directories`](https://docs.rs/tower-http/latest/tower_http/services/struct.ServeDir.html#method.append_index_html_on_directories)


jlizen (changes requested):
@its-the-shrimp i'm fine to add this API, it is indeed a common convention, but it should default to false

Let me know if you no longer have time to work on this, I don't mind finishing it up.

its-the-shrimp:
Done that

jlizen (commented):
@its-the-shrimp - Impl looks good, thanks! Can we also have a test case for when the suffix is added? Glad to merge with that in place.

jlizen (approved):
Added some tests and a bugfix. Thanks for the contribution!

Issue #520, opened 2024-09-22
decompression: worse throughput when using `tower_http::decompression` than manual impl with `async-compression`


magurotuna:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

I've seen in some situations the throughput of decompression gets significantly worse when using `tower_http::decompression` compared to manually implementing a similar logic with `async-compression` crate.

### Platform

Apple silicon macOS

(Not 100% sure, but should happen in Linux as well)

### Description

In Deno, we switched the inner implementation of [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) (JavaScript API) from [`reqwest`](https://docs.rs/reqwest/latest/reqwest/) based to [`hyper-util`](https://docs.rs/hyper-util/latest/hyper_util/) based.

https://github.com/denoland/deno/pull/24593

In the hyper-util based implementation, it uses `tower_http::decompression` to decompress the fetched data if necessary. Note here that reqwest doesn't use tower_http.

After this change, we started to see the throughput to be degraded especially when the server serves compressed large data. Looks at the following graph, showing how long each Deno version takes to 2k requests where it fetches compressed data from the upstream server and then forwards it the end client.

![performance of proxying compressed data in different versions of Deno](https://github.com/user-attachments/assets/093de1ce-fa03-4fa3-a647-35610cac97b8)

v1.45.2 is _before_ we switched to hyper-based `fetch` implementation. Since v1.45.3 when we landed it, the throughput got 10x worse.

Then I identified that `tower_http::decompression` causes this issue, and figured out that if we implement a decompression logic by directly using the `async-compression` crate, the performance gets back to what it was. (see https://github.com/denoland/deno/pull/25800 for how manual implementation with `async-compression` affects the performance)

You can find how I performed the benchmark at https://github.com/magurotuna/deno_fetch_decompression_throughput

PR #521, opened 2024-09-22
(de)compression: reduce memory allocation to improve performance

magurotuna:
## Motivation

Currently, every time `WrapBody::poll_frame` is called, new instance of `BytesMut` is created with the default capacity, which is effectively 64 bytes. This ends up with a lot of memory allocation in certain situations, making the throughput significantly worse.

#520 describes that a manual implementation of decompression logic using `async-compression` gets the performance issue with `tower_http::decompression` to be resolved. To identify what is making this difference, I captured a flamegraph. Here's the result:

flamegraph

As annotated, `tower_http::decompression` has more memory allocation which is the reason why the performance is degraded.

## Solution

To optimize memory allocation, `WrapBody` now gets `BytesMut` as its field, with initial capacity of 4096 bytes. This buffer will be reused as much as possible across multiple `poll_frame` calls, and only when its capacity becomes 0, new allocation of another 4096 bytes is performed.

The capacity of 4096 bytes is taken from the default capacity of the internal buffer used in `tokio_util::io::ReaderStream`: https://docs.rs/tokio-util/0.7.12/src/tokio_util/io/reader_stream.rs.html#8

## Performance Improvement

With this optimization, the flamegraph is changed as follows, where memory allocation is significantly reduced and optimized.

optimized flamegraph

Also, the throughput gets 10x better when we measure it using Deno's `fetch` implementation that internally uses `hyper` and `tower_http::decompresson`.
This graph shows how long each Deno version takes to handle 2k requests. The rightmost one named `v2.0.0-rc.4-tower-http-patched` is the one with this patch applied. The leftmost one, v1.45.2, does not use `tower_http::decompression`.

performance

For more information on how this result was measured, please refer to https://github.com/magurotuna/deno_fetch_decompression_throughput

Fixes: #520

### DIFF
--- a/tower-http/src/compression_utils.rs
+++ b/tower-http/src/compression_utils.rs
@@ -141,8 +141,15 @@ pin_project! {
         // rust-analyer thinks this field is private if its `pub(crate)` but works fine when its
         // `pub`
         pub read: M::Output,
+        // A buffer to temporarily store the data read from the underlying body.
+        // Reused as much as possible to optimize allocations.
+        buf: BytesMut,
         read_all_data: bool,
     }
+}
+
+impl<M: DecorateAsyncRead> WrapBody<M> {
+    const INTERNAL_BUF_CAPACITY: usize = 4096;
 }
 
 impl<M: DecorateAsyncRead> WrapBody<M> {
@@ -167,6 +174,7 @@ impl<M: DecorateAsyncRead> WrapBody<M> {
 
         Self {
             read,
+            buf: BytesMut::with_capacity(Self::INTERNAL_BUF_CAPACITY),
             read_all_data: false,
         }
     }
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Phenomenal write-up, thanks!

Dav1dde:
I just wanna say thanks for this.

I work on [Relay](https://github.com/getsentry/relay), which get's quite a chunk of compressed traffic. I updated tower and axum the other day and noticed about a 20% decrease in tokio runtime busy threads, a 4x increase in time spent reading (+decompressing) request bodies, a 50% reduction in the average latency and about 30% reduced latency in the max and p95's.

Now looking through several changelogs and git commits trying to find what caused this, I am relatively sure it was this PR!

PR #524, opened 2024-10-02
Update brotli dev-dependency from 6 to 7

musicinmybrain:
## Motivation

Version 7.0.0 of the `brotli` crate was just relased. There were few changes between 6.0.0 and 7.0.0, and it looks like https://github.com/dropbox/rust-brotli/commit/41cadaabb6c650ae518b30509a5fbe43ca8cfab9 was the only breaking one. I see no reason not to keep this dev-dependency up to date.

## Solution

I adjusted the version of `brotli` from `6` to `7` and confirmed that `cargo test` still passes.


jplatte (approved):
Thanks!

PR #533, opened 2024-11-22
body timeout - remove unnecessary bounds on service error

markdingram:
## Motivation

Removing these bounds `S::Error: Into<Box<dyn std::error::Error>>` allows these layers to be used with e.g. tower-test which supplies `Box<dyn std::error::Error + Send + Sync>` aka `BoxError`

## Solution

Drop unnecessary bounds - the service error, whatever that may be, is passed straight through


jplatte (approved):
Thanks!

PR #537, opened 2024-12-24
Add conversions into `UnsyncBoxBody`

nanoqsh:
## Motivation

I want to return http bodies of different types, for example from multiple match branches. The easiest way to do this is to box the body like:
```rust
use {http_body_util::combinators::UnsyncBoxBody, /**/};

let serve = ServeFile::new(/**/);
let res = match req.uri().path() {
    "s1" => Ok(Response::new(UnsyncBoxBody::new(/**/))),
    "s2" => Ok(Response::new(UnsyncBoxBody::new(/**/))),
    _ => serve
        .oneshot(req)
        .await
        .map(|res| res.map(UnsyncBoxBody::new)),
};
```
But in the case of the body type from the `ServeFile`/`ServeDir` service, I have to box what is already boxed. This requires double allocation and, more critically, double indirection. I would like to avoid these overhead costs.

## Solution

My proposal to add a conversion impl into the [`UnsyncBoxBody`](https://docs.rs/http-body-util/latest/http_body_util/combinators/struct.UnsyncBoxBody.html) type. This eliminates the need for additional boxing:
```diff
- .map(|res| res.map(UnsyncBoxBody::new))
+ .map(|res| res.map(UnsyncBoxBody::from))
```

## Alternatives
- I could use the [`Either`](https://docs.rs/http-body-util/latest/http_body_util/enum.Either.html) type, but it's not convenient with more than two different types. I could do infinite levels of nesting, but I want a more straightforward solution.
- I could use the service's [`fallback`](https://docs.rs/tower-http/latest/tower_http/services/fs/struct.ServeDir.html#method.fallback) method, but I want to resolve routes at a higher level, e.g. preferring a route even if a file with same name exists.
- If exposing implementation details is undesirable, it could be converted directly to the `Pin<Box<dyn Body<Data = D, Error = E> + Send>>` type.

### DIFF
--- a/tower-http/src/body.rs
+++ b/tower-http/src/body.rs
@@ -103,6 +103,12 @@ where
     }
 }
 
+impl<D, E> From<UnsyncBoxBody<D, E>> for http_body_util::combinators::UnsyncBoxBody<D, E> {
+    fn from(body: UnsyncBoxBody<D, E>) -> Self {
+        body.inner
+    }
+}
+
 impl<D, E> UnsyncBoxBody<D, E> {
     #[allow(dead_code)]
     pub(crate) fn new(inner: http_body_util::combinators::UnsyncBoxBody<D, E>) -> Self {
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -493,6 +493,12 @@ opaque_body! {
     pub type ResponseBody = UnsyncBoxBody<Bytes, io::Error>;
 }
 
+impl From<ResponseBody> for http_body_util::combinators::UnsyncBoxBody<Bytes, io::Error> {
+    fn from(body: ResponseBody) -> Self {
+        body.inner.inner
+    }
+}
+
 /// The default fallback service used with [`ServeDir`].
 #[derive(Debug, Clone, Copy)]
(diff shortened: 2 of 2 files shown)


jlizen (changes requested):
@nanoqsh sorry fro the extended review delay...

The crate has a policy of not leaking `http_body_util` directly in its public interface. 

But, we could avoid the double boxing by letting callers build `tower_http::body::UnsyncBoxBody` and internally converting it:

```
  // tower-http/src/body.rs
  impl<D, E> UnsyncBoxBody<D, E>
  where
      D: Buf + 'static,
  {
      pub fn new<B>(body: B) -> Self
      where
          B: Body<Data = D, Error = E> + Send + 'static,
      {
          Self {
              inner: http_body_util::combinators::UnsyncBoxBody::new(body),
          }
      }
  }

  // tower-http/src/services/fs/serve_dir/mod.rs
  impl From<ResponseBody> for UnsyncBoxBody<Bytes, io::Error> {
      fn from(body: ResponseBody) -> Self {
          body.inner
      }
  }
```

and that lets you just do the following, pretty ergonomic:
```
Response::new(UnsyncBoxBody::new(body_a))
```

Does that work for you? Interested in implementing?

jlizen:
Superseded by #682

PR #541, opened 2025-01-20
Constantify serve_{dir,file} test file paths and content

nc7s:
## Motivation

Mainly a QoL change for downstream packagers, but might also prove useful for e.g. future refactors.

Currently we have to [patch all those file paths](https://salsa.debian.org/rust-team/debcargo-conf/-/blob/ee75c0c18a/src/tower-http/debian/patches/missing-testdata) to something exsistent, due to test-files/ not being included in crates.io releases, while the Debian Rust Team pulls Rust crate source from there.

## Solution

With these constants we could change only them.

Test files are also changed to the same content, so one constant fits all.


nc7s:
We frequently patch code in distros, so running tests there make sure our patches don't break things.

The diff - sure could be improved. Now that Debian's in freeze, it's not any urgent.

jlizen:
Closing in favor of #689

PR #552, opened 2025-03-24
feat(request-id): Add request-id constructor

tottoto:
## Motivation

Provides the same constructor of request-id as x-request-id.

## Solution

Adds request-id constructor.

### DIFF
--- a/tower-http/src/request_id.rs
+++ b/tower-http/src/request_id.rs
@@ -181,6 +181,7 @@ use tower_layer::Layer;
 use tower_service::Service;
 use uuid::Uuid;
 
+pub(crate) const REQUEST_ID: HeaderName = HeaderName::from_static("request-id");
 pub(crate) const X_REQUEST_ID: HeaderName = HeaderName::from_static("x-request-id");
 
 /// Trait for producing [`RequestId`]s.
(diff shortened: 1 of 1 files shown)


tower-http/src/request_id.rs
```diff
@@ -181,6 +181,7 @@ use tower_layer::Layer;
 use tower_service::Service;
 use uuid::Uuid;
 
+pub(crate) const REQUEST_ID: HeaderName = HeaderName::from_static("x-request-id");
```
jplatte, line 184: This still has the `x-` 😄 
But also, do you have a link to docs for the non-`x-`-prefixed header?

tottoto: Fixed, and rebased to resolve the conflict.

tottoto: https://www.rfc-editor.org/rfc/rfc6648 deprecates "X-" prefix header, and `request-id` is also commonly used same as `x-request-id`.

jplatte: It says "deprecates the convention for *newly defined* parameters" (emphasis mine) though. Do you have a source for the claim that it's commonly used?

jlizen (changes requested):
@tottoto I didn't see a lot of public usage. A smattering of larger clients (stripe, microsoft stuff, etc). But nothing reaching the point of convention.

Would be happy to accept a doc change to show the full custom header usage:
```
  const REQUEST_ID: HeaderName = HeaderName::from_static("request-id");

  let svc = ServiceBuilder::new()
      .set_request_id(REQUEST_ID, MyMakeRequestId::default())
      .propagate_request_id(REQUEST_ID)
      .service(handler);
```

It also might nice to offer a cleaner helper...
```
  let svc = ServiceBuilder::new()
      .emit_request_id("request-id", MyMakeRequestId::default())
      .service(handler);
```

jlizen:
Closing due to inactivity but happy to discuss further @tottoto , ref: https://github.com/tower-rs/tower-http/pull/552#pullrequestreview-4223982153

PR #555, opened 2025-04-02
feat(redirect): add `new` method for `Attempt`

linyihai:
## Motivation

Due to `Attempt` fields are all private, that dissallows producting `Attempt` outside.

My another PR needs a way to construct a `tower-http` Attempt, see https://github.com/seanmonstar/reqwest/pull/2617

## Solution

Add a `new` method and passing all the needed fileds to produce the `Attempt`.  This way has less expansibility when adding new fields to `Attempt`.

### DIFF
--- a/tower-http/src/follow_redirect/policy/mod.rs
+++ b/tower-http/src/follow_redirect/policy/mod.rs
@@ -217,6 +217,15 @@ impl<'a> Attempt<'a> {
     pub fn previous(&self) -> &'a Uri {
         self.previous
     }
+
+    /// Creates a new `Attempt`.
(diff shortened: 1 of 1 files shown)


linyihai:
> I don't think this method should be required. If the `loop` in reqwest (which mostly exists to handle redirects) is replaced by putting the internal `hyper.call()` with a `svc.call()`, such that `svc` is a the hyper client (and any other eventual layers) wrapped in a `tower_http::follow_redirect::FollowRedirect`, then the `Attempt` will be constructed automatically and passed into our `impl Policy`.

I'll look into `tower_http::follow_redirect::FollowRedirect`, thanks for you reminders.  

> Would you like me to sketch out futher what I mean?

Yes, It would be nice if there were more implementation details

PR #558, opened 2025-04-15
Update mod.rs

12bitentropy:
## Motivation

The component check only works on windows. If this was run on linux, the code will always push the component to the path_to_file variable, even if it contains c: in the path. 

## Solution

Use #[cfg(target_os = "windows")] to disable it

### DIFF
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -477,7 +477,7 @@ impl ServeVariant {
                                     return None;
                                 }
                             }
-                            
+
                             path_to_file.push(comp);
                         }
                         Component::CurDir => {}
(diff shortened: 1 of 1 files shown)


jplatte:
The particular example of `c:` indeed is not going to hit the early-exit branch on unix. That does not mean that this check should only happen on Windows though. I'm pretty sure it's possible to hit non-`Normal` path segments on non-Windows targets and I'd need to see strong proof that it's impossible for a particular target before this check for that target.

jplatte:
I'm not following your reasoning. Actually I think this change may be fine because this is only for checking individual segments reinterpreted as separate paths, but it's really not obvious to me, and what the documentation about prefix says seems not very relevant to the discussion. (the code in question rejects anything that is not `Normal`, not just `Prefix`)

jlizen:
@12bitentropy this check is basically free (no syscall, no allocation, very unlikely to see any overhead from it in profiling). We prefer defense in depth here:
- the rust stdlib doesn't guarantee its current behavior that makes it redundant in its own contract
- it's a maintenance hazard if we add further platform targets (wasi, fuschia, etc), since we need to remember to go and check for quirks in that platform's parsing
- parsing can be tricky to reason correctly about and there might be future issues that we haven't thought of that do impact *nix platforms

So, agreed with jplatte@'s conclusion here.

jplatte:
> fuschia

fuchsia ;)

PR #559, opened 2025-04-17
follow_redirect: expose previous and next request methods

lucab:
This augments follow-redirect `Attempt` in order to expose two relevant HTTP methods during redirects:
 - the method for the original/previous request
 - the method to be performed on the next request upon following a redirection

This allows consumers to write policies based on both previous and next requests methods.

### DIFF
--- a/tower-http/src/follow_redirect/mod.rs
+++ b/tower-http/src/follow_redirect/mod.rs
@@ -310,7 +310,7 @@ where
 
         let attempt = Attempt {
             status: res.status(),
-            next_method: &this.method,
+            method: this.method,
             location: &location,
             previous_method,
             previous: this.uri,
@@ -446,6 +446,31 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn stops_post_redirect_get() {
+        let policy = policy::redirect_fn(|attempt| -> Result<_, Infallible> {
+            if attempt.previous_method() == Method::POST && attempt.method() == Method::GET {
+                Ok(Action::Stop)
+            } else {
+                Ok(Action::Follow)
+            }
+        });
+        let svc = ServiceBuilder::new()
+            .layer(FollowRedirectLayer::with_policy(policy))
+            .service_fn(handle);
+        let req = Request::builder()
+            .method(Method::POST)
+            .uri("http://example.com/42")
+            .body(Body::empty())
+            .unwrap();
+        let res = svc.oneshot(req).await.unwrap();
+        assert_eq!(*res.body(), 42);
+        assert_eq!(
+            res.extensions().get::<RequestUri>().unwrap().0,
+            "http://example.com/42"
+        );
+    }
+
+    #[tokio::test]
     async fn limited() {
         let svc = ServiceBuilder::new()
             .layer(FollowRedirectLayer::with_policy(Limited::new(10)))
@@ -463,7 +488,7 @@ mod tests {
         );
     }
 
-    /// A server with an endpoint `GET /{n}` which redirects to `/{n-1}` unless `n` equals zero,
+    /// A server with an endpoint `/{n}` which redirects to `/{n-1}` unless `n` equals zero,
     /// returning `n` as the response body.
     async fn handle<B>(req: Request<B>) -> Result<Response<u64>, Infallible> {
         let n: u64 = req.uri().path()[1..].parse().unwrap();
--- a/tower-http/src/follow_redirect/mod.rs
+++ b/tower-http/src/follow_redirect/mod.rs
@@ -259,6 +259,7 @@ where
         let mut res = ready!(this.future.as_mut().poll(cx)?);
         res.extensions_mut().insert(RequestUri(this.uri.clone()));
 
+        let previous_method = &this.method.clone();
         let drop_payload_headers = |headers: &mut HeaderMap| {
             for header in &[
                 CONTENT_TYPE,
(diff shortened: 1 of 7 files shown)


lucab:
For context:
 * This is a re-spin of https://github.com/seanmonstar/reqwest/pull/2430
 * The original usecase is described in https://github.com/seanmonstar/reqwest/issues/1777#issuecomment-2363279906

tower-http/src/follow_redirect/policy/mod.rs
```diff
@@ -208,11 +210,21 @@ impl<'a> Attempt<'a> {
         self.status
     }
 
+    /// Returns the method for the next request, after applying redirection logic.
+    pub fn next_method(&self) -> &Method {
```
jlizen: should return `'a` lifetime

tower-http/src/follow_redirect/policy/mod.rs
```diff
@@ -208,11 +210,21 @@ impl<'a> Attempt<'a> {
...
+        self.next_method
+    }
+
     /// Returns the destination URI of the redirection.
     pub fn location(&self) -> &'a Uri {
         self.location
     }
 
+    /// Returns the method for the previous request, before redirection.
+    pub fn previous_method(&self) -> &Method {
```
jlizen: should return `'a` lifetime

tower-http/src/follow_redirect/policy/mod.rs
jlizen: this should probably be `method` to match `location`

tower-http/src/follow_redirect/policy/same_origin.rs
```diff
@@ -33,7 +33,7 @@ impl<B, E> Policy<B, E> for SameOrigin {
 #[cfg(test)]
```
jlizen, line 33: would be nice to throw in some simple assertions using these new accessors, if they fit into any existing tests

jlizen (changes requested):
Sorry for the slow review @lucab . Looks good! The lifetimes on the methods need fixing, and the rest are nits.

Might be nice to have a module example like, detecting cycles:
```
if attempt.previous_method() != attempt.method() { }
```

lucab:
@jlizen thanks for the review! I'll try to find some time in the next days to rebase and address your feedback, I'll ping you when this is ready for another pass.

Just one question on:
> Might be nice to have a module example like, detecting cycles

There is already a cyclic-redirection example at https://docs.rs/tower-http/latest/tower_http/follow_redirect/policy/trait.Policy.html#example, maybe it would be better to tweak that to use a `HashSet<(Uri, Method)>`?

jlizen:
> I'll try to find some time in the next days to rebase and address your feedback

Sounds good! No rush.

> cycle detection example

Yeah that sounds like a nice improvement, as long as we can capture cycles where it is flipping methods (which sounds like your HashSet idea would). Totally optional though. Just was thinking about an example that would realistically surface this new api's usage.

tower-http/src/follow_redirect/policy/same_origin.rs
lucab: I've added a `stops_post_redirect_get` test to cover this.

lucab:
@jlizen updated, PTAL.

tower-http/src/follow_redirect/policy/mod.rs
```diff
@@ -208,11 +210,21 @@ impl<'a> Attempt<'a> {
         self.status
     }
 
+    /// Returns the method for the next request, after applying redirection logic.
```
jlizen:
```suggestion
    /// Returns the destination method of the redirection.
```

tower-http/src/follow_redirect/mod.rs
```diff
@@ -259,6 +259,7 @@ where
         let mut res = ready!(this.future.as_mut().poll(cx)?);
         res.extensions_mut().insert(RequestUri(this.uri.clone()));
 
+        let previous_method = &this.method.clone();
```
jlizen:
```suggestion
        let previous_method = this.method.clone();
```

tower-http/src/follow_redirect/mod.rs
```diff
@@ -309,7 +310,9 @@ where
 
         let attempt = Attempt {
             status: res.status(),
+            method: this.method,
             location: &location,
+            previous_method,
```
jlizen:
```suggestion
            previous_method: &previous_method
```

tower-http/src/follow_redirect/mod.rs
```diff
@@ -442,6 +445,31 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn stops_post_redirect_get() {
```
jlizen: Optional: This is a good opportunity to flesh out our test coverage a bit, since we have new accessors. Up to you if you care to!

We could test:
- `307/308 POST` preservation
- `303 POST` -> GET
- `303 HEAD` preservation

Probably is worth parameterizing a helper at that point.

jlizen (changes requested):
Looks great! Couple more nits, then another one that's up to you.

tower-http/src/follow_redirect/mod.rs
lucab: That makes sense. I added a shared redirection handler and then dedicated tests to cover the following matrix:
 - 301
   * POST -> GET
   * GET -> GET
 - 302
   * POST -> GET
   * PUT -> PUT
   * HEAD -> HEAD
 - 303
   * POST -> GET
   * PUT -> GET
   * HEAD -> HEAD
 - 307/308
   * POST -> POST

jlizen (approved):
Looks good, thanks for the contribution!

PR #561, opened 2025-04-24
fix: use mem::swap on cloned inner service

tomharmon:
## Motivation

Without this change the inner service may `panic!` since the cloned service may not be the ready service as described in 
[the tower docs](https://docs.rs/tower/latest/tower/trait.Service.html#be-careful-when-cloning-inner-services)

## Solution

Use `std::mem::replace` to swap the ready and cloned services

### DIFF
--- a/tower-http/src/auth/async_require_authorization.rs
+++ b/tower-http/src/auth/async_require_authorization.rs
@@ -121,6 +121,7 @@ use http::{Request, Response};
 use pin_project_lite::pin_project;
 use std::{
     future::Future,
+    mem,
     pin::Pin,
     task::{ready, Context, Poll},
 };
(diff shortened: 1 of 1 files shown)


tower-http/src/auth/async_require_authorization.rs
```diff
@@ -204,6 +205,7 @@ where
     fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
         let inner = self.inner.clone();
         let authorize = self.auth.authorize(req);
+        let inner = mem::replace(&mut self.inner, inner);
```
jplatte, line 208: Can use `mem::swap`, which is shorter. Also, mind adding a comment about why the service and its clone need swapping?

jplatte (approved):

jplatte:
Thanks!

PR #566, opened 2025-05-09
Make type inference work for ServiceExt

jplatte:
Unfortunately, making type inference work for this requires removing the generic parameter from `ServiceExt`, which in turn requires removing the `tower::Service` supertrait. This made the `Sealed` trait useless, and while I was at it I also removed the `Sized` supertrait in favor of `Sized` bounds on all current trait methods (this allows backwards-compatibly adding methods that don't require Self to be Sized to be added).

This means that now you can call all of these methods on _anything_, the resulting type just isn't going to be very useful if the type you're calling the method on isn't a `tower::Service<http::Request<B>>` (extra bounds may apply depending on the exact method).

I also tested that this works in a real-world setting.
I think the only realistic alternative to this is not having `ServiceExt` at all.

### DIFF
--- a/tower-http/src/service_ext.rs
+++ b/tower-http/src/service_ext.rs
@@ -1,28 +1,23 @@
 #[allow(unused_imports)]
 use http::header::HeaderName;
 
-mod sealed {
-    #[allow(unreachable_pub, unused)]
-    pub trait Sealed<R> {}
-}
-
 /// Extension trait that adds methods to any [`Service`] for adding middleware from
 /// tower-http.
 ///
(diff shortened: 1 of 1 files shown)


seanmonstar:
Just curious, I didn't see an issue or someone point out compilation errors. Is there an example somewhere of what was broken?

jplatte:
Type inference can't figure out `R` when you chained these methods with the previous definition. I saw the same thing with tower's `ServiceExt` actually. Try to use one of its methods on a `ServeDir`.

LucioFranco (approved):

PR #574, opened 2025-05-28
Clarify TimeoutBody documentation

mherrerarendon:
Adds a couple of clarifying lines to the `TimeoutBody` documentation to explain that the timeout applies to the time interval between frame polls, as opposed to the entirety of the body.

## Motivation
I was confused about why the timeout was not being triggered even though the time it took to stream the response was longer than the timeout. It wasn't until I looked at what the code was doing that I realized why the timeout was not being triggered.

### DIFF
--- a/tower-http/src/timeout/body.rs
+++ b/tower-http/src/timeout/body.rs
@@ -12,10 +12,10 @@ use tokio::time::{sleep, Sleep};
 pin_project! {
     /// Middleware that applies a timeout to request and response bodies.
     ///
-    /// Wrapper around a [`http_body::Body`] to time out if data is not ready within the specified duration. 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for improving the docs! I'm always for making things clearer. One question: does the following existing line say something similar?

> If the body does not produce a requested data frame within the timeout period, it will return an error.

Should the lines be collapsed somehow?

mherrerarendon:
> Thanks for improving the docs! I'm always for making things clearer. One question: does the following existing line say something similar?
> 
> > If the body does not produce a requested data frame within the timeout period, it will return an error.
> 
> Should the lines be collapsed somehow?

@seanmonstar You are right, I've updated it. Let me know what you think!

PR #575, opened 2025-06-01
feat(redirect): Removed `headers` after empty payload set.

linyihai:
## Motivation

This patch removed the `content-type`, `content-length`, `content-encoding`, `transfer-encoding` headers in the follow_redirect::poll after the response `body` is set to empty. This logic works well in `reqwest`.

Currently, the follow_redirect::poll doesn't remove these `header` after redirecting, this behavior is a violation of the [HTTP/1.1 spec](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2)

Also see https://github.com/seanmonstar/reqwest/issues/2705

## Solution

Remove the headers before making redirection


tower-http/src/follow_redirect/mod.rs
```diff
@@ -257,13 +259,24 @@ where
         let mut res = ready!(this.future.as_mut().poll(cx)?);
         res.extensions_mut().insert(RequestUri(this.uri.clone()));
 
+        let drop_payload_headers = |headers: &mut HeaderMap<HeaderValue>| {
```
jplatte, line 262:
```suggestion
        let drop_payload_headers = |headers: &mut HeaderMap| {
```

linyihai: Thanks 👍

PR #578, opened 2025-06-03
Fix an indexing panic in compression

sulami:
This guard was backwards, not only breaking the functionality it was intended to provide, but also causing panics under certain conditions.

In this case, the haystack is the pre-existing Vary header on the response, and the needle is "Accept-Econding".

If we are compressing a response, and there is already a Vary header on the response, it is <= 15 bytes and does not match Accept-Encoding, this would cause a range index error as the while statement would not provide an effective guard.

I'm happy to contribute a test or two as well, if so desired.


seanmonstar (approved):
🙈

Issue #300, opened 2022-10-04
Should `TimeoutLayer` middleware allow returning other status codes?


mlodato517:
### Motivation

I'd like to be able to return a 504 if the timeout given to the `TimeoutLayer`
is exceeded. This is because a 408, which is currently returned, [is meant to
communicate][0]:
> that the server did not receive a complete request message within the time
that it was prepared to wait.

I don't know if in general we can tell if the issue is the client sending the
request or the server processing the request but in the case of the latter
(especially because this service is proxying to another) I'd like to return
a 504 Gateway Timeout instead (which is still not a guarantee - the issue
could have been the upstream or any post-processing of the upstream's
response but that's rare in this case). Servers like `axum` and `actix-web`
provide other mechanisms for timeouts when receiving things like request
headers[^1][^2] so maybe something about receiving the request should be handled
at the server level or by something like [`BodyTimeout`][2]?

[0]: https://httpwg.org/specs/rfc9110.html#status.408
[2]: https://github.com/tower-rs/tower-http/issues/295
[^1]: https://docs.rs/axum-server/latest/axum_server/struct.HttpConfig.html#method.http1_header_read_timeout
[^2]: https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.client_request_timeout

### Proposal

Changing the response status from 408 all the time to 504 all the time
would equally be incorrect sometimes and would be a breaking change.
Luckily we already have a `new()` constructor for `TimeoutLayer` so I
imagine that could stay the same and default to 408 and then we could
have a `new_with_response` constructor that accepts an `http::Response<B>`
or similar that can be returned instead.

### Alternatives

The most obvious alternative I can see follows from [this comment][1]
which is to use `tower` instead of `tower-http` and explicitly catch the
error and perform custom logic on it. This isn't too much code but I
wonder if the alternate constructor is valuable because I'm not sure
how frequent 408 will be the desired response code (though HTTP
semantics are fuzzy enough that maybe it's not worth it!).

[1]: https://github.com/tower-rs/tower-http/pull/270#issue-1239038234

davidpdrsn:
> Servers like axum and actix-web
provide other mechanisms for timeouts when receiving things like request
headers

From a tower middleware it's not possible to distinguish between reading the request head timing out or the server handling the request. That requires a specific integration into hyper.

I think if you want a different status you can write your own middleware for it. It's not complicated. If you're using axum it's even easier with `axum::middleware::from_fn`.

[...]

PR #579, opened 2025-06-04
Change `Timeout` middleware to return HTTP `503 Service Unavailable` status code

teohhanhui:
## Motivation

See #300

## Solution

Change `Timeout` middleware to return HTTP `503 Service Unavailable` status code.

Returning HTTP `408 Request Timeout` status code wrongly indicates a client error.

### DIFF
--- a/tower-http/src/timeout/mod.rs
+++ b/tower-http/src/timeout/mod.rs
@@ -1,7 +1,7 @@
 //! Middleware that applies a timeout to requests.
 //!
-//! If the request does not complete within the specified timeout it will be aborted and a `408
-//! Request Timeout` response will be sent.
+//! If the request does not complete within the specified timeout it will be aborted and a `503
+//! Service Unavailable` response will be sent.
 //!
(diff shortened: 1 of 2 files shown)


seanmonstar:
While in general this does seem like a better error code, it's not always true. For instance, if the inner service times out because it's currently waiting on `req.body.collect()`, then that correctly would be a 408.

Looking at the linked issue, we could probably solve this by adding some configuration to the service/layer, so you can customize the response code returned.

teohhanhui:
> For instance, if the inner service times out because it's currently waiting on `req.body.collect()`

Doesn't that tie into [`TimeoutBody`](https://docs.rs/tower-http/latest/tower_http/timeout/struct.TimeoutBody.html)?

> adding some configuration to the service/layer

We'd have to take in a `Future`, right?

jplatte:
Feel free to open a new PR, yeah. But what the heck are you doing where timeout should return 204 (which is certainly _not_ an *error* code)?

PR #580, opened 2025-06-05
Call `on_eos` for successful streams

### LINKED ISSUE #550: `tower_http::trace::OnEos` only called if EOS is classified
It seems a bit odd to me that when using [`ServerErrorsAsFailures`](<https://docs.rs/tower-http/0.6.2/tower_http/classify/struct.ServerErrorsAsFailures.html>) (the default classifier used by `TraceLayer::new_for_http()`), the `OnEos` callback is never called. After looking at the code, I found that the response body only carries the callback if the response could not yet be classified: <https://docs.rs/tower-http/0.6.2/src/tower_http/trace/future.rs.html#83>

I thought if `OnBodyChunk` is called each time the body is (successfully) polled, I would also get a callback for when it's done, but the way the traits are organized, I don't think I can get what I want without changing `tower_http`'s internals. Is there a reason for why these need to be tied together, or was this just an oversight? Or perhaps I'm missing something?

mladedav:
## Motivation

Fixes #550.

`on_eos` is called only if classification declares that it cannot classify the response until eos.

The documentation does not mention such limitation.

## Solution

Pass the callback to the response body for it to be called even if it's not needed for response classification.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# Unreleased
+
+## Fixed
+
+- `on_eos` is now called even for successful responses. ([#580])
+
+[#580]: https://github.com/tower-rs/tower-http/pull/580
+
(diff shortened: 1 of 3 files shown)


mladedav:
@jplatte would you mind taking a look? It would help me a lot if this made it into the next release (whenever that might be)

jplatte (approved):
Seems to match what the documentation says this does.

PR #581, opened 2025-06-06
feat(follow_redirect): Preserve extensions during redirection

0x676e67:
## Motivation

Extensions are passed during request redirection. Extensions that currently rely on `http 1.x` are now cloneable.

https://github.com/hyperium/http/blob/613d3d465f222666327d61f5971133cad155f190/src/extensions.rs#L35

## Solution

Solve some scenarios where redirection requires passing `Extensions`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -81,6 +81,7 @@ jobs:
     - run: cargo update -p tokio-util --precise 0.7.11
     - run: cargo update -p flate2 --precise 1.0.35
     - run: cargo update -p once_cell --precise 1.20.3
+    - run: cargo update -p tracing-core --precise 0.1.33
     - run: cargo check -p tower-http --all-features
 
   style:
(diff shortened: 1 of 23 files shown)


0x676e67:
Is this approach justifiable?

0x676e67:
Four hours ago, tracing-core updated its version and upgraded msrv. CI seems to have failed because of this.

jplatte:
Well, FWIW I didn't know this repo had a `CONTRIBUTING.md` 😅
@seanmonstar do you care about the stuff in there? In particular force-pushes to PRs?

jlizen (changes requested):
@0x676e67 Sorry for the slow review...

I think that this was probably an oversight to omit, but there are security implications to turning it on. Extensions can leak sensitive information that previously was dropped. So certainly a breaking change.

We should also keep parity with the headers APIs and have `FilterCredentials` support for opt out via policy. It's a trickier problem than headers since mostly userspace types that we can't have a default excludelist for. I guess we could do some clever things if we have any types gated behind peer feature flags that we want to filter out.

Otherwise it seems like its mostly a user-provided filter. Probably something doable with macros to make it not-horrid to pass in a static list of type names/ids.

A simpler and non-breaking option would just be a opt-in API. (Long term I think it should be opt-out, but it can be a follow up)

jlizen:
Closing in favor of #706

PR #586, opened 2025-06-13
ServeDir to call fallback for invalid filenames (fixes #573)

### LINKED ISSUE #573: %00 sequence in path returns ISE (even though the problem is in request)

### Version

tower-http v0.6.4

### Platform

`Linux silmeth-laptop 6.14.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Sun Apr  6 15:05:05 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

### Description

When `ServeDir` service gets requests with the sequence `%00` (percent-encoded null byte) in the path (eg. `/app/%00/foo.xml`), it returns status code `500` Internal Server Error – suggesting there is a problem with the server’s file system, but fault is in the request (invalid file path character encoded in the request’s path).

It also logs this prints this problem to the log:
```
tower_http::services::fs::serve_dir: Failed to read file error=file name contained an unexpected NUL byte
```

I would expect it to return some 4xx bad request status instead.

aryaveersr:
## Motivation

( See #573 )

ServeDir currently returns ISE for any request involving invalid filenames (eg: `/path|name/`). This fixes it so that it instead forwards the request to the fallback (as it should).

## Solution

This uses a rather ugly solution of manually checking Raw OS error codes for windows. If we ever upgrade MSRV to 1.87 or beyond, we should instead use `io::ErrorKind::InvalidFilename` (stabilized in 1.87). I've also put a comment with FIXME regarding this.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -7,8 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # Unreleased
 
+## Fixed
+
+- `on_eos` is now called even for successful responses. ([#580])
 - `ServeDir`: call fallback when filename is invalid ([#586])
 
+[#580]: https://github.com/tower-rs/tower-http/pull/580
 [#586]: https://github.com/tower-rs/tower-http/pull/586
 
 # 0.6.6
--- a/tower-http/src/trace/future.rs
+++ b/tower-http/src/trace/future.rs
@@ -80,7 +80,7 @@ where
                         let res = res.map(|body| ResponseBody {
                             inner: body,
                             classify_eos: None,
-                            on_eos: None,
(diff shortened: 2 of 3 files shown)


tower-http/src/services/fs/serve_dir/future.rs
```diff
@@ -105,7 +105,7 @@ where
                         break Poll::Ready(Ok(res));
                     }
 
-                    Ok(OpenFileOutput::FileNotFound) => {
+                    Ok(OpenFileOutput::FileNotFound) | Ok(OpenFileOutput::InvalidFilename) => {
```
jplatte, line 108:
```suggestion
                    Ok(OpenFileOutput::FileNotFound | OpenFileOutput::InvalidFilename) => {
```

jplatte (approved):
Thanks!

PR #587, opened 2025-07-01
Fix serve_dir method not allowed handling when no fallback is configured

soerenmeier:
Fixes a bug in `ServeDir` where non-GET/HEAD requests were incorrectly processed when `call_fallback_on_method_not_allowed` is enabled but no fallback service is provided.

**Problem**
When configured with `call_fallback_on_method_not_allowed(true)` but no fallback service, non-GET/HEAD requests would proceed through normal file serving logic instead of returning `405 Method Not Allowed`.


jlizen (approved):
Appreciate it!

PR #591, opened 2025-07-10
Deprecate bearer/basic auth middlewares

### LINKED ISSUE #585: Deprecate the simple Bearer and Basic auth middlewares
Hi folks!

I noticed both the [bearer auth](https://github.com/tower-rs/tower-http/blob/20c7c9a5be035815da3663ee0475ca91e219f0e4/tower-http/src/auth/require_authorization.rs#L178) and [basic auth](https://github.com/tower-rs/tower-http/blob/20c7c9a5be035815da3663ee0475ca91e219f0e4/tower-http/src/auth/require_authorization.rs#L235) middlewares are using normal equality comparisons (`==`) on the header values. From what I understand, this can potentially lead to timing attacks, since string/byte comparisons are not constant-time and exit once the first non-equal byte/character is found.

Should this be updated to use a constant-time comparison (e.g. https://docs.rs/constant_time_eq/latest/constant_time_eq/)?

Thanks!

gjabell:
Closes #585

### DIFF
--- a/tower-http/src/auth/require_authorization.rs
+++ b/tower-http/src/auth/require_authorization.rs
@@ -1,3 +1,4 @@
+#![deprecated(since = "0.6.7", note = "too basic to be useful in real applications")]
 //! Authorize requests using [`ValidateRequest`].
 //!
 //! [`Authorization`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization


tower-http/src/auth/require_authorization.rs
```diff
@@ -1,3 +1,7 @@
+#![deprecated(
+    since = "0.6.6",
```
jplatte, line 2:
```suggestion
    since = "0.6.7",
```

tower-http/src/auth/require_authorization.rs
```diff
@@ -1,3 +1,7 @@
...
+    note = "uses non-constant time string comparison for tokens which could allow timing attacks"
```
jplatte, line 3: Tbh, not a fan of this deprecation note. If this was the only problem with this middleware, I don't think that would be a reason to deprecate it. Rather, it's

- as Sean noted, this middleware just isn't applicable to real-world scenarios due to how limited it is
- we don't have the resources / motivation to turn it into something useful
- writing your own without this is very easy, especially with something like axum's `middleware::from_fn`

jplatte (approved):

jplatte:
Thanks!

PR #595, opened 2025-08-09
chore(examples): resolve clap deprecated api

tottoto:
Resolves the clap deprecated api.

### DIFF
--- a/examples/axum-key-value-store/src/main.rs
+++ b/examples/axum-key-value-store/src/main.rs
@@ -25,7 +25,7 @@ use tower_http::{
 #[derive(Debug, Parser)]
 struct Config {
     /// The port to listen on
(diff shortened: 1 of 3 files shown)


jplatte:
By the way, if you could start your commit message (after the conventional commit prefix) with an uppercase character, that would be appreciated.

Also, this is just my personal opinion, but I think `examples: ` is much nicer than `chore(examples): `. We don't use any tooling for conventional commits anyways.

jplatte:
Well, I guess I am kind of the primary maintainer, so I think I can say yes ^^

However, would also like to hear @seanmonstar's opinion (I didn't notice he used a different style, to me the uppercase thing always seemed like an accepted standard amongst the vast majority of people care about their commit message formatting).

PR #596, opened 2025-08-10
doc: Update link to tower-http main repository

tottoto:
## Motivation

The branch name seems to be changed.

## Solution

Replaces them to the updated name.


jplatte (approved):
Thanks!

PR #601, opened 2025-08-21
examples: Drop protoc dependency

tottoto:
## Motivation

Removes protoc dependency to simplify CI.

## Solution

Uses protox to build proto.


jplatte (approved):
Nice :)

PR #607, opened 2025-09-15
fix gate-ing of atomic64 in tests

alexanderkjall:
## Motivation

Fixes tests on some architectures.

based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/tower-http/debian/patches/gate-tests-atomic-u64

## Solution

use `#[cfg(target_has_atomic = "64")]`

### DIFF
--- a/tower-http/src/request_id.rs
+++ b/tower-http/src/request_id.rs
@@ -483,12 +483,11 @@ mod tests {
     use crate::test_helpers::Body;
     use crate::ServiceBuilderExt as _;
     use http::Response;
+    #[cfg(target_has_atomic = "64")]
+    use std::sync::atomic::AtomicU64;
     use std::{
         convert::Infallible,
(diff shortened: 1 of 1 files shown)


jplatte (commented):
How about switching to `AtomicU16`? Would that allow the tests to run on the architectures in question?

jlizen:
@alexanderkjall I think `AtomicUsize`everywhere would be our best bet. Since, we should also update our doc tests (and we don't want to show an `AtomicU16` in them).

That will get rid of the cfg flags, which will be nicer to maintain.

Are you interested in cutting another commit?

alexanderkjall:
Sure, replaced AtomicU64 with AtomicUsize, also switched it out in a doctest.

jlizen (approved):
Appreciate it!

PR #608, opened 2025-10-01
Remove deprecated annotations and Refactor From implementations

sinder38:
## Motivation

I saw multiple instances of:

```rust
#[allow(deprecated)] // Can be changed when MSRV >= 1.53
```

Since the current MSRV is 1.66, these can be simplified.
https://github.com/tower-rs/tower-http/blob/35740decc663f4921b85b234ae33580f40fcbb31/README.md?plain=1#L40 

This is not strictly a refactor, I replaced some specific array implementations with a blanket implementation.

## Solution

Replaced array-based `From` implementations with a generic, iterable-based approach

```rust
// Before:
impl<const N: usize> From<[HeaderName; N]> for AllowHeaders {
    fn from(arr: [HeaderName; N]) -> Self {
        #[allow(deprecated)] // Can be changed when MSRV >= 1.53
        Self::list(array::IntoIter::new(arr))
    }
}

// After:
impl<I> From<I> for AllowHeaders
where
    I: IntoIterator<Item = HeaderName>,
{
    fn from(iter: I) -> Self {
        Self::list(iter)
    }
}
```

For `Inner` structs supporting conversions from a single item, I kept the singular conversion method and removed deprecated array handling:

```rust
impl From<Method> for AllowMethods {
    fn from(method: Method) -> Self {
        Self::exact(method) // Single-item conversion
    }
}

// Removed deprecated array-based impl:
impl<const N: usize> From<[Method; N]> for AllowMethods {
    fn from(arr: [Method; N]) -> Self {
        #[allow(deprecated)] // Can be changed when MSRV >= 1.53
        Self::list(array::IntoIter::new(arr))
    }
}
```

This simplifies the code, reduces duplication, and removes no longer necessary `#[allow(deprecated)]` annotations.
:cat: :butter:

### DIFF
--- a/tower-http/src/cors/vary.rs
+++ b/tower-http/src/cors/vary.rs
@@ -44,11 +44,14 @@ impl Default for Vary {
     }
 }
 
-impl<I> From<I> for Vary
-where
-    I: IntoIterator<Item = HeaderName>,
-{
-    fn from(iter: I) -> Self {
-        Self::list(iter)
+impl<const N: usize> From<[HeaderName; N]> for Vary {
+    fn from(arr: [HeaderName; N]) -> Self {
+        Self::list(arr)
+    }
+}
+
+impl From<Vec<HeaderName>> for Vary {
+    fn from(vec: Vec<HeaderName>) -> Self {
+        Self::list(vec)
     }
 }
--- a/tower-http/src/cors/allow_headers.rs
+++ b/tower-http/src/cors/allow_headers.rs
@@ -86,12 +86,15 @@ impl From<Any> for AllowHeaders {
     }
 }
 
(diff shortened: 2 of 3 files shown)


tower-http/src/cors/vary.rs
```diff
@@ -46,15 +44,11 @@ impl Default for Vary {
...
-}
-
-impl From<Vec<HeaderName>> for Vary {
-    fn from(vec: Vec<HeaderName>) -> Self {
-        Self::list(vec)
+impl<I> From<I> for Vary
+where
+    I: IntoIterator<Item = HeaderName>,
+{
+    fn from(iter: I) -> Self {
```
jplatte: Again.

PR #613, opened 2025-10-21
ci: Switch cargo-public-api-crates to cargo-check-external-types

tottoto:
## Motivation

`cargo-public-api-crates` seems to have version conflict at the dependencies.

```
error: failed to compile `cargo-public-api-crates v0.2.0 (https://github.com/davidpdrsn/cargo-public-api-crates#13bf827a)`, intermediate artifacts can be found at `/tmp/cargo-installDFx8K1`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

Caused by:
  rustc 1.80.0-nightly is not supported by the following package:
    indexmap@2.12.0 requires rustc 1.82
  Either upgrade rustc or select compatible dependency versions with
  `cargo update <name>@<current-ver> --precise <compatible-ver>`
  where `<compatible-ver>` is the latest version supporting rustc 1.80.0-nightly
```

https://github.com/tower-rs/tower-http/actions/runs/18697143413/job/53317419830

## Solution

Switches `cargo-public-api-crates` to `cargo-check-external-types`.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -95,20 +95,17 @@ jobs:
         manifest-path: tower-http/Cargo.toml
         command: check ${{ matrix.checks }}
 
-  cargo-public-api-crates:
+  external-types:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v5
-    # Pinned version due to failing `cargo-public-api-crates`.
     - uses: dtolnay/rust-toolchain@master
       with:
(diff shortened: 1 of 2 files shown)


tottoto:
I noticed that I posted https://github.com/tower-rs/tower-http/pull/613#issuecomment-3605764328 with a copy and paste mistake: the place of `cargo-public-api-crates` in the post was intended to be `cargo-check-external-types`. I fixed it.

PR #614, opened 2025-10-23
Don't automatically enable tracing for fs feature

ginnyTheCat:
## Motivation

Currently the `"fs"` features is exceptional in the way that it enables tracing even if the user didn't opt into this using the `"tracing"` feature.

## Solution

Make tracing opt-in for the `"fs"` feature like it is for the other features as well.

### DIFF
--- a/tower-http/src/services/fs/serve_dir/open_file.rs
+++ b/tower-http/src/services/fs/serve_dir/open_file.rs
@@ -350,7 +350,7 @@ fn append_slash_on_path(uri: Uri) -> Result<Uri, OpenFileOutput> {
 
     uri_builder.build().map_err(|_err| {
         #[cfg(feature = "tracing")]
-        tracing::error!(?_err, "redirect uri failed to build");
+        tracing::error!(err = ?_err, "redirect uri failed to build");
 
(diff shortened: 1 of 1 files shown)


jplatte:
I'm a bit skeptical. I think people should get these error logs if they only enable the fs feature (I suppose we could make `tracing` a default feature, not sure what else this would affect).

Also: `tracing` is a pretty common dependency in the async Rust ecosystem, can you even get around it when depending on `tower-http`?

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -348,8 +348,10 @@ fn append_slash_on_path(uri: Uri) -> Result<Uri, OpenFileOutput> {
         uri_builder.path_and_query("/")
     };
 
-    uri_builder.build().map_err(|err| {
-        tracing::error!(?err, "redirect uri failed to build");
+    uri_builder.build().map_err(|_err| {
+        #[cfg(feature = "tracing")]
+        tracing::error!(?_err, "redirect uri failed to build");
```
jplatte:
```suggestion
        tracing::error!(err = ?_err, "redirect uri failed to build");
```

Issue #617, opened 2025-11-10
Handle empty body in decompression layer


ducaale:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Ignore decompression `UnexpectedEOF` errors for responses with empty body

### Motivation
Servers sometimes return `content-encoding`/`content-length` with empty body for `HEAD` requests. I was looking into refactoring reqwest's decompression to use `tower-http` and would like to port this functionality to `tower-http::decompression`.

### Proposal
Add boolean flag to `StreamErrorIntoIoError` to track whether any data has been read. Swallow any decompression errors in `WrapBody` if flag is not set.  

### Alternatives
I have thought about having two layers before and after decompression layer. One tracks if any data has been read while the other conditionally suppresses error (inspired by https://github.com/ducaale/xh/pull/411).

But before pursuing this idea any further, I wanted to see if this can be handled by `tower-http` instead.

PR #618, opened 2025-11-15
Suppress EOF errors caused by decompressing empty body

ducaale:
Resolves https://github.com/tower-rs/tower-http/issues/617


seanmonstar:
I think this looks right. I fixed that CI task, if you wanna rebase we should be good to go :)

PR #548, opened 2025-03-08
feat(decompression): support HTTP responses containing multiple ZSTD frames

0x676e67:
## Motivation
Prepare for reqwest refactoring
Tracing: https://github.com/seanmonstar/reqwest/issues/2575

## Solution

Ported from reqwest, solves the problem: https://github.com/seanmonstar/reqwest/issues/2574

PR #621, opened 2025-11-26
Disable `multiple_members` option for gzip decoder

ducaale:
## Motivation

When decompressing a chunked response with extra bytes like in https://github.com/seanmonstar/reqwest/blob/a2aa5a34e48724be0c1089b0f5afe49b82ece30e/tests/gzip.rs#L306, the client hangs and doesn't return error until the server drops connection.

I think this happens because those extra bytes are treated as the start of another gzip frame.

## Solution

We can disable `multiple_members` and assert that any frames after decompressing body are trailers.

### DIFF
--- a/tower-http/src/decompression/body.rs
+++ b/tower-http/src/decompression/body.rs
@@ -395,7 +395,9 @@ where
     type Output = ZstdDecoder<Self::Input>;
 
     fn apply(input: Self::Input, _quality: CompressionLevel) -> Self::Output {
-        ZstdDecoder::new(input)
+        let mut decoder = ZstdDecoder::new(input);
+        decoder.multiple_members(true);
+        decoder
     }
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
@0x676e67 I noticed you filed #548, and while I approved that, I no longer remember why. Seems like we're undoing that here, and perhaps never should have enabled it? I see it is enabled in reqwest for zstd, but all that I remember from gzip, that perhaps is wrong there too?

0x676e67:
> @0x676e67 I noticed you filed #548, and while I approved that, I no longer remember why. Seems like we're undoing that here, and perhaps never should have enabled it? I see it is enabled in reqwest for zstd, but all that I remember from gzip, that perhaps is wrong there too?

@seanmonstar Yes, I remember the context now — it started from these two issues/PRs:
https://github.com/seanmonstar/reqwest/issues/2574
https://github.com/seanmonstar/reqwest/pull/2583

PR #622, opened 2025-11-30
ci: Pin tracing in MSRV job

ducaale:
CI is failing in main since latest versions of `tracing` and `tracing-subscriber` now require rust 1.65


jplatte:
Thanks!

PR #628, opened 2025-12-24
chore: remove implicit tokio feature

### LINKED ISSUE #511: Remove implicit tokio feature
We have an implicit `tokio` Cargo feature because we don't use `dep:` syntax for it.

WaterWhisperer:
Fixes #511

## Motivation

> We have an implicit `tokio` Cargo feature because we don't use `dep:` syntax for it.

## Solution

makes the `tokio` dependency explicit


jplatte:
That CI error is really strange. Could you try moving the `rust-toolchain` action step in the MSRV job after the `cargo update` steps? Maybe this is an old bug in cargo that since got fixed.

WaterWhisperer:
> That CI error is really strange. Could you try moving the `rust-toolchain` action step in the MSRV job after the `cargo update` steps? Maybe this is an old bug in cargo that since got fixed.

Sure, thanks for the review

WaterWhisperer:
> Okay, so that didn't work. I have another idea: Can you (after rebasing, this PR seems to have a conflict now) move the toolchain action back, and also add `--all-features` to the problematic `cargo update` invocation? I think it's maybe saying that tokio is not a dependency that can be updated because it needs to resolve what dependencies actually exist and uses the default features for that unless told otherwise.

Seems that `cargo update` command does not support the `--all-features` flag.

WaterWhisperer:
This error is really strange :(

jlizen:
Updated to use @tottoto's approach, thanks:

```
  -metrics = ["dep:http-body", "tokio/time"]
  +metrics = ["dep:http-body", "dep:tokio", "tokio?/time"]
```

jlizen (approved):

PR #636, opened 2026-01-09
feat(on-early-drop): Add middleware for client early drop detection

### LINKED ISSUE #396: `tower_http::trace::OnEarlyDrop`
### Motivation

My service has recently had problems in which (for still-undiagnosed reasons), it sometimes takes long enough to respond that the client has hit its HTTP timeout and closed the connection before response headers were sent. We want to monitor for this condition.

`tower_http::trace` is a great framework for observing related problems (HTTP server/client errors, excessive latency, etc.) but appears to have a complete blind spot here. The span will end without any further log entries (and in my service's case, we've chosen not to log on *receiving* the request, so these timed-out requests are invisible in logs).

### Proposal

From experiment, I've found `hyper` drops the `call` future on client disconnect, so the obvious way to do this is to via a `Drop` impl on that future. If response headers haven't been sent at `drop` time, call a new hook similar to `tower_http::trace::OnFailure`.

The default impl could log this happened and the latency; maybe it needs a few parameters:

1. a duration threshold,
2. the log level under that threshold, and
3.  the log level at/above that threshold.

E.g., I don't consider it "my service's problem" if the client disconnects after waiting only 10ms but definitely do if it waited 10+s. The threshold is at least service-specific; in some cases decided based in a more complex fashion by a custom impl.

Name-wise, maybe `OnEarlyDrop` is more appropriate than say `OnClientDisconnect`, because if say you have a server-side `Timeout` layer in front of this one, the trace future's `drop` will be called on timeout. `OnEarlyDrop` is not as descriptive of the cause I'm interested in but also shouldn't ever be misleading.

A related idea: this isn't currently a focus of mine, but some folks may also want to know if the client closed the connection after a `Response` was returned but before the full body was generated. I think similarly `OnEos` just doesn't get called and there's no alternate plugin which does.

fbergero:
# Add on-early-drop middleware for client early drop detection

Introduces a new middleware that detects when a client disconnects before receiving a full response. This allows for monitoring and logging of early connection terminations.

Fixes tower-rs/tower-http#396

## Motivation

When clients disconnect before receiving a complete response (e.g., due to browser navigation, timeouts, or network issues), the entire future chain is typically dropped with no indication of what happened. This creates "disappearing requests" that never show up in logs or metrics, making it difficult to:

1. Identify patterns of client disconnects
2. Properly monitor the full request lifecycle
3. Handle or recover from incomplete operations

## Solution

This PR introduces two main components:

### 1. OnEarlyDropGuard

A standalone guard that executes a callback function when dropped, unless explicitly marked as completed:

- **Simple API**: Create the guard with a closure that executes on early drop
- **Callback Control**: Mark operations as successfully completed to prevent the callback
- **FnOnce Support**: Designed to work with closures that take ownership of their environment

### 2. OnEarlyDropLayer/Service

A Tower middleware that applies early drop detection to HTTP services:

- **Request Context**: Closures can capture specific request information

This implementation enables various use cases:
- Logging dropped requests with context
- Capturing metrics about client disconnections
- Maintaining trace context across early terminations
- Performing cleanup operations for abandoned requests


jlizen:
cc @seanmonstar  if you have any time, no problem if not. I've reviewed it back when it was on the fork, so fairly polished.

jlizen:
Hey @fbergero , like we chatted about offline, I pushed some changes on top of yours (and also dealt with the rebase).

Generally I added some more integration with the rest of the `tower-http` stack:
- there is an adapter to pass it into the `trace` layer's `OnFailure` handler
- you can fire a separate callback if the body drops before streaming finishes (that has access to response parts)
- axum example now uses this (via the tracing integration)

fbergero:
@jlizen Loved the `EarlyDropsAsFailures` for a nice-easy setup. Thanks for the improvements.

IDK if we need to update the PR desc to include the new changes (it's on the rustdocs anyhow so up to you)

tower-http/src/on_early_drop/mod.rs
```diff
@@ -0,0 +1,106 @@
+//! Middleware that detects when a response future or response body is
+//! Middleware that detects when a response future or response body is
```
rcoh: duplicate line

rcoh (approved):
there is one duplicated doc line otherwise LGTM

jlizen (approved):

PR #650, opened 2026-02-17
examples: Use axum::body::to_bytes

tottoto:
`axum::body::to_bytes` could be used to collect the response body.


jplatte (commented):
Why, to get rid of the extra dependency? I think I'd prefer the current code for clarity (limiting to `usize::MAX` is a bit weird).

jlizen (approved):
Thanks, agreed that it makes sense to use `axum` utilities here

PR #658, opened 2026-04-11
In ServeDir, change the behaviour for handling symlinks

alexanderkjall:
that point outside the base_dir to return errors, and introduce a new setting that will allow it (on linux)

## Motivation

Let me preface this with that I consider this a potential security problem, but I emailed security@tokio.rs and they didn't agree and encouraged me to open a public issue.

Now on to the issue:

ServeDir in tower-http doesn't have any configuration to control symlink behaviour. ( https://docs.rs/tower-http/latest/tower_http/services/struct.ServeDir.html ).

This means that it always follows symlinks.

Attack scenario:

A local user have write permissions to the directory that files are served from, and the webserver process have access to secrets that the user shouldn't be able to read, for example the https certificate with the private key.

Then the user can create a symlink in the directory that is served and exfiltrate those secrets.

## Solution

This solution is not really ready to be merged, as it's implemented as a linux only solution for now, but maybe this can act as a starting point for a discussion on how it should be solved?

I have implemented this using the openat2 syscall, as that is as far as I know the only/best way to avoid a TOCTOU (https://cwe.mitre.org/data/definitions/367.html) race condition. This makes the solution linux only, and would need to be expanded to other platforms, but I don't own to any computers running macos/windows/freebsd. 

The new open function I introduced `canonicalize_and_openat2` is also not async, maybe a good way forward would be for tokio::fs::File to gain an async `open_at()` function?

I'm also a bit unsure on what to do if `ServeFile` points to a symlink, happy to change that behavior.

### DIFF
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -40,9 +40,9 @@ const DEFAULT_CAPACITY: usize = 65536;
 ///   existing file (`/file.html/something`)
 /// - We don't have necessary permissions to read the file
 ///
-/// On linux, if a file is a symlink to a file outside the base directory,
-/// a 500 error will be returned, unless `follow_symlinks_outside_base`
-/// is set to true.
+/// If the requested file is a symlink, it will be followed, even outside `base`.
+/// On Linux this behaviour can be modified to only allow symlinks below `base`
+/// with `follow_symlinks_outside_base`.
 ///
 /// # Example
 ///
@@ -84,7 +84,7 @@ impl ServeDir<DefaultServeDirFallback> {
             },
             fallback: None,
             call_fallback_on_method_not_allowed: false,
-            follow_symlinks_outside_base: false,
+            follow_symlinks_outside_base: true,
         }
     }
 
@@ -99,7 +99,7 @@ impl ServeDir<DefaultServeDirFallback> {
             variant: ServeVariant::SingleFile { mime },
             fallback: None,
             call_fallback_on_method_not_allowed: false,
-            follow_symlinks_outside_base: false,
+            follow_symlinks_outside_base: true,
         }
     }
 }
@@ -257,11 +257,9 @@ impl<F> ServeDir<F> {
         self
     }
 
-    /// Start following symlinks outside the base directory.
-    ///
-    /// Warning: if you have a writeable directory inside the base directory, this
-    /// means that a local user can exfiltrate any file that the server process have
-    /// read access to.
+    /// Disallow following symlinks outside the base directory on Linux.
+    /// Setting this to false on Linux introduces a dependency on at least
+    /// kernel version 5.6, as the openat2 syscall was introduced in that version.
     pub fn follow_symlinks_outside_base(mut self, follow_symlinks_outside_base: bool) -> Self {
         self.follow_symlinks_outside_base = follow_symlinks_outside_base;
         self
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -40,6 +40,10 @@ const DEFAULT_CAPACITY: usize = 65536;
 ///   existing file (`/file.html/something`)
 /// - We don't have necessary permissions to read the file
 ///
+/// On linux, if a file is a symlink to a file outside the base directory,
+/// a 500 error will be returned, unless `follow_symlinks_outside_base`
+/// is set to true.
+///
 /// # Example
 ///
 /// ```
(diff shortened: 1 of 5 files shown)


tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -79,6 +84,7 @@ impl ServeDir<DefaultServeDirFallback> {
             },
             fallback: None,
             call_fallback_on_method_not_allowed: false,
+            follow_symlinks_outside_base: false,
```
seanmonstar: This should default to true. And relevant docs should mention the difference only if enabled.

seanmonstar (changes requested):

tower-http/src/services/fs/serve_dir/mod.rs
alexanderkjall: thanks for the quick review, I guess I was a bit too paranoid. I have flipped this and tried to improve the docs.

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -226,17 +235,75 @@ fn preferred_encoding(
     preferred_encoding
 }
 
+#[cfg(target_os = "linux")]
+fn canonicalize_and_openat2(base_path: &Path, path: &Path) -> io::Result<File> {
```
jlizen, line 239: Currently `ServeDir` uses `tokio::fs::File::open` to avoid blocking the tokio runtime.

`canonicalize_and_openat2` instead uses sync APIs which would block the runtime under load.

Can we shift to using the `tokio::spawn_blocking` API or something else that will avoid problems with blocking?

jlizen, line 239: This will return an `EXDEV` on failure, which then passes through as a `500` error inside our error handling in `future.rs`. We should instead map it to a 404 to avoid leaking information about our system context.

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -249,6 +257,14 @@ impl<F> ServeDir<F> {
         self
     }
 
+    /// Disallow following symlinks outside the base directory on Linux.
+    /// Setting this to false on Linux introduces a dependency on at least
+    /// kernel version 5.6, as the openat2 syscall was introduced in that version.
+    pub fn follow_symlinks_outside_base(mut self, follow_symlinks_outside_base: bool) -> Self {
+        self.follow_symlinks_outside_base = follow_symlinks_outside_base;
```
jlizen, line 264: Is there a reason not to gate this method behind conditional configuration targeting linux? I'm fine to punt on support for other platforms, but I'd rather not have this be a silent non-op.

tower-http/src/services/fs/serve_dir/open_file.rs
jlizen, line 94: Can we also keep HEAD requests from escaping base with symlinks?

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -226,17 +235,75 @@ fn preferred_encoding(
...
+    let (path, base_path2) = if base_path.is_file() || !base_path.exists() {
```
jlizen, line 240: We already know whether we are `ServeVariant::SingleFile` vs `ServeVariant::Directory`, can we just thread that down to skip the fs probe?

jlizen (changes requested):
Hi @alexanderkjall , thanks for those changes, looks good overall and strongly agree with the intent here. I'd be interested in discussing flipping this to a default behavior in a future breaking release, but opt-in is good for now.

I did have some concerns though, see comments. (Also needs some merge conflict resolution against main)

As a side note, I'm wondering, is there a good place to more clearly documenting the "no escape base" semantics for `ServeFile`? For `ServeDir`, its fairly intuitive ("this directory is the boundary"). But for `ServeFile`, it mostly implies, the selected file can't symlink outside its parent, yes?

jlizen:
Hey @seanmonstar , I am considering bringing in [cap-std](https://crates.io/crates/cap-std) to have some reliable cross-platform implementations of the symlink restrictions on Windows, Mac, etc. 

We could easily implement this using native platform APIs with linux (as shown in this PR) and FreeBSD, but it means that we silently lose this guardrail on certain platforms. To hit Mac and Windows, it would be a pretty large code footprint to have race-free implementations in-line, maybe ~1000 LoC.

But, `cap-std` is a somewhat heavy dependency graph... it brings in an older `windows-sys`,  `rustix`, and a fair number of other crates... Rustix at least is likely to be in most dependency graphs anyway?

I don't particularly mind just writing the code inline when I have time, but inlining would anyway bring in many of the same dependencies unless we want to bring in unsafe code for FFI...

We could also just ostrich algorithm it :) Or do linux/free-bsd only. Or do some uglier restrictions like disallow all symlinks for platforms like OSX.

What do you think? There are some *very mild* security implications here... it would require already having fs access to exploit...

PR #659, opened 2026-04-11
fix clippy warnings

alexanderkjall:
## Motivation

When I did my other pull request I noticed some clippy warnings, but thought that they should be fixed as a separate PR. 

## Solution

clippy::match_on_vec_items have been superseded by clippy::indexing_slicing, but that one only produced false positives so I opted to remove it instead.


jplatte (approved):
Thanks!

Issue #660, opened 2026-04-20
ServeDir/ServeFile validate multipart Range requests before rejecting them with 416


shblue21:
- [x] I have looked for existing issues (including closed) about this

### Version

`tower-http 0.6.8`

`cargo tree | grep tower-http`:
- `tower-http v0.6.8`

### Platform

Linux 6.14.0-1017-oracle (Ubuntu 24.04.1)

### Crates

`tower-http` (`services::fs::ServeDir`, `services::fs::ServeFile`)

### Description

`ServeDir` / `ServeFile` do not support multipart range responses, but they still parse and validate multipart `Range` headers before later rejecting them with `416 Range Not Satisfiable`.

Current flow is roughly:

- `open_file.rs::try_parse_range()` calls `http_range_header::parse_range_header(...).and_then(|p| p.validate(file_size))`
- later, `future.rs::build_response()` rejects `ranges.len() > 1` with `416`

So a request like:

```http
Range: bytes=0-0,2-2,4-4,6-6
```

is guaranteed to be rejected, but still pays multipart validation cost first, with the validation step doing `O(n²)` pairwise overlap checks across ranges.

Since multipart responses are not supported here, it seems better to reject `ranges.len() > 1` before calling `validate(file_size)`.

A minimal fix seems to be:

- parse the header
- if there is more than one range, immediately return unsatisfiable
- only call validate(file_size) for single-range requests

seanmonstar:
It sounds reasonable to return an error after noticing a second range. Want to submit a PR?

shblue21:
Sure, I can send a PR.

jlizen:
fixed by #661

PR #661, opened 2026-04-20
fix: reject multipart ranges before validation

shblue21:
## Motivation

`ServeDir` / `ServeFile` do not support multipart range responses, but they validate multi-range `Range` headers before later rejecting them with `416 Range Not Satisfiable`.

This means requests that are guaranteed to be rejected still pay the cost of `validate(file_size)`, including its pairwise overlap checks.

## Solution

Reject multiple ranges before calling `validate(file_size)`:

- parse the range header as before
- return a local `MultipleRangesNotSupported` error when more than one range is present
- only call `validate(file_size)` for single-range requests
- preserve the multipart-specific `416` response body for valid multi-range requests

This also adds regression coverage for the multipart-specific response body.

Note: syntactically parsed multi-range requests that would previously fail semantic validation, such as overlapping or reversed ranges, are now rejected as unsupported multiple ranges before validation. The status and `Content-Range` remain unchanged, but the `416` body becomes the multipart-specific message.

Refs: #660

### DIFF
--- a/tower-http/src/services/fs/serve_dir/future.rs
+++ b/tower-http/src/services/fs/serve_dir/future.rs
@@ -13,6 +13,7 @@ use http::{
     HeaderValue, Request, Response, StatusCode,
 };
 use http_body_util::{BodyExt, Empty, Full};
+use http_range_header::RangeUnsatisfiableError;
 use pin_project_lite::pin_project;
 use std::{
     convert::Infallible,
@@ -320,6 +321,14 @@ fn build_response(output: FileOpened) -> Response<ResponseBody> {
                     .unwrap()
             }
         }
+
+        Some(Err(RangeUnsatisfiableError::OverlappingRanges)) => builder
+            .header(header::CONTENT_RANGE, format!("bytes */{}", size))
+            .status(StatusCode::RANGE_NOT_SATISFIABLE)
+            .body(body_from_bytes(Bytes::from(
+                "Cannot serve multipart range requests",
+            )))
+            .unwrap(),
 
         Some(Err(_)) => builder
             .header(header::CONTENT_RANGE, format!("bytes */{}", size))
--- a/tower-http/src/services/fs/serve_dir/future.rs
+++ b/tower-http/src/services/fs/serve_dir/future.rs
@@ -1,5 +1,5 @@
 use super::{
-    open_file::{FileOpened, FileRequestExtent, OpenFileOutput},
+    open_file::{FileOpened, FileRequestExtent, OpenFileOutput, RangeError},
     DefaultServeDirFallback, ResponseBody,
 };
 use crate::{
@@ -13,7 +13,6 @@ use http::{
     HeaderValue, Request, Response, StatusCode,
 };
 use http_body_util::{BodyExt, Empty, Full};
-use http_range_header::RangeUnsatisfiableError;
 use pin_project_lite::pin_project;
 use std::{
     convert::Infallible,
@@ -322,7 +321,7 @@ fn build_response(output: FileOpened) -> Response<ResponseBody> {
             }
         }
 
-        Some(Err(RangeUnsatisfiableError::OverlappingRanges)) => builder
+        Some(Err(RangeError::MultipleRangesNotSupported)) => builder
             .header(header::CONTENT_RANGE, format!("bytes */{}", size))
             .status(StatusCode::RANGE_NOT_SATISFIABLE)
             .body(body_from_bytes(Bytes::from(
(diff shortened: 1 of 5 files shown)


jplatte:
> Note: this changes the `416` response body for valid multipart range requests from `Cannot serve multipart range requests` to empty, since they now fall through the existing generic `Err(_) -> 416` path in `future.rs`.

I think it would make sense to keep the specific error message body. Can you bring it back?

shblue21:
Thanks. I’ll keep the multipart-specific body so the response stays behavior-preserving. That will require a small `future.rs` update, and I’ll keep the patch as minimal as possible.

shblue21:
Added the small `future.rs` change. Multi-range requests still skip `validate(file_size)`, and the regression test now checks the multipart-specific 416 body.

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -309,8 +309,15 @@ fn try_parse_range(
...
 ) -> Option<Result<Vec<RangeInclusive<u64>>, RangeUnsatisfiableError>> {
     maybe_range_ref.map(|header_value| {
-        http_range_header::parse_range_header(header_value)
-            .and_then(|first_pass| first_pass.validate(file_size))
+        let parsed = http_range_header::parse_range_header(header_value)?;
+
+        if parsed.ranges.len() > 1 {
+            // ServeDir/ServeFile do not support multipart responses, so reject
+            // multi-range requests before validate() runs overlap checks.
+            return Err(RangeUnsatisfiableError::OverlappingRanges);
```
jlizen: This doesn't seem like quite the right error, this refers to overlapping ranges within a multipart request. Meanwhile we don't support them at all.

Can we have a local error type instead that more clearly denotes what's happening, if we preserve this behavior?

jlizen (commented):
@shblue21 do you see a reason not to support the behavior of flattening to single-part GET instead? Even if opt-in due to DOS risk (in which case your approach seems fine for rejection, though I have a small nit on internal error modeling).

IETF seems to prefer that,the [RFC](https://datatracker.ietf.org/doc/html/rfc7233#section-3.1) states:
> A server MAY ignore the Range header field.  However, origin servers
   and intermediate caches ought to support byte ranges when possible,
   since Range supports efficient recovery from partially failed
   transfers and partial retrieval of large representations.  A server
   MUST ignore a Range header field received with a request method other
   than GET.

> Range requests are an OPTIONAL feature of HTTP, designed so that recipients not implementing this feature (or not supporting it for the target resource) can respond as if it is a normal GET request without impacting interoperability.

jlizen:
@shblue21 are you still interested in pursuing this PR?

shblue21:
@jlizen Sorry, I missed this. I’ll take another look at the comments now.

shblue21:
@jlizen For this PR, I’d prefer to keep returning 416 and only skip the unnecessary validation. We can discuss ignoring the `Range` header separately.

tower-http/src/services/fs/serve_dir/future.rs
jlizen: This appears to be dead code now?

tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -683,6 +686,27 @@ async fn read_partial_errs_on_bad_range() {
     )
 }
 
+#[tokio::test]
```
jlizen, line 689: This tests non-overlapping multi-range, but this PR is mostly touching the overlapping case (which previously reached `validate()`). It's good to have, thanks, but we should cover the new behavior as well.

(IE, this test would pass on `main` even without this change)

tower-http/src/services/fs/serve_dir/future.rs
jlizen: Also appears unreachable?

tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -683,6 +686,27 @@ async fn read_partial_errs_on_bad_range() {
...
+async fn multipart_range_valid_returns_multipart_error_body() {
+    let svc = ServeDir::new("..");
```
jlizen: Can we use `REPO_ROOT` and `README_PATH` instead?

jlizen (changes requested):
The only blocking feedback is that we don't have proper test coverage for this new behavior, the rest are quick nits.

I'm fine to punt on changing away from the 416 representation in favor of this refactor staying perf only.Do you have cycles to file the issue? I can if not. Ideally we would serve the non-range 200 where possible to stay RFC compliant. 

Also given such a change change, for cases where we do still have an unsatisfiable 416, we are improperly preserving content headers here (`Content-Type`, `Content-Encoding`, etc), which could result in the plain text (non-gzipped) body being improperly parsed if the original request was with gzip or a different content type.

(And the above header issue applies to the present handling too, but it's fine to keep your PR scoped more narrowly to saving the extra `validate()` if preferred)

jlizen (approved):
thanks!

PR #667, opened 2026-05-05
fix: restore tokio and async-compression as no-op features

jlizen:
As discussed in: https://github.com/tower-rs/tower-http/pull/666

Adding back in the previously implicit `tokio` and `async-compression` features as no-ops. We can rip them out come `0.7.0`.

In the meantime, no need to break users that (for whatever reason) might be naming `tower-http = ["tokio"]` while using some other dependency that requires `tower-http = ">= 0.6.9"`.


jplatte (approved):

jlizen:
I'll make sure this goes out by end of week, or shortly after #559 lands, whichever comes first.

PR #672, opened 2026-05-07
feat(set_header): refactor and improve multiple header middleware

### LINKED ISSUE #571: Add multiple headers in SetResponseHeaderLayer or similar

## Feature Request

I want to be able to set multiple headers with `SetResponseHeaderLayer`: needing one layer per header (plus me having issues returning `impl Layer<S>` relating to `Send`, `Sync`, etc), means that to set a bunch of default headers in a separate function than the one defining the service itself requires naming a type that is really long, made of a pile of `tower::layer::util::Stack`.

The use case for this is stuff like X-Content-Type-Options, X-Frame-Options, etc.

### Motivation

There's a crate `tower-default-headers` which does this, but it's a third party crate and a pretty common use case.

### Proposal

Some kind of `SetResponseHeadersLayer` that allows setting multiple headers, either using a Vec of them or a compile-time array or so.

### Alternatives

Just use the third party crate;
Just write your own middleware which involves unfun async shenanigans.

seun-ja:
## Motivation

Fixes #571

Offering the ability to add multiple headers at once

## Solution

For better code organisation, the `response.rs` file, which handles the response request, is now a module and split into two files. One for setting a single header and the other for multiple, while still maintaining the current API structure.

The functions handling setting headers take a dynamic variable that implements `HeaderMetadataGenerator`. The newly created `HeaderMetadata` and tuple of `(HeaderName, M)` currently implemented in `M` can be `HeaderValue` (A test case covers this). This also helps with maintainability

### DIFF
--- a/tower-http/src/set_header/response/multiple_header.rs
+++ b/tower-http/src/set_header/response/multiple_header.rs
@@ -1,4 +1,8 @@
-use http::{header::HeaderName, Request, Response};
+//! Set multiple headers on the response.
+//!
+//! See the root [`crate::set_header::response`] module for full documentation and usage examples.
+//!
+use http::{header::HeaderName, HeaderValue, Request, Response};
 use pin_project_lite::pin_project;
 use std::{
     fmt,
@@ -11,37 +15,74 @@ use tower_service::Service;
 
 use crate::set_header::{InsertHeaderMode, MakeHeaderValue};
 
-pub trait HeaderMetadataGenerator<M> {
-    fn to_metadata(self) -> HeaderMetadata<M>;
+/// A trait that combines MakeHeaderValue and Clone capability for trait objects.
+trait CloneableMakeHeaderValue<T>: MakeHeaderValue<T> + Send + Sync {
+    fn clone_box(&self) -> Box<dyn CloneableMakeHeaderValue<T>>;
+}
+
+impl<T, M> CloneableMakeHeaderValue<T> for M
+where
+    M: MakeHeaderValue<T> + Clone + Send + Sync + 'static,
+{
+    fn clone_box(&self) -> Box<dyn CloneableMakeHeaderValue<T>> {
+        Box::new(self.clone())
+    }
+}
+
+/// A "Bridge" struct that allows for trait object-based header value generation.
+struct BoxedMakeHeaderValue<T>(Box<dyn CloneableMakeHeaderValue<T>>);
+
+impl<T> BoxedMakeHeaderValue<T> {
+    /// Create a new BoxedMakeHeaderValue from any maker that implements MakeHeaderValue and Clone.
+    fn new<M>(maker: M) -> Self
+    where
+        M: MakeHeaderValue<T> + Clone + Send + Sync + 'static,
+    {
+        Self(Box::new(maker))
+    }
+}
+
+impl<T> Clone for BoxedMakeHeaderValue<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone_box())
+    }
+}
+
+impl<T> MakeHeaderValue<T> for BoxedMakeHeaderValue<T> {
+    fn make_header_value(&mut self, message: &T) -> Option<HeaderValue> {
+        self.0.make_header_value(message)
+    }
+}
+
+impl<T> fmt::Debug for BoxedMakeHeaderValue<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("BoxedMakeHeaderValue").finish()
+    }
 }
 
 /// Metadata describing a response header to be set by [`SetMultipleResponseHeadersLayer`] or [`SetMultipleResponseHeader`].
 #[derive(Clone, Debug)]
-pub struct HeaderMetadata<M> {
+pub struct HeaderMetadata<T> {
     /// The name of the header to set.
-    pub header_name: HeaderName,
+    header_name: HeaderName,
     /// The value or value factory for the header.
-    pub make: M,
+    make: BoxedMakeHeaderValue<T>,
 }
 
-impl<M> HeaderMetadataGenerator<M> for HeaderMetadata<M> {
-    fn to_metadata(self) -> HeaderMetadata<M> {
-        self
-    }
-}
-
-impl<M> HeaderMetadataGenerator<M> for (HeaderName, M) {
-    fn to_metadata(self) -> HeaderMetadata<M> {
-        HeaderMetadata {
-            header_name: self.0,
-            make: self.1,
+impl<T> HeaderMetadata<T> {
+    /// Create a new HeaderMetadata with the given header name and value factory.
+    fn new<M: CloneableMakeHeaderValue<T> + Clone + 'static>(
+        header_name: HeaderName,
+        make: M,
+    ) -> Self {
+        Self {
+            header_name,
+            make: BoxedMakeHeaderValue::new(make),
         }
     }
-}
 
-impl<M> HeaderMetadata<M> {
     /// Convert this metadata into a [`HeaderInsertionConfig`] with the given insertion mode.
-    fn convert_to_header_config(self, mode: InsertHeaderMode) -> HeaderInsertionConfig<M> {
+    fn build_config(self, mode: InsertHeaderMode) -> HeaderInsertionConfig<T> {
         HeaderInsertionConfig {
             header_name: self.header_name,
             make: self.make,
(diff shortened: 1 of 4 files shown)


tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+
+use crate::set_header::{InsertHeaderMode, MakeHeaderValue};
+
+pub trait HeaderMetadataGenerator<M> {
+    fn to_metadata(self) -> HeaderMetadata<M>;
+}
+
+/// Metadata describing a response header to be set by [`SetMultipleResponseHeadersLayer`] or [`SetMultipleResponseHeader`].
+#[derive(Clone, Debug)]
+pub struct HeaderMetadata<M> {
```
jlizen: The `M` type parameter makes this a less useful than it seems, since you can't mix types. So if you want eg a static `Content-Type` but a closure-derived `Content-Length`, you are out of luck.

Another approach might be a builder that collects heterogeneous header constructors and type erase.

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+    future::Future,
+    pin::Pin,
+    task::{ready, Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+use crate::set_header::{InsertHeaderMode, MakeHeaderValue};
+
+pub trait HeaderMetadataGenerator<M> {
```
jlizen: This is unnecessary, we can just implement `Into`

tower-http/src/set_header/response/single_header.rs
```diff
@@ -1,100 +1,3 @@
...
-//! let response = svc.ready().await?.call(request).await?;
-//!
-//! assert_eq!(response.headers()["content-length"], "10");
-//! #
-//! # Ok(())
-//! # }
-//! ```
-
-use super::{InsertHeaderMode, MakeHeaderValue};
 use http::{header::HeaderName, Request, Response};
```
jlizen, line 5: We should add at least some simple docs for each submodule that points back to root for bigger docs

tower-http/src/set_header/response/multiple_header.rs
jlizen: If we did want it for some reason, we should also re-export it from root. I don't think we need it.

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+struct HeaderInsertionConfig<M> {
+    header_name: HeaderName,
+    make: M,
+    mode: InsertHeaderMode,
+}
+
+/// Layer that applies [`SetMultipleResponseHeader`] which adds multiple response headers.
+///
+/// See [`SetMultipleResponseHeader`] for more details.
+pub struct SetMultipleResponseHeadersLayer<M> {
```
jlizen: This should be re-exported from module root for consistency.

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+        SetMultipleResponseHeader {
+            inner,
+            headers: self.headers.clone(),
+        }
+    }
+}
+
+/// Middleware that sets multiple headers on the response.
+#[derive(Clone)]
+pub struct SetMultipleResponseHeader<S, M> {
```
jlizen: This should be re-exported from module root for consistency.

tower-http/src/set_header/response/multiple_header.rs
jlizen: Should we validate for header name collisions? Currently last one wins.

I could be convinced either way, but for whichever, we should document it clearly in the relevant places... where do you suggest?

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+
+pub trait HeaderMetadataGenerator<M> {
+    fn to_metadata(self) -> HeaderMetadata<M>;
+}
+
+/// Metadata describing a response header to be set by [`SetMultipleResponseHeadersLayer`] or [`SetMultipleResponseHeader`].
+#[derive(Clone, Debug)]
+pub struct HeaderMetadata<M> {
+    /// The name of the header to set.
+    pub header_name: HeaderName,
```
jlizen: I'd prefer a constructor or builder pattern here to have the fields stay private

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+        HeaderMetadata {
+            header_name: self.0,
+            make: self.1,
+        }
+    }
+}
+
+impl<M> HeaderMetadata<M> {
+    /// Convert this metadata into a [`HeaderInsertionConfig`] with the given insertion mode.
+    fn convert_to_header_config(self, mode: InsertHeaderMode) -> HeaderInsertionConfig<M> {
```
jlizen: nit: `with_mode()`?

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+/// Layer that applies [`SetMultipleResponseHeader`] which adds multiple response headers.
+///
+/// See [`SetMultipleResponseHeader`] for more details.
+pub struct SetMultipleResponseHeadersLayer<M> {
+    headers: Vec<HeaderInsertionConfig<M>>,
+}
+
+impl<M> fmt::Debug for SetMultipleResponseHeadersLayer<M> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        for metadata in &self.headers {
```
jlizen: this would produce malformed output. use `debug_list` or a single `debug_struct` with a `headers` field

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,421 @@
...
+            header
+                .mode
+                .apply(&header.header_name, &mut res, &mut header.make);
+        }
+
+        Poll::Ready(Ok(res))
+    }
+}
+
+#[cfg(test)]
```
jlizen, line 320: More tests I'd like to see:
- construction with the layer
- multiple different headers in the same vec
- empty vec
- both static and closure-based headers (once working)

jlizen (changes requested):
I like the intent here, but there are a few issues as written, see below.

tower-http/src/set_header/response/multiple_header.rs
jlizen: Rough example:
```
  struct ErasedMakeHeaderValue<B> {
      make: Box<dyn FnMut(&Response<B>) -> Option<HeaderValue> + Send + Sync>,
  }

  pub struct SetResponseHeadersLayer<B> {
      headers: Vec<(HeaderName, ErasedMakeHeaderValue<B>, InsertHeaderMode)>,
  }

  impl<B> SetResponseHeadersLayer<B> {
      pub fn new() -> Self {
          Self { headers: vec![] }
      }

      /// Add a header that overrides any existing value.
      pub fn overriding(mut self, name: HeaderName, make: M) -> Self
      where
          M: MakeHeaderValue<Response<B>> + Send + Sync + 'static,
      {
          self.headers.push((name, erase(make), InsertHeaderMode::Override));
          self
      }

      /// Add a header that appends to existing values.
      pub fn appending(mut self, name: HeaderName, make: M) -> Self
      where
          M: MakeHeaderValue<Response<B>> + Send + Sync + 'static,
      {
          self.headers.push((name, erase(make), InsertHeaderMode::Append));
          self
      }

      /// Add a header only if not already present.
      pub fn if_not_present(mut self, name: HeaderName, make: M) -> Self
      where
          M: MakeHeaderValue<Response<B>> + Send + Sync + 'static,
      {
          self.headers.push((name, erase(make), InsertHeaderMode::IfNotPresent));
          self
      }
  }

  Usage:

  ServiceBuilder::new()
      .layer(
          SetResponseHeadersLayer::new()
              .if_not_present(X_CONTENT_TYPE_OPTIONS, HeaderValue::from_static("nosniff"))
              .if_not_present(X_FRAME_OPTIONS, HeaderValue::from_static("DENY"))
              .overriding(CONTENT_LENGTH, |res: &Response<Body>| {
                  res.body().size_hint().exact().map(|s|
  HeaderValue::from_str(&s.to_string()).unwrap())
              })
      )
      .service(inner)
```

That pays the cost of the boxing once, on construction, but anyway we are cloning the `Vec` every reqeuest, so it's just a bit of extra vtable dispatch on the order of nanoseconds.

tower-http/src/set_header/response/multiple_header.rs
seun-ja: Yeah, I'm taking that out, implementing `From` is better. I should have done a cleanup for that; my initial approach required that. Thanks for catching that

tower-http/src/set_header/response/multiple_header.rs
seun-ja: I agree there should be some validation. I think it should be restricted to `InsertHeaderMode::Append` mode. 

Does that sound good?

seun-ja: Shouldn't this be a different issue? 

I would suggest just a warn, with something like this

```Rust
InsertHeaderMode::Append => {
  if let Some(value) = make.make_header_value(target) {
      if target.headers().contains_key(header_name) {
          #[cfg(feature = "tracing")]
          tracing::warn!(
              "Appending header `{}`: existing value(s) will be preserved; consider using `Override` mode if you want to replace instead.",
              header_name
          );
      }
      target.headers_mut().append(header_name.clone(), value);
  }
}
```

jlizen: a warn seems good to me, gated by the tracing feature

tower-http/src/set_header/response/multiple_header.rs
```diff
@@ -0,0 +1,533 @@
...
+            header_name: self.header_name,
+            make: self.make,
+            mode,
+        }
+    }
+}
+
+impl<T, M> From<(HeaderName, M)> for HeaderMetadata<T>
+where
+    M: CloneableMakeHeaderValue<T> + Clone + 'static,
```
jlizen: We should do this based on the public traits instead:

```
  impl<T, M> From<(HeaderName, M)> for HeaderMetadata<T>
  where
      M: MakeHeaderValue<T> + Clone + Send + Sync + 'static,
  {
      fn from((header_name, make): (HeaderName, M)) -> Self {
          HeaderMetadata::new(header_name, make)
      }
  }

```

tower-http/src/set_header/response/multiple_header.rs
jlizen: Nevermind, I think the way you have it with users explicitly opting into the mode, is fine. If we did have a warn, we should make it rate limited or only at debug level, which is probably more irritating than its worth.

jlizen (changes requested):
@seun-ja changes look great! Just one last bit of cleanup

jlizen (approved):
Great stuff! Thanks for the contribution. Will get a release out.

seun-ja:
Hi @jlizen, can I add an issue to implement this for request headers?

jlizen:
@seun-ja definitely, thanks!

PR #674, opened 2026-05-09
Relaxed Vary header defaults

### LINKED ISSUE #539: CORS module excessively sets the `Vary` response header
`cors::CorsLayer` sets the `Vary` response header to `origin, access-control-request-method, access-control-request-headers` (added in #199) regardless of the provided CORS configuration.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary):

> The HTTP Vary response header describes the parts of the request message (aside from the method and URL) that influenced the content of the response it occurs in. Including a Vary header ensures that responses are separately cached based on the headers listed in the Vary field.

`cors::CorsLayer` should only add a request header to the `Vary` response header if a dynamic response is calculated based on that request header. `.permissive()`, for example, does not dynamically adjust the response at all.

In general, dynamic behaviour appears to occur through various `.mirror_request()` methods. It can also occur when closures are passed, though there it isn't trivial to determine what request headers have been used to determine the response; that case may instead require a call-out in the documentation.

seun-ja:
## Motivation

Fixes #539 

Relaxes the default Vary header to an empty list.

## Solution

Instead of the custom `Default` implementation adding some header values, it now simply returns the out-of-the-box `Default`, which is an empty Vector.

Also, I altered the test to reflect the new change and an additional test.

### DIFF
--- a/tower-http/src/cors/allow_methods.rs
+++ b/tower-http/src/cors/allow_methods.rs
@@ -73,7 +73,7 @@ impl AllowMethods {
     }
 
     pub(super) fn varies_with_request_method(&self) -> bool {
-        matches!(&self.0, AllowMethodsInner::MirrorRequest)
+        !matches!(&self.0, AllowMethodsInner::Const(_))
     }
 
     pub(super) fn to_header(&self, parts: &RequestParts) -> Option<(HeaderName, HeaderValue)> {
--- a/tower-http/src/cors/allow_methods.rs
+++ b/tower-http/src/cors/allow_methods.rs
@@ -72,6 +72,10 @@ impl AllowMethods {
         matches!(&self.0, AllowMethodsInner::Const(Some(v)) if v == WILDCARD)
     }
 
+    pub(super) fn varies_with_request_method(&self) -> bool {
+        matches!(&self.0, AllowMethodsInner::MirrorRequest)
+    }
+
     pub(super) fn to_header(&self, parts: &RequestParts) -> Option<(HeaderName, HeaderValue)> {
         let allow_methods = match &self.0 {
             AllowMethodsInner::Const(v) => v.clone()?,
--- a/tower-http/src/cors/mod.rs
+++ b/tower-http/src/cors/mod.rs
@@ -431,12 +431,20 @@ impl CorsLayer {
 
     /// Set the value(s) of the [`Vary`][mdn] header.
     ///
-    /// In contrast to the other headers, this one has a non-empty default of
-    /// [`preflight_request_headers()`].
+    /// By default, this value is derived from whether CORS response headers are
+    /// request-dependent:
     ///
-    /// You only need to set this if you want to remove some of these defaults,
-    /// or if you use a closure for one of the other headers and want to add a
-    /// vary header accordingly.
+    /// - `Origin` is included when `Access-Control-Allow-Origin` depends on the
+    ///   request's `Origin` header (for example, origin lists or predicates).
+    /// - `Access-Control-Request-Method` is included when
+    ///   `Access-Control-Allow-Methods` mirrors `Access-Control-Request-Method`.
+    /// - `Access-Control-Request-Headers` is included when
+    ///   `Access-Control-Allow-Headers` mirrors `Access-Control-Request-Headers`.
+    /// - If none of those values are request-dependent, no `Vary` header is
+    ///   added.
+    ///
+    /// Calling this method sets `Vary` explicitly and pins it to the provided
+    /// value, regardless of future changes to those other CORS settings.
     ///
     /// [mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary
     pub fn vary<T>(mut self, headers: T) -> Self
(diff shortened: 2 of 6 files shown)


jplatte:
This is worse than the current state of things IMO.  Has the current default caused some specific issue for you?

seun-ja:
No @jplatte.

I haven't had any issue using it in this context. 

I'm just fixing because of the issue raised in #539 and seems it was ok to fix as you were ok with PR on it

jlizen:
++ to @jplatte 's feedback that this will break a lot of users since we are shifting from "conservative" defaults (include `vary` headers unnecessarily), to *empty* as a default, meaning we force anybody with dynamic CORS to manually set the header or risk caching correctness problems. Definitely a breaking change, probably not a justifiable one.

Would it work to instead have `permissive()` implicitly set `.vary(Vary::list([]))` rather than shifting defaults globally?

Or are there other overly conservative cases you are concerned with @seun-ja ?

jplatte:
So, what I was thinking of when writing the referenced comment was having the various methods like `allow_origin` subtract from the `vary` list (only if not customized), if and only if the input is one that doesn't result in different header values for different requests. For `AllowOrigin` this would be `any` and `exact` (`AllowOriginInner::Const`). For `AllowHeaders` it would be `any` and `list` (`AllowHeadersInner::Const`).

jlizen (changes requested):
The approach of eagerly stripping the vary headers based on `permissive()` creates a problem:

```
CorsLayer::permissive().allow_origin(AllowOrigin::predicate(...))
```

This would produce no `vary: origin` header, when in fact we might have dynamic CORS that has multiple allowed origins. That could result in a stale cache improperly rejecting origins that DO match the predicate.

The proper way to handle this would be, like jplatte@ suggested, doing the stripping on the explicit `allow_*` methods.

And then, we also would need an extra `Is_custom_vary` parameter in the builder (or something), to make sure that we don't overwrite the user's explicit configuration.

tower-http/src/cors/mod.rs
```diff
@@ -221,6 +201,26 @@ impl CorsLayer {
...
+                if !vary.to_header().map_or(false, |(_, v)| {
+                    v.to_str()
+                        .unwrap_or("")
+                        .split(',')
+                        .any(|s| s.trim().eq_ignore_ascii_case(name.as_str()))
+                }) {
+                    vary = Vary::list([name]);
+                }
+                self.vary = vary;
+            }
```
jplatte: Instead of this complicated logic to add back the vary header if it has been removed from the default list, I think we could just do the checks to omit some vary headers in the `Layer` impl for `CorsLayer` (and additionally the `Service` impl of `Cors`, since that one also has the builder-style methods, though they are probably much less frequently used). What do you think?

seun-ja: Yeah this makes more sense. 

It's a lot cleaner and maintainable

tower-http/src/cors/mod.rs
```diff
@@ -493,10 +493,34 @@ impl<S> Layer<S> for CorsLayer {
...
         ensure_usable_cors_rules(self);
 
-        Cors {
-            inner,
-            layer: self.clone(),
+        // Clone the layer to modify Vary header logic
+        let mut layer = self.clone();
+
+        // Only set Vary if not custom
+        if !layer.is_vary_custom {
```
jlizen: Can we extract this logic into a shared helper that can be used across both `Layer::layer()` and `Cors::map_layer()`

tower-http/src/cors/mod.rs
```diff
@@ -493,10 +493,34 @@ impl<S> Layer<S> for CorsLayer {
...
+            // If all origins, methods, and headers are allowed, omit Vary
+            let all_origins = layer.allow_origin.is_wildcard();
```
jlizen: This logic looks fine in that it won't ever remove Vary headers from potentially dynamic CORS configurations.

But, it could be broadened to also include eg: `AllowOrigin::exact("http://example.com")`.

Also the list method for headers/methods are constant (return all values), only the origin will vary what is returned per request.

The current approach is fine as a first step, optionally you could leave a TODO to sweep up the rest, or just add the others in now.

tower-http/src/cors/tests.rs
```diff
@@ -1,37 +1,106 @@
...
 
+const INITIAL_VARY_HEADERS: HeaderValue = HeaderValue::from_static("accept, accept-encoding");
+const ADDITIONAL_VARY_HEADERS: [HeaderName; 3] = [
+    header::ORIGIN,
+    header::ACCESS_CONTROL_REQUEST_METHOD,
+    header::ACCESS_CONTROL_REQUEST_HEADERS,
+];
+
+#[tokio::test]
+#[allow(
```
jlizen: I don't think we need these clippy annotations on this test. Can we flip them over to `expect()` for all usages in this file to avoid future buildup?

tower-http/src/cors/tests.rs
```diff
@@ -1,37 +1,106 @@
 use std::convert::Infallible;
```
jlizen, line 1: Can we also have tests for:
- "mixed" case where eg origin is a wildcard, but headers/method include vary headers
- very_permissive emitting vary headers
- at least a couple smoke tests on the Cors::map_layer test with and without vary headers

tower-http/src/cors/mod.rs
jlizen: This should be updated to explain that the default is derived from which other configuration is set, and that setting it explicitly will pin it regardless of other configuration.

jlizen (changes requested):
Coming together! Down to just small tweaks.

tower-http/src/cors/mod.rs
```diff
@@ -493,10 +525,13 @@ impl<S> Layer<S> for CorsLayer {
...
         ensure_usable_cors_rules(self);
 
-        Cors {
-            inner,
-            layer: self.clone(),
-        }
+        // Clone the layer to modify Vary header logic
+        let mut layer = self.clone();
+
+        // Only set Vary if not custom
```
jlizen: @seun-ja This comment needs cleaning up in both blocks, it belongs in the helper. Looks ready to merge otherwise.

tower-http/src/cors/allow_methods.rs
```diff
@@ -72,6 +72,10 @@ impl AllowMethods {
         matches!(&self.0, AllowMethodsInner::Const(Some(v)) if v == WILDCARD)
     }
 
+    pub(super) fn varies_with_request_method(&self) -> bool {
+        matches!(&self.0, AllowMethodsInner::MirrorRequest)
```
jlizen: Nit: I'd probably prefer method/header checks to be a negation of the known const methods on these, to be defensive against some future dynamic variant that misses updating this check (or maybe we switch to a dynamic list). Similar to how you have origin now. Better to be conservative about serving vary headers we don't need.

Unlikely, but who knows, harmless to do it that way.

jlizen (approved):
Appreciate the contribution!

PR #677, opened 2026-05-12
feat: set multiple request header

### LINKED ISSUE #676: Add multiple headers in SetRequestHeaderLayer

## Feature Request

One should be able to set multiple headers with SetRequestHeaderLayer.

### Motivation

It's inspired by the #571 (closed by #672), which only makes it possible to add multiple headers for Responses via SetResponseHeaderLayer.

### Proposal

Would pretty much be the request version of what is done with the response, still maintain backward compatibility

### Alternatives

Use third-party crates like `tower-default-headers` or write your own middleware, which involves repetitive layers.

seun-ja:
## Motivation

Fixes #676

Adds native support for inserting multiple header middleware per layer

## Solution

For better code organisation, the request.rs file, which handles the request, is now a module and split into two files. One for setting a single header and the other for multiple, while still maintaining the current API structure.

The functions handling setting headers take a dynamic variable that implements HeaderMetadataGenerator. The newly created HeaderMetadata and a tuple of (HeaderName, M), currently implemented in `M`, can be HeaderValue (A test case covers this). This also helps with maintainability.

It also includes tests

The PR also moved some reused code into the root module for easy access by both request and response headers

### DIFF
--- a/tower-http/src/set_header/mod.rs
+++ b/tower-http/src/set_header/mod.rs
@@ -12,7 +12,7 @@ pub mod response;
 #[doc(inline)]
 pub use self::{
     request::{SetRequestHeader, SetRequestHeaderLayer},
-    response::{HeaderMetadata, SetResponseHeader, SetResponseHeaderLayer},
+    response::{SetResponseHeader, SetResponseHeaderLayer},
 };
 
 /// Trait for producing header values.
@@ -153,5 +153,75 @@ impl<T> MakeHeaderValue<T> for BoxedMakeHeaderValue<T> {
 impl<T> fmt::Debug for BoxedMakeHeaderValue<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("BoxedMakeHeaderValue").finish()
+    }
+}
+
+/// Metadata describing a response header to be set.
+#[derive(Clone, Debug)]
+pub struct HeaderMetadata<T> {
+    /// The name of the header to set.
+    header_name: HeaderName,
+    /// The value or value factory for the header.
+    make: BoxedMakeHeaderValue<T>,
+}
+
+impl<T> HeaderMetadata<T> {
+    /// Create a new HeaderMetadata with the given header name and value factory.
+    fn new<M: MakeHeaderValue<T> + Clone + 'static + Send + Sync>(
+        header_name: HeaderName,
+        make: M,
+    ) -> Self {
+        Self {
+            header_name,
+            make: BoxedMakeHeaderValue::new(make),
+        }
+    }
+
+    /// Convert this metadata into a [`HeaderInsertionConfig`] with the given insertion mode.
+    fn build_config(self, mode: InsertHeaderMode) -> HeaderInsertionConfig<T> {
+        HeaderInsertionConfig {
+            header_name: self.header_name,
+            make: self.make,
+            mode,
+        }
+    }
+}
+
+impl<T, M> From<(HeaderName, M)> for HeaderMetadata<T>
+where
+    M: MakeHeaderValue<T> + Clone + 'static + Send + Sync,
+{
+    fn from((header_name, make): (HeaderName, M)) -> Self {
+        HeaderMetadata::new(header_name, make)
+    }
+}
+
+/// Configuration for inserting a header into a response or request.
+struct HeaderInsertionConfig<T> {
+    header_name: HeaderName,
+    make: BoxedMakeHeaderValue<T>,
+    mode: InsertHeaderMode,
+}
+
+impl<T> Clone for HeaderInsertionConfig<T>
+where
+    BoxedMakeHeaderValue<T>: Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            header_name: self.header_name.clone(),
+            make: self.make.clone(),
+            mode: self.mode,
+        }
+    }
+}
+
+impl<T> fmt::Debug for HeaderInsertionConfig<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("HeaderInsertionConfig")
+            .field("header_name", &self.header_name)
+            .field("mode", &self.mode)
+            .field("make", &"BoxedMakeHeaderValue")
+            .finish()
     }
 }
--- a/tower-http/src/set_header/mod.rs
+++ b/tower-http/src/set_header/mod.rs
@@ -2,6 +2,8 @@
 //!
 //! See [request] and [response] for more details.
 
+use std::fmt;
+
 use http::{header::HeaderName, HeaderMap, HeaderValue, Request, Response};
 
 pub mod request;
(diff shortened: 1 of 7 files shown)


tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -0,0 +1,422 @@
...
+    task::{Context, Poll},
+};
+use tower_layer::Layer;
+use tower_service::Service;
+
+use crate::set_header::{BoxedMakeHeaderValue, InsertHeaderMode, MakeHeaderValue};
+
+/// Metadata describing a request header to be set by [`SetMultipleRequestHeadersLayer`] or [`SetMultipleRequestHeader`].
+#[derive(Clone, Debug)]
+pub struct HeaderMetadata<T> {
```
jlizen: We are going to the trouble of lifting `BoxedMakeHeaderValue` and friends into a shared module, can we not lift out `HeaderMetadata` and `HeaderInsertionConfig` as well?

They have effectively identical structures, usages, and doc comments.

It also reduces the surface area where we need to have parallel tests for everything (and also the number of places we can have copy/paste bugs ^^)

tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -0,0 +1,422 @@
+//! Set multiple headers on the request.
```
jlizen, line 1: This is `multiple_headers.rs` but response is `multiple_header.rs`. I prefer `multiple_headers`. We can safely rename the response file name without breakage, it looks like?

tower-http/src/set_header/mod.rs
jlizen, line 14: we should be re-exporting consistently across both (or ideally `HeaderMetadata` is unified anyway)

tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -0,0 +1,422 @@
...
+}
+
+/// Layer that applies [`SetMultipleRequestHeader`] which adds multiple request headers.
+///
+/// See [`SetMultipleRequestHeader`] for more details.
+pub struct SetMultipleRequestHeadersLayer<T> {
+    headers: Vec<HeaderInsertionConfig<T>>,
+}
+
+impl<T> fmt::Debug for SetMultipleRequestHeadersLayer<T> {
```
jlizen: We should implement Clone for this where `T: Clone`... 
That lets you do, eg:
```
let layer = ...;
svc1.layer(layer.clone()); svc2.layer(layer.clone())
```

Refer to `SetRequestHeaderLayer` for an example.

I missed this in the `response` PR, we need it there too. I'd be fine to have it in this PR.

tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -0,0 +1,422 @@
...
+            }));
+    }
+
+    #[tokio::test]
+    async fn test_layer_with_static_and_closure_headers_fixed() {
+        // Wrap the static value
+        let static_meta = (header::CONTENT_TYPE, HeaderValue::from_static("text/html")).into();
+
+        // Wrap the closure
+        let closure_meta = (header::X_FRAME_OPTIONS, |_res: &Response<Body>| {
```
jlizen: This is passing a response body closure into a layer that expects a request body. It only works because there is no actual usage of this service.  Let's add a `oneshot.await()` to make sure we flush it out.

We already are doing this properly on the response side.

tower-http/src/set_header/request/multiple_headers.rs
jlizen: The big win here would be to have just a single: 
`tower_http::set_header::HeaderMetadata`

Rather than two different types with the same name, that behave identically.

I imagine the colliding name is part of why only `response` is re-exporting that type...

jlizen (changes requested):

tower-http/src/set_header/request/multiple_headers.rs
seun-ja: Yes, I realised late and wasn't sure if it was fine to fix in this PR

tower-http/src/set_header/request/mod.rs
```diff
@@ -89,7 +89,8 @@
 //! ```
 //! use http::{Request, Response, header::{self, HeaderValue}};
 //! use tower::{Service, ServiceExt, ServiceBuilder};
-//! use tower_http::set_header::request::{SetMultipleRequestHeadersLayer, HeaderMetadata};
+//! use tower_http::set_header::request::{SetMultipleRequestHeadersLayer};
```
jlizen: let's clean this up into a single `tower_http` import with `request::SetMultipleRequestHeadersLayer` inline

tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -2,106 +2,51 @@
...
-///
-/// See [`SetMultipleRequestHeader`] for more details.
-pub struct SetMultipleRequestHeadersLayer<T> {
-    headers: Vec<HeaderInsertionConfig<T>>,
-}
-
-impl<T> fmt::Debug for SetMultipleRequestHeadersLayer<T> {
+impl<M> fmt::Debug for SetMultipleRequestHeadersLayer<M>
+where
+    M: Clone,
```
jlizen: we don't need clone bounds on the debug, only the clone impl

tower-http/src/set_header/response/mod.rs
```diff
@@ -88,7 +88,8 @@
 //! ```
 //! use http::{Request, Response, header::{self, HeaderValue}};
 //! use tower::{Service, ServiceExt, ServiceBuilder};
-//! use tower_http::set_header::response::{SetMultipleResponseHeadersLayer, HeaderMetadata};
+//! use tower_http::set_header::response::{SetMultipleResponseHeadersLayer};
```
jlizen: let's consolidate the `tower_http` imports

tower-http/src/set_header/response/mod.rs
```diff
@@ -122,7 +123,8 @@
...
```
jlizen: let's consolidate the `tower_http` imports

tower-http/src/set_header/response/multiple_headers.rs
```diff
@@ -13,98 +13,32 @@ use std::{
...
-pub struct SetMultipleResponseHeadersLayer<T> {
-    headers: Vec<HeaderInsertionConfig<T>>,
+pub struct SetMultipleResponseHeadersLayer<M> {
+    headers: Vec<HeaderInsertionConfig<M>>,
 }
 
-impl<T> fmt::Debug for SetMultipleResponseHeadersLayer<T> {
+impl<M> fmt::Debug for SetMultipleResponseHeadersLayer<M>
+where
+    M: Clone,
```
jlizen: don't need clone bound on debug, do need the standalone clone impl with clone bound

tower-http/src/set_header/mod.rs
```diff
@@ -155,3 +155,73 @@ impl<T> fmt::Debug for BoxedMakeHeaderValue<T> {
         f.debug_struct("BoxedMakeHeaderValue").finish()
     }
 }
+
+/// Metadata describing a response header to be set.
```
jlizen: Either remove `response`, or say `request or response`

tower-http/src/set_header/response/mod.rs
jlizen, line 84: Can we have a note here, and also for request, on performance? Something like:
```
  //! Note: this layer uses boxing (allocation + dynamic dispatch) to support mixed producer
  //! types in a single vec. Stacking multiple [`SetResponseHeaderLayer`] avoids this at the
  //! cost of a more complex composed service type.
```

jlizen (changes requested):

tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -0,0 +1,374 @@
...
+use crate::set_header::{HeaderInsertionConfig, HeaderMetadata, InsertHeaderMode};
+
+/// Layer that applies [`SetMultipleRequestHeader`] which adds multiple request headers.
+///
+/// See [`SetMultipleRequestHeader`] for more details.
+pub struct SetMultipleRequestHeadersLayer<M> {
+    headers: Vec<HeaderInsertionConfig<M>>,
+}
+
+impl<M> Clone for SetMultipleRequestHeadersLayer<M>
```
jlizen: Can we implement this for response too? I also think we can just `#[derive(Clone)]` for these, actually...

jlizen (changes requested):
Looks good, one last Clone impl needed, sorry to scope creep you.

jlizen (approved):

PR #678, opened 2026-05-12
fix(services)!: reject trailing slash for file paths

Isvane:
> BREAKING CHANGE: Requests to file paths with a trailing slash (e.g. /index.html/) will now correctly return a 404 FileNotFound instead of serving the file.
> This alters previous behavior where the trailing slash was ignored

## Motivation

This fixes #654.
The goal is to make ServeDir behave more intuitively and consistently with common web server expectations: a trailing slash on a file path should be treated as not found, while trailing slashes on actual directories (including when append_index_html_on_directories is enabled) should continue to work as before.

##  Solution

- The trailing slash check for files is now performed in open_file.rs right after maybe_redirect_or_append_path.
- At that point we know we're dealing with a file (not a directory), so we can cleanly reject paths ending in / (except for the root /).
- Early returns use Option<OpenFileOutput> to keep the control flow clean without extra syscalls.
- Added testcase.


jlizen:
@jplatte this feels like probably the right default, but something that needs to ship across a breaking semver release due to the behavioral change, would you agree? One could call it a bugfix, but that seems a bit sketchy to me.

(specific PR code aside, I will review this more closely)

jplatte:
Yeah, seems likely enough that some users are currently (accidentally) relying on `file/` paths working.

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -62,9 +63,8 @@ pub(super) async fn open_file(
         ServeVariant::Directory {
             append_index_html_on_directories,
         } => {
-            // Might already at this point know a redirect or not found result should be
-            // returned which corresponds to a Some(output). Otherwise the path might be
-            // modified and proceed to the open file/metadata future.
+            let was_directory = is_dir(&path_to_file).await;
```
jlizen: This will result in a redundant syscall since we anyway call `is_dir` inside `maybe_redirect_or_append_path`. 

We could instead just check for the trailing slash AFTER that call, since if it doesn't have an early return, we know that we are dealing with file rather than a directory (perhaps because `index.html` was appended)

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -62,9 +63,8 @@ pub(super) async fn open_file(
         ServeVariant::Directory {
             append_index_html_on_directories,
         } => {
-            // Might already at this point know a redirect or not found result should be
```
jlizen, line 65: I would probably keep this comment, it is largely accurate still? Instead of we moved the slash check until after per [this comment](https://github.com/tower-rs/tower-http/pull/678/changes#r32671069830) we would want a new comment explaining, "at this point we know we are are dealing with a file, so make sure it doesn't have a trailing slash"

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -47,6 +47,7 @@ pub(super) async fn open_file(
     negotiated_encodings: Vec<(Encoding, QValue)>,
     range_header: Option<String>,
     buf_chunk_size: usize,
+    trailing_slash: bool,
```
jlizen: I would probably just check the path directly inline where we need it rather than threading throughout.

tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -893,6 +893,23 @@ async fn calls_fallback_on_null() {
     assert_eq!(res.headers()["from-fallback"], "1");
 }
 
+#[tokio::test]
+async fn not_found_when_file_requested_with_trailing_slash() {
```
jlizen, line 897: Can we also have a check where there is a trailing slash + file where there is a fallback configured?

jlizen (changes requested):
Thanks for the contribution! Implementation could use a bit of tweaking, see comments.

minor: It might be worth applying similar validation to `ServeVariant::SingleFile` as well, no? Probably more confusing to have handling be inconsistent?

Isvane:
> Thanks for the contribution! Implementation could use a bit of tweaking, see comments.
> 
> minor: It might be worth applying similar validation to `ServeVariant::SingleFile` as well, no? Probably more confusing to have handling be inconsistent?

Hi @jlizen,

Thanks so much for the review. I've updated the PR with your feedback. 

I took out the redundant `is_dir` check and moved the trailing slash logic to happen inline right after `maybe_redirect_or_append_path`. I also made sure to leave the original comment alone and added a new one like you suggested.

I also did the suggestion for `SingleFile`. Since this is my first PR ever, I mostly just copied and pasted the existing test structures to write the two new tests (one for the fallback behavior and one for the single file trailing slash check). Hopefully everything looks okay.

I really appreciate you taking the time for the review and the helpful guidance. I learned a lot messing around with this. Let me know if there's anything else I should tweak.

jplatte:
@jlizen are you sure about `SingleFile`? Isn't that (only) used for `ServeFile`, where we (should) just ignore the request path entirely?

jlizen:
> @jlizen are you sure about SingleFile? Isn't that (only) used for ServeFile, where we (should) just ignore the request path entirely?

Yeah, after thinking more with it, I agree that this check doesn't belong with ServeFile.

My knee-jerk previously was, "well, it is intrinsically confusing to serve a file from a directory-like URI". But, it is literally the contract of ServeFile to serve a given file regardless of path, which is already off the rail for "URI intuition", so I think the concern is misplaced.

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -492,7 +492,13 @@ impl ServeVariant {
                 }
                 Some(path_to_file)
             }
-            ServeVariant::SingleFile { mime: _ } => Some(base_path.to_path_buf()),
+            ServeVariant::SingleFile { mime: _ } => {
```
jlizen: Sorry for the churn on this, but let's rip out the `SingleFile` check [as discussed here](https://github.com/tower-rs/tower-http/pull/678#issuecomment-4504783161).

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -75,6 +75,13 @@ pub(super) async fn open_file(
                 return Ok(output);
             }
 
+            // At this point we know we are dealing with a file mapping.
+            // A trailing slash is only valid if it's the root path "/" fetching an index file, otherwise it's invalid.
+            let uri_path = req.uri().path();
+            if uri_path.ends_with('/') && uri_path != "/" {
```
jlizen: This introduces a bug where `/foo/` with `append_index_html_on_directories: true` will now 404 out.

The check should instead be something like:
```
            if uri_path.ends_with('/') && !path_to_file.ends_with("index.html") { ...
```

Another option would be to return an enum from `maybe_redirect_or_append_path`:
```
  enum PathResolution {
      EarlyOutput(OpenFileOutput),
      AppendedIndexHtml,  // was a directory, index.html appended
      NotADirectory,      // not a directory, path unchanged
  }
```

I have no strong preference for one versus the other, dealer's choice. The enum is cleaner in principal but probably a bit overengineered compared to the one-liner.

jlizen: Also, let's definitely have a test case with `/foo/` that appends `index.html` and makes sure it is still served properly.

Let's throw in a test against `/` that appends `index.html` while we're at it.

jlizen (changes requested):

tower-http/src/services/fs/serve_dir/mod.rs
Isvane: It's okay!

tower-http/src/services/fs/serve_dir/open_file.rs
Isvane: I've implemented the fix using the `PathResolution` enum approach.

Thanks for the review and let me know if you'd like any adjustments to the tests or the implementation!

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -282,25 +291,34 @@ async fn maybe_redirect_or_append_path(
...
     append_index_html_on_directories: bool,
-) -> Option<OpenFileOutput> {
-    if !is_dir(path_to_file).await {
-        return None;
+) -> PathResolution {
+    let uri_path = uri.path();
+
+    let is_directory = is_dir(path_to_file).await;
+
+    if uri_path.ends_with('/') && uri_path != "/" && !is_directory {
```
jlizen, line 291: I'm happy to embed the check and early return directly in here, it's probably even cleaner still than having the check outside.

But at that point the enum is not really accomplishing anything we don't do anything different with the appended index html vs is a file variants. We might as well just return the option that might contain an early return like before.

We could instead have the appended variant contain whatever suffix to tack on and not pass `path_to_file` into this function at all, but I don't think it's worth the downstream refactoring.

jlizen: Ie, let's rip out the enum, and then it is perfect!

jlizen (changes requested):

tower-http/src/services/fs/serve_dir/open_file.rs
Isvane: Done. I removed the `PathResolution` enum and went back to returning `Option<OpenFileOutput>`

jlizen (approved):
Awesome!! Thanks for the contribution.

I'm planning a breaking release next that this will go out with. Just waiting to land #422 as well.

Isvane:
> Awesome!! Thanks for the contribution.
> 
> I'm planning a breaking release next that this will go out with. Just waiting to land #422 as well.

Thanks a lot for the thorough review and your patience with all the back-and-forth! 

Happy to see this one finally cleaned up.

PR #679, opened 2026-05-18
Implement `Default` for `limit::ResponseBody`

muhamadazmy:
Implement `Default` for `limit::ResponseBody`

Summary:
This adds a `Default` implementation to `limit::ResponseBody` if the wrapped
body also implements `Default`

This allows combining services as follows:

```
ServiceBudiler::new()
  .layer(CorsLayer::very_permissive())
  .layer(RequestBodyLimitLayer::new(limit))
```


jlizen (approved):
Thanks!

PR #682, opened 2026-05-22
feat(body): add UnsyncBoxBody::new constructor and From<ResponseBody>

jlizen:
Supersedes: #537

## Motivation

Users who return responses with different body types (e.g., from multiple match branches) need to box them into a common type. When one branch uses `ServeDir`/`ServeFile`, the response body is already internally boxed, but wrapping it in `UnsyncBoxBody::new()` causes double-boxing (double allocation + double indirection).

PR #537 by @nanoqsh added `From` impls directly on `http_body_util::combinators::UnsyncBoxBody`, but tower-http has a policy of not leaking `http_body_util` types in its public API.

## Solution

Expose tower-http's own `UnsyncBoxBody` wrapper with:

1. A public `new()` constructor that type-erases any `Body` impl (analogous to `http_body_util::combinators::UnsyncBoxBody::new`)
2. A `From<ServeFileSystemResponseBody>` impl that unwraps the already-boxed inner body without re-boxing

Usage becomes:

```rust
// Before (double-boxing):
.map(|res| res.map(UnsyncBoxBody::new))

// After (zero-cost conversion from ServeDir response):
.map(|res| res.map(UnsyncBoxBody::from))
```

The internal `pub(crate) fn new()` that previously accepted the raw `http_body_util` inner type was renamed to `from_inner()` to avoid ambiguity with the new public constructor.

Tests verify that `UnsyncBoxBody::new` correctly boxes a body and that the `From<ResponseBody>` conversion produces a valid, readable body from a real `ServeDir` response. The conversion is zero-cost by construction (it just moves the inner field), so correctness is primarily a compile-time guarantee.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#628]: https://github.com/tower-rs/tower-http/pull/628
 [#642]: https://github.com/tower-rs/tower-http/pull/642
 
+## Added
+
+- `body`: `UnsyncBoxBody::new()` constructor and `From<ServeFileSystemResponseBody>` conversion to avoid double-boxing when combining `ServeDir` responses with other body types ([#537])
+
 # 0.6.11
 
 ## Added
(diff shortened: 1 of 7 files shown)


jlizen:
@seanmonstar / @davidpdrsn - either of y'all have time to review this?

rcoh (approved):

seanmonstar (approved):

PR #684, opened 2026-05-22
Add configurable Backend trait for ServeDir

### LINKED ISSUE #309: Support configurable backend for ServeDir

jlizen:
Builds on #313 (draft by @davidpdrsn, dormant since Dec 2022). Closes #309.

### Summary

`ServeDir` is hardcoded to `tokio::fs`. This adds a `Backend` trait so it can serve from other sources (rust-embed, include_dir, S3, etc.).

`ServeDir<F, B>` gains a second generic defaulting to `TokioBackend`, so existing code compiles unchanged. A new `ServeDir::with_backend(path, backend)` constructor accepts custom implementations.

The trait surface is `open(PathBuf) -> File` and `metadata(PathBuf) -> Metadata`. ServeDir still owns compression fallback, range handling, trailing slash logic, and index.html appending. Backends just map paths to files.

Divergences from the #313 draft:

- Methods take `PathBuf` rather than `AsRef<Path>`. Since the returned futures must be `'static`, every implementation would clone anyway. Owning the path makes that cost explicit and avoids a hidden allocation.
- GATs are stable now but only used on `File::MetadataFuture<'a>` (where borrowing from self is needed). `Backend` futures are `'static` so GATs would add complexity with no benefit.
- No new feature flag. The trait is small and ships under the existing `fs` feature.

MSRV bumped from 1.64 to 1.65 (GATs require 1.65). This will go out in a breaking release since we are pre-msrv-aware-resolver.

### Testing

- Existing `ServeDir` and `ServeFile` test suites pass unchanged against `TokioBackend`.
- New tests with a mock in-memory backend exercise the full request lifecycle: GET, HEAD, range requests, directory handling, and missing paths.
- Verified pre-compressed file fallback works through a custom backend (ServeDir probes suffixed paths like `.br`/`.gz`/`.zst` via `open()`).

### DIFF
--- a/tower-http/src/services/fs/serve_dir/backend.rs
+++ b/tower-http/src/services/fs/serve_dir/backend.rs
@@ -29,7 +29,7 @@ pub trait Metadata: Send + 'static {
 ///
 /// Must support async reading and seeking (for HTTP range requests).
 /// In-memory backends can use [`std::io::Cursor`] to satisfy the `AsyncSeek` requirement.
-pub trait File: AsyncRead + AsyncSeek + Unpin + Send {
+pub trait File: AsyncRead + AsyncSeek + Unpin + Send + Sync {
     /// The metadata type returned by this file.
     type Metadata: Metadata;
 
@@ -65,8 +65,6 @@ pub trait Backend: Clone + Send + Sync + 'static {
     /// Retrieve metadata for the given path without opening the file.
     fn metadata(&self, path: PathBuf) -> Self::MetadataFuture;
 }
-
-// --- TokioBackend implementation ---
 
 /// Default [`Backend`] implementation using `tokio::fs`.
 #[derive(Clone, Debug, Default)]
--- a/tower-http/src/services/fs/serve_dir/backend.rs
+++ b/tower-http/src/services/fs/serve_dir/backend.rs
@@ -1,0 +1,142 @@
+//! Pluggable backend trait for [`ServeDir`](super::ServeDir).
+//!
+//! The [`Backend`] trait abstracts file system operations so that `ServeDir` can serve
+//! files from sources other than the local filesystem (e.g. rust-embed, include_dir, S3).
+
+use std::{future::Future, io, path::PathBuf, pin::Pin, time::SystemTime};
+use tokio::io::{AsyncRead, AsyncSeek};
+
+/// Trait for file metadata.
+///
+/// This is the information `ServeDir` needs about a file or directory without opening it.
+pub trait Metadata: Send + 'static {
+    /// Returns `true` if this metadata refers to a directory.
+    fn is_dir(&self) -> bool;
+
+    /// Returns the last modification time, if available.
+    fn modified(&self) -> io::Result<SystemTime>;
+
+    /// Returns the size of the file in bytes.
+    fn len(&self) -> u64;
+
+    /// Returns `true` if the file is empty (zero bytes).
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+/// Trait for an opened file.
+///
+/// Must support async reading and seeking (for HTTP range requests).
+/// In-memory backends can use [`std::io::Cursor`] to satisfy the `AsyncSeek` requirement.
+pub trait File: AsyncRead + AsyncSeek + Unpin + Send {
+    /// The metadata type returned by this file.
+    type Metadata: Metadata;
+
+    /// Future returned by [`File::metadata`].
+    type MetadataFuture<'a>: Future<Output = io::Result<Self::Metadata>> + Send
+    where
+        Self: 'a;
+
+    /// Returns metadata for this opened file.
+    fn metadata(&self) -> Self::MetadataFuture<'_>;
+}
+
+/// Trait abstracting filesystem operations for [`ServeDir`](super::ServeDir).
+///
+/// Implement this trait to serve files from non-filesystem sources.
+/// The default implementation ([`TokioBackend`]) wraps `tokio::fs`.
+pub trait Backend: Clone + Send + Sync + 'static {
+    /// The file type returned by [`Backend::open`].
+    type File: File<Metadata = Self::Metadata>;
+
+    /// The metadata type returned by [`Backend::metadata`].
+    type Metadata: Metadata;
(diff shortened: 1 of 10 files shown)


jlizen:
FYI: @Folyd, @SamuNatsu, @kolektiv, @wrapperup, @paul-hansen, @d-k-bo, @CliffHan, @esmevane, @artob, @nyarly,  I see you were interested in the older draft. Feedback on the API is welcome.

tower-http/src/services/fs/serve_dir/backend.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl AsyncRead for TokioFile {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> std::task::Poll<io::Result<()>> {
+        Pin::new(&mut self.0).poll_read(cx, buf)
+    }
+}
```
rcoh, line 103: does this end up being buffered? is this going to result in a ton of tiny spawn_blockings?

esmevane:
@jlizen This looks exactly how I imagined it would, I appreciate the ping! This seems like it'd let me throw out a bunch of workarounds in some of my projects w/ both embeds and testing. Thanks :)

rcoh (approved):
lgtm. my main concern is how this is going to interact w/ the blocking pool, but I suppose this isn't any worse than it was before?

tower-http/src/services/fs/serve_dir/backend.rs
jlizen: This uses `ServeDir` default of `64kb` (or `buf_chunk_size` to override)

jlizen:
> how it interacts with the blocking pool

We buffer at the stream level, default of 64KB but configurable. Same behavior as existing.

I'd like to look at how the new `tokio::fs` usage of `io_uring` would work here instead of the blocking spawn pool... but anyway separate scope (and AIUI that is anyway in userspace code based on their tokio runtime config).

seanmonstar:
Hm, I think something might have happened with the commits, because the current diff doesn't match the description or comments (I only see some additions for etags.)

Zooming out slightly, I have some thoughts regarding exposing traits for users to implement in an unstable crate. If it makes a tiny ecosystem of people depending on different implementations, then "breaking" changes, like going from 0.6 to 0.7, can be disruptive for them. I don't really have a good "so do this instead", I guess. If @jplatte or @davidpdrsn approve of this, that's fine.

jlizen:
Fixed the git history, sorry about that @seanmonstar .

And yeah, that's a good point. The standard alternative would be to make a `tower-http-core` crate that contains the trait definitions, which stays very stable and only is additive. That way `tower-http` can keep moving forward without the same sort of ecosystem churn.

Does that feel warranted in this case?

tower-http/src/services/fs/serve_dir/backend.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+/// Trait for an opened file.
+///
+/// Must support async reading and seeking (for HTTP range requests).
+/// In-memory backends can use [`std::io::Cursor`] to satisfy the `AsyncSeek` requirement.
+pub trait File: AsyncRead + AsyncSeek + Unpin + Send + Sync {
```
jlizen, line 32: This is currently coupling ourselves to `tokio::io::AsyncSeek` in our external interface - we already had `AsyncRead`. We do have `tokio::io` allowlisted already for external types.

I'm open to encapsulating these tokio::io types if we would rather not leak them here, though it's not clear to me how much benefit that provides given that we are deeply coupled to `tokio::io` already (and it is more stable than we are currently).

rcoh:
> Zooming out slightly, I have some thoughts regarding exposing traits for users to implement in an unstable crate. If it makes a tiny ecosystem of people depending on different implementations, then "breaking" changes, like going from 0.6 to 0.7, can be disruptive for them. I don't really have a good "so do this instead", I guess. If @jplatte or @davidpdrsn approve of this, that's fine.

Yeah I see a couple of paths forward;
- if serve-dir doesn't really couple itself to other stuff in tower-http you could make `tower-servedir` as its own crate. Honestly, I don't hate this idea. It is a generally useful concept.
- if there are a key set of traits that we think are "right", you could put them in `tower-http-traits` / `tower-http-core` / `tower-http-api` what have you. Personally, I think this often sounds appealing in their but doesn't often work out that well in practice because _something_ wants to get broken. If the `ServeDir` trait _does_ end up having a little ecosystem around this is always the semver trick of releasing `0.6.<last version>` as re-exporting 0.7 when the trait does not actually change in a breaking way

seanmonstar:
I don't think it's essential enough of a trait to need a -core crate. It'll be fine. Just want to hear from the others about maintaining it into the future.

davidpdrsn:
> If the ServeDir trait does end up having a little ecosystem around this is always the semver trick of releasing 0.6.<last version> as re-exporting 0.7 when the trait does not actually change in a breaking way

I agree with this. Keep it in tower-http for now and if that ends up being a problem then take it from there.

seanmonstar (approved):

PR #691, opened 2026-05-23
Add ETag support and fix conditional request handling in ServeDir

### LINKED ISSUE #324: Hash-based etag layer
I'm thinking about writing a hash-based etag layer.

My plan is to:

- When a response is going to go out, hash the body with blake3 or xxhash
- Add a strong etag header with the digest
- If the request contains an `If-Match` with a matching etag value and the response code is 200, set the code to 304 and remove the body

I probably can't contribute what I wrote to tower-http, since I'm very busy with the project that this is for, but if I wind up writing it I can definitely post or link to the code, in case a tower-http contributor is interested in getting it into tower-http. Also, even if it doesn't get into tower-http, it might give some insight into the viability of the approach.

Some reasons why this might not be a good idea:

- Calculating response body digests adds too much CPU load for not enough benefit
- Replacing the body of responses interacts badly with content encoding, or other layers

Any thoughts on the viability of this approach are welcome, especially if it's been considered for tower-http. I've been looking around for other applications that automatically add etags using a hash function, and I haven't found many. This makes me wonder if the idea is actually wrothwhile.

### LINKED ISSUE #411: ServeDir lacks the `Last-Modified` response header
### Version

0.4.4

### Platform

64-bit (Windows)

### Description

The `ServeDir` lacks the `Last-Modified` response header even when it returns a 304 response.

image

jlizen:
Closes #324, closes #411

## Motivation

`ServeDir` only supports time-based conditional requests today. This causes real problems: if you roll back a deploy, files get an older mtime and clients keep getting 304s for stale content. HTTP dates also have 1-second granularity, so rapid redeploys within the same second are invisible to `If-Modified-Since`.

## Solution

Generate a strong ETag from file metadata (mtime with nanosecond precision + file size) and implement the full RFC 9110 §13.2.2 precondition evaluation order:

1. `If-Match` → 412 (enables safe range request resumption)
2. `If-Unmodified-Since` (when `If-Match` absent) → 412
3. `If-None-Match` → 304
4. `If-Modified-Since` (when `If-None-Match` absent) → 304

The ETag format is `"<hex_secs>.<hex_nanos>-<hex_size>"`. Similar to what nginx does (`<mtime>-<size>`) but with nanosecond precision. No content hashing, no extra I/O; we already call `metadata()` on every request.

304 responses now include `ETag` and `Last-Modified` headers per RFC 9110 §15.4.5 (previously they were bare, which was a spec violation).

No new public API surface. All new types are `pub(super)`. The ETag is always-on, matching the existing `Last-Modified` behavior.

## Testing

Covers: ETag presence on 200, `If-None-Match` producing 304 (exact match, wildcard, weak prefix), `If-Match` producing 412 (non-matching, weak prefix rejected by strong comparison), RFC precedence (ETag conditions override time-based when both present), and validator headers on 304 responses.

### DIFF
--- a/tower-http/src/services/fs/serve_dir/headers.rs
+++ b/tower-http/src/services/fs/serve_dir/headers.rs
@@ -2,6 +2,147 @@ use http::header::HeaderValue;
 use httpdate::HttpDate;
 use std::time::SystemTime;
 
+/// A strong ETag derived from file metadata (size + mtime with nanosecond precision).
+///
+/// Format is an implementation detail and may change between versions. Clients should
+/// treat ETags as opaque values per RFC 9110 §8.8.3.
+#[derive(Clone, Debug)]
+pub(super) struct ETag(HeaderValue);
+
+impl ETag {
+    /// Generate an ETag from file size and modification time.
+    ///
+    /// Returns `None` only for pre-epoch modification times, which are unsupported.
+    pub(super) fn from_metadata(size: u64, modified: SystemTime) -> Option<Self> {
+        let duration = modified.duration_since(SystemTime::UNIX_EPOCH).ok()?;
+        // NOTE: Changing this format is a cache-busting event for all clients,
+        // but is not a semver break (ETags are opaque per RFC 9110 §8.8.3).
+        let value = format!(
+            "\"{:x}.{:08x}-{:x}\"",
+            duration.as_secs(),
+            duration.subsec_nanos(),
+            size
+        );
+        HeaderValue::from_str(&value).ok().map(ETag)
+    }
+
+    pub(super) fn into_header_value(self) -> HeaderValue {
+        self.0
+    }
+
+    /// Strong comparison per RFC 9110 §8.8.3.2: both must not be weak,
+    /// and the opaque-tags must be identical.
+    fn strong_eq(&self, other: &[u8]) -> bool {
+        if other.starts_with(b"W/") {
+            return false;
+        }
+        self.0.as_bytes() == other
+    }
+
+    /// Weak comparison per RFC 9110 §8.8.3.2: ignore W/ prefix,
+    /// compare opaque-tags.
+    fn weak_eq(&self, other: &[u8]) -> bool {
+        let this = self.0.as_bytes();
+        let other = other.strip_prefix(b"W/").unwrap_or(other);
+        let this = this.strip_prefix(b"W/").unwrap_or(this);
+        this == other
+    }
+}
+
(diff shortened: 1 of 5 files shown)


tower-http/src/services/fs/serve_dir/headers.rs
```diff
@@ -2,6 +2,132 @@ use http::header::HeaderValue;
...
+        if bytes == b"*" {
+            return true;
+        }
+        for_each_etag(bytes, |tag| etag.strong_eq(tag))
+    }
+}
+
+/// Iterate over comma-separated ETags in a header value, trimming OWS.
+/// Returns true if `predicate` returns true for any tag (short-circuits).
+fn for_each_etag(header: &[u8], mut predicate: impl FnMut(&[u8]) -> bool) -> bool {
```
rcoh, line 105: is this always correct?  etags can be quoted strings right? (although obviously this implementation does not produce them)

tower-http/src/services/fs/serve_dir/open_file.rs
```diff
@@ -91,11 +106,12 @@ pub(super) async fn open_file(
...
-            last_modified.as_ref(),
-            if_unmodified_since,
-            if_modified_since,
-        ) {
+        let etag = meta
+            .modified()
+            .ok()
+            .and_then(|mtime| ETag::from_metadata(meta.len(), mtime));
+
+        if let Some(output) = preconditions.check(etag.as_ref(), last_modified.as_ref()) {
```
rcoh, line 128: it seems like when we fail to assign an etag, that something weird is going on in most cases (and perhaps the user would want to know about it).

I wonder if its worth a rate-limited warning log or some other mechanism to tell the user why we didn't produce an etag

rcoh (approved):

tower-http/src/services/fs/serve_dir/headers.rs
jlizen: Ehh technically the spec does allow for quoted strings with commas, I can make it resilient to that, it's not too much code.

tower-http/src/services/fs/serve_dir/open_file.rs
jlizen: Probably the best mechanism we have available in the crate is a tracing event, yeah. Eventually I'd like to wire through metrics support in the other middleware behind a feature flag, similar to how we have a tracing feature that adds log lines in.

Anyway for now i'll add a rate-limited warn, good suggestion.

seanmonstar (approved):

PR #693, opened 2026-05-23
Fix: Handle wildcard * in Accept-Encoding per RFC 9110 §12.5.3

### LINKED ISSUE #215: CompressionLayer: `accept-encoding` header parsed incorrectly

jlizen:
Closes: https://github.com/tower-rs/tower-http/issues/215 (the remaining case in it as best I can tell)

## Motivation

The compression middleware silently ignores the `*` wildcard in Accept-Encoding headers. This means requests like `*;q=0,gzip` (which per RFC 9110 §12.5.3 should select gzip and reject everything else) fall back to identity encoding instead.

Similarly, `identity;q=0` with no other encoding listed should signal that the client cannot accept an uncompressed response, but the middleware happily sends one anyway.

## Solution

`Encoding::from_headers` now returns `Option<Encoding>` where `None` signals that no encoding is satisfiable. A new `preferred_encoding_with_wildcard` function handles the negotiation: when a wildcard is present, its q-value applies as the default for any encoding not explicitly listed. When the resulting acceptable set is empty, the middleware returns 406 Not Acceptable.

The `ResponseFuture` handles the `None` case by overwriting the response status to 406 while passing through the inner service's body unchanged. This avoids introducing a new error type or changing the middleware's `Service` output type.

The `fs` module is unaffected since it uses `preferred_encoding()` directly with its own encoding negotiation logic.

Note: this is a behavioral change for clients sending `identity;q=0` without any other encoding. Previously the middleware would silently return an uncompressed response; now it returns 406. This matches the RFC semantics (the client explicitly rejected the only encoding the server can provide without a compression match) but may surprise users who relied on the old fallback.

That said, we are anyway preparing a breaking change, so we can document this behavior change as part of this.

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `DefaultOnEos` now explicitly parent their tracing events to the request span
   rather than relying on the ambient span context. This fixes intermittent cases
   where events could appear without their request span attached ([#655])
+- **breaking:** `compression`: the middleware now handles the `*` wildcard and
+  `identity;q=0` in Accept-Encoding per RFC 9110 §12.5.3. Requests that
+  previously fell back to identity (e.g. `*;q=0` or `identity;q=0` with no
+  other acceptable encoding) now receive a 406 Not Acceptable response. Clients
+  that explicitly reject all encodings without listing an alternative will see
+  different behavior. ([#215])
 - The implicit `tokio` and `async-compression` features are removed (BREAKING).
   These were kept as no-op features in 0.6.x for backwards compatibility after
   the switch to `dep:` syntax in [#642]. Downstream crates that activate
@@ -27,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   by the features that need them (e.g. `compression-gzip`, `fs`, `timeout`).
   ([#628])
 
+[#215]: https://github.com/tower-rs/tower-http/issues/215
 [#628]: https://github.com/tower-rs/tower-http/pull/628
 [#642]: https://github.com/tower-rs/tower-http/pull/642
 
--- a/tower-http/src/compression/future.rs
+++ b/tower-http/src/compression/future.rs
@@ -22,7 +22,7 @@ pin_project! {
     pub struct ResponseFuture<F, P> {
         #[pin]
         pub(crate) inner: F,
-        pub(crate) encoding: Encoding,
+        pub(crate) encoding: Option<Encoding>,
         pub(crate) predicate: P,
         pub(crate) quality: CompressionLevel,
     }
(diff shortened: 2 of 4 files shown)


jlizen:
MSRV check is failing due to `let... else` usage (we are 1.64, requires 1.65) but we are anyway bumping to 1.65 in #684 . I'll just leave this check red and bump it there to avoid conflicts.

rcoh (approved):

seanmonstar (approved):

PR #699, opened 2026-05-24
feat(csrf): add cross-origin protection middleware

yawn:
Ports the CSRF protection scheme introduced in Go 1.25 (described in Filippo Valsorda's blog post) as a new optional `csrf` feature. The middleware combines `Sec-Fetch-Site`, an `Origin` allow-list, and an `Origin`/`Host` fallback to reject cross-origin state-changing requests without per-request token state.

See #656 for context.

### DIFF
--- a/tower-http/src/csrf/future.rs
+++ b/tower-http/src/csrf/future.rs
@@ -56,7 +56,7 @@ where
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.project().kind.project() {
             KindProj::Future { future } => future.poll(cx),
-            KindProj::Rejected { error, _body } => {
+            KindProj::Rejected { error, .. } => {
                 let error = error
                     .take()
                     .expect("ResponseFuture polled after completion");
--- a/tower-http/src/csrf/future.rs
+++ b/tower-http/src/csrf/future.rs
@@ -1,72 +1,71 @@
 use std::future::Future;
-use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-use http::{Response, StatusCode};
 use pin_project_lite::pin_project;
-
-use super::ProtectionError;
 
 pin_project! {
     /// Response future for [`Csrf`].
     ///
     /// [`Csrf`]: super::Csrf
-    pub struct ResponseFuture<F, ResBody> {
+    pub struct ResponseFuture<F>
+    where
+        F: Future,
+    {
         #[pin]
-        kind: Kind<F, ResBody>,
+        kind: Kind<F>,
     }
 }
 
 pin_project! {
     #[project = KindProj]
-    enum Kind<F, ResBody> {
-        Future { #[pin] future: F },
+    enum Kind<F>
+    where
+        F: Future,
+    {
+        Future {
+            #[pin]
+            future: F,
+        },
         Rejected {
-            error: Option<ProtectionError>,
-            _body: PhantomData<ResBody>,
+            response: Option<F::Output>,
         },
     }
 }
 
-impl<F, ResBody> ResponseFuture<F, ResBody> {
+impl<F> ResponseFuture<F>
+where
+    F: Future,
+{
     pub(super) fn future(future: F) -> Self {
         Self {
             kind: Kind::Future { future },
         }
     }
 
-    pub(super) fn rejected(error: ProtectionError) -> Self {
+    pub(super) fn rejected(response: F::Output) -> Self {
         Self {
             kind: Kind::Rejected {
-                error: Some(error),
-                _body: PhantomData,
+                response: Some(response),
             },
         }
     }
 }
 
-impl<F, E, ResBody> Future for ResponseFuture<F, ResBody>
+impl<F> Future for ResponseFuture<F>
 where
-    F: Future<Output = Result<Response<ResBody>, E>>,
-    ResBody: Default,
+    F: Future,
 {
-    type Output = Result<Response<ResBody>, E>;
+    type Output = F::Output;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.project().kind.project() {
             KindProj::Future { future } => future.poll(cx),
-            KindProj::Rejected { error, .. } => {
-                let error = error
+            KindProj::Rejected { response } => Poll::Ready(
+                response
                     .take()
-                    .expect("ResponseFuture polled after completion");
-                let mut response = Response::new(ResBody::default());
-
-                *response.status_mut() = StatusCode::FORBIDDEN;
-                response.extensions_mut().insert(error);
-
-                Poll::Ready(Ok(response))
-            }
+                    .expect("ResponseFuture polled after completion"),
+            ),
         }
     }
 }
--- a/tower-http/src/csrf/future.rs
+++ b/tower-http/src/csrf/future.rs
@@ -1,0 +1,72 @@
+use std::future::Future;
+use std::marker::PhantomData;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use http::{Response, StatusCode};
+use pin_project_lite::pin_project;
+
+use super::ProtectionError;
+
+pin_project! {
+    /// Response future for [`Csrf`].
+    ///
+    /// [`Csrf`]: super::Csrf
+    pub struct ResponseFuture<F, ResBody> {
+        #[pin]
+        kind: Kind<F, ResBody>,
+    }
+}
+
+pin_project! {
+    #[project = KindProj]
+    enum Kind<F, ResBody> {
+        Future { #[pin] future: F },
+        Rejected {
+            error: Option<ProtectionError>,
+            _body: PhantomData<ResBody>,
+        },
+    }
+}
+
+impl<F, ResBody> ResponseFuture<F, ResBody> {
+    pub(super) fn future(future: F) -> Self {
+        Self {
+            kind: Kind::Future { future },
+        }
+    }
+
+    pub(super) fn rejected(error: ProtectionError) -> Self {
+        Self {
+            kind: Kind::Rejected {
+                error: Some(error),
+                _body: PhantomData,
+            },
+        }
+    }
+}
+
+impl<F, E, ResBody> Future for ResponseFuture<F, ResBody>
+where
+    F: Future<Output = Result<Response<ResBody>, E>>,
+    ResBody: Default,
+{
+    type Output = Result<Response<ResBody>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project().kind.project() {
+            KindProj::Future { future } => future.poll(cx),
+            KindProj::Rejected { error, _body } => {
+                let error = error
+                    .take()
+                    .expect("ResponseFuture polled after completion");
+                let mut response = Response::new(ResBody::default());
+
+                *response.status_mut() = StatusCode::FORBIDDEN;
+                response.extensions_mut().insert(error);
+
+                Poll::Ready(Ok(response))
+            }
+        }
+    }
+}
--- a/tower-http/src/csrf/layer.rs
+++ b/tower-http/src/csrf/layer.rs
@@ -5,7 +5,7 @@ use http::{Method, Uri};
 use tower_layer::Layer;
 
 use super::service::Csrf;
-use super::url::TrustedOrigin;
+use super::url::UriExt;
 use super::{BypassFn, ConfigError, DebugFn, DefaultResponseForProtectionError, Origins};
 
 /// Layer that applies the [`Csrf`] middleware.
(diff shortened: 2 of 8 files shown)


tower-http/src/csrf/mod.rs
```diff
@@ -0,0 +1,729 @@
...
+//!
+//! Rejected requests receive a `403 Forbidden` response. The originating
+//! [`ProtectionError`] is attached to the response's extensions so handlers can
+//! distinguish between explicit cross-origin rejections and conservative
+//! fallback rejections (e.g. requests from old browsers without
+//! `Sec-Fetch-Site`).
+//!
+//! # Deployment caveat
+//!
+//! The middleware trusts whatever `Origin` and `Host` reach it. Reverse proxies
```
jlizen, line 27: Thanks for this note

tower-http/src/csrf/mod.rs
```diff
@@ -0,0 +1,729 @@
+//! Modern protection against [cross-site request forgery] (CSRF) attacks.
+//!
+//! This middleware implements the CSRF protection scheme [introduced in Go 1.25][go]
```
jlizen, line 3: It looks like we diverge from the Go impl in a couple ways:
- trusted origin check canonicalizes both sides of the assertion (go is strict byte match)
- origin/host check also canonicalizes (and lowercases)

This seems like a reasonable choice to me, the Go approach seems like it'd be a bit more brittle than needed, but is it intended?  We should at least call out the divergence even if we do want it. Are there any other reference implementations for this type of CSRF that do similar?

tower-http/src/csrf/mod.rs
```diff
@@ -0,0 +1,729 @@
...
+    fn contains(&self, origin: &str) -> bool {
+        self.0.contains(origin)
+    }
+
+    fn insert(&mut self, origin: impl Into<String>) {
+        Arc::make_mut(&mut self.0).insert(origin.into());
+    }
+}
+
+#[cfg(test)]
```
jlizen, line 255: We should have some tests that exercise the full `Service:call()` flow, generally via `ServiceExt::oneshot()`.

Specifically I'd like to see:
- allowed path
- the error path checking that we get a 403 + `ProtectionError` in extensions

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+            None | Some(b"") => {} // fall through to Origin check
+            Some(_) if self.is_exempt(req.method(), req.uri(), origin_uri.as_ref()) => {
+                return Ok(())
+            }
+            Some(_) => return Err(ProtectionError::CrossOriginRequest),
+        }
+
+        if matches!(origin, None | Some(b"")) {
+            #[cfg(feature = "tracing")]
+            tracing::trace!(uri = %req.uri(), "request passed: neither sec-fetch-site nor origin header (same-origin or not a browser request)");
```
jlizen:
```suggestion
            tracing::trace!(uri = %req.uri().path(), "request passed: neither sec-fetch-site nor origin header (same-origin or not a browser request)");
```

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+            if let (Some(origin_host), Some(host_name)) = (uri.host(), host_name) {
+                let port_origin = uri.effective_port();
+                // Host carries no scheme of its own; assume Origin's scheme so an implicit
+                // Host port resolves to Origin's scheme default (443/80) rather than
+                // silently inheriting Origin's explicit port.
+                let port_host = port_host.or_else(|| uri.scheme_default_port());
+
+                if origin_host.eq_ignore_ascii_case(host_name) && port_origin == port_host {
+                    #[cfg(feature = "tracing")]
+                    tracing::trace!(uri = %req.uri(), "request passed: origin is same as host");
```
jlizen:
```suggestion
                    tracing::trace!(uri = %req.uri().path(), "request passed: origin is same as host");
```

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    fn is_exempt(&self, method: &Method, uri: &Uri, origin: Option<&Uri>) -> bool {
+        if self
+            .insecure_bypass
+            .as_ref()
+            .map_or(false, |p| p(method, uri))
+        {
+            #[cfg(feature = "tracing")]
+            tracing::trace!(uri = %uri, "request passed: bypassed");
```
jlizen:
```suggestion
            tracing::trace!(uri = %uri.path(), "request passed: bypassed");
```

Let's strip query params to avoid issues with sensitive data.

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+        false
+    }
+
+    pub(super) fn verify<Body>(&self, req: &Request<Body>) -> Result<(), ProtectionError> {
+        if matches!(
+            req.method(),
+            &Method::GET | &Method::HEAD | &Method::OPTIONS
+        ) {
+            #[cfg(feature = "tracing")]
+            tracing::trace!(uri = %req.uri(), "request passed: safe method");
```
jlizen:
```suggestion
            tracing::trace!(uri = %req.uri().path(), "request passed: safe method");
```

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+            .and_then(|s| s.parse::<Uri>().ok())
+            .filter(|u| matches!(u.scheme_str(), Some("http" | "https")));
+
+        let sec_fetch_site = req.headers().get("sec-fetch-site").map(|h| h.as_bytes());
+
+        // Fetch spec mandates lowercase here; exact byte match is intentional.
+        match sec_fetch_site {
+            Some(b"same-origin" | b"none") => {
+                #[cfg(feature = "tracing")]
+                tracing::trace!(uri = %req.uri(), "request passed: sec-fetch-site is same-origin or none");
```
jlizen:
```suggestion
                tracing::trace!(uri = %req.uri().path(), "request passed: sec-fetch-site is same-origin or none");
```

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+        trusted_origins: Origins,
+    ) -> Self {
+        Self {
+            inner,
+            insecure_bypass,
+            trusted_origins,
+        }
+    }
+
+    fn is_exempt(&self, method: &Method, uri: &Uri, origin: Option<&Uri>) -> bool {
```
jlizen: can we make this take a struct as input, or otherwise make the two uris more legible at call site

tower-http/src/csrf/future.rs
```diff
@@ -0,0 +1,72 @@
...
+where
+    F: Future<Output = Result<Response<ResBody>, E>>,
+    ResBody: Default,
+{
+    type Output = Result<Response<ResBody>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project().kind.project() {
+            KindProj::Future { future } => future.poll(cx),
+            KindProj::Rejected { error, _body } => {
```
jlizen:
```suggestion
            KindProj::Rejected { error, .. } => {
```

tower-http/src/csrf/future.rs
```diff
@@ -0,0 +1,72 @@
...
+                let error = error
+                    .take()
+                    .expect("ResponseFuture polled after completion");
+                let mut response = Response::new(ResBody::default());
```
jlizen: This diverges from the go in that it forces a default response on rejection... meanwhile the Go way lets you customize. It wouldn't be particularly difficult to support similar. 

We can reuse`catch_panic`'s`ResponseForPanic` API basically as-is. I think it's worth scoping in, though if you don't have the time, it can wait. If you did want to punt, let's document usage of `MapResponse` to customize instead.

tower-http/src/csrf/future.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+impl<F, E, ResBody> Future for ResponseFuture<F, ResBody>
+where
+    F: Future<Output = Result<Response<ResBody>, E>>,
+    ResBody: Default,
+{
+    type Output = Result<Response<ResBody>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
jlizen, line 61: I'd rather move the failure response body construction out into the service `call()` rather than constructing it inside the poll.

That has the downside of a tiny bit of extra work if the response future is dropped, but simplifies things a decent amount. Notably, we can drop the `ResBody` generic entirely from the response future.

That produces roughly:
```
  pub struct ResponseFuture<F> {
      kind: Kind<F>,
  }

  enum Kind<F: Future> {
      Future { #[pin] future: F },
      Rejected { response: Option<F::Output> },
  }

  impl<F, E, ResBody> Future for ResponseFuture<F>
  where
      F: Future<Output = Result<Response<ResBody>, E>>,
  {
      type Output = F::Output;
      fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
          match self.project().kind.project() {
              KindProj::Future { future } => future.poll(cx),
              KindProj::Rejected { response } => Poll::Ready(
                  response.take().expect("polled after completion")
              ),
          }
      }
  }
```

And then the service handles the error:
```
 Err(err) => {
      let mut response = Response::new(ResBody::default());
      *response.status_mut() = StatusCode::FORBIDDEN;
      response.extensions_mut().insert(err);
      ResponseFuture::rejected(Ok(response))
  }
```

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,179 @@
...
+
+use super::future::ResponseFuture;
+use super::url::UriExt;
+use super::{BypassFn, DebugFn, Origins, ProtectionError};
+
+/// Middleware that enforces cross-origin request forgery (CSRF) protection.
+///
+/// See the [module docs](crate::csrf) for an example.
+#[derive(Clone)]
+pub struct Csrf<S> {
```
jlizen: add `#[must_use]` here too

tower-http/src/csrf/mod.rs
```diff
@@ -0,0 +1,729 @@
...
+}
+
+impl std::error::Error for ConfigError {}
+
+/// Reason a request was rejected by [`Csrf`].
+///
+/// Attached to the `403 Forbidden` response's extensions so handlers can
+/// distinguish between explicit cross-origin rejections and conservative
+/// fallback rejections.
+#[derive(Clone, Debug, PartialEq)]
```
jlizen: It's tempting to derive `Copy` here, but I guess probably better to leave open the door to more complex variants. Same for eq.

jlizen (changes requested):
Thanks for this, @yawn , looks great overall. 

Had one small refactor to keep the future a bit simpler.

And then, up to you if you care to add rejection customization, or just document how to use the `MapResponse` layer for it.

tower-http/src/csrf/mod.rs
yawn: TBH I think the safer choice would be to do exactly the same as the original Go version. I am definitly not more qualified to justify deviations then the original author and reviewers.

My gut feeling was the same as yours: it _seems_ reasonable. But it might not be worth the tradeoff of the port being harder to compare.

tower-http/src/csrf/mod.rs
yawn: Your call. My gut feeling is that the error surface is unlikely to change a lot.

tower-http/src/csrf/mod.rs
yawn: I am porting the original [axum example](https://github.com/yawn/tower-csrf/blob/main/tests/axum.rs).

tower-http/src/csrf/service.rs
yawn: Inlined it as a closure.

tower-http/src/csrf/future.rs
yawn: Done. One minor point: I have added a `trace!` when encountering a protection error at the service level. Seems logical but putting it into the default response would make overwriting more simple. WDYT?

yawn:
One last time from by side: `BypassFn` is currently private. I don't think it's too much trouble for users to use the middleware as is but wanted to highlight nevertheless.

tower-http/src/csrf/mod.rs
jlizen: I think let's match exactly, but open a tracking issue describing the area we are concerned is brittle, specifying the alternative, and request feedback from users about whether the strict version is causing them issues?

tower-http/src/csrf/mod.rs
jlizen: Ehh unlikely to change but also not all that helpful to have, let's stay defensive with semver until we have a good reason. Honestly I'd prefer to have an opaque struct inside each variant rather than unit variants to leave open the door for schema evolution, though I wouldn't block over it.

[This skill](https://github.com/battery-pack-rs/battery-pack/blob/main/battery-packs/error-battery-pack/skills/library-errors/SKILL.md) roughly captures my preferences for library errors, though open to alternative approaches (or just keeping it simple, it's fine).

jlizen:
> One last time from by side: BypassFn is currently private. I don't think it's too much trouble for users to use the middleware as is but wanted to highlight nevertheless.

Fine by me, let's keep it opaque for now until we have a reason to open it up.

tower-http/src/csrf/future.rs
jlizen: Yeah, I'm fine with this approach, given that is only a trace log level i'd probably prefer wiping it out if the builder method is used.

tower-http/src/csrf/layer.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Replaces the response builder used when a request is rejected.
+    ///
+    /// Accepts any type that implements [`ResponseForProtectionError`](super::ResponseForProtectionError),
+    /// including a `FnMut(ProtectionError) -> Response<B> + Clone` closure.
+    /// The default builder returns a `403 Forbidden` with an empty body and
+    /// the [`ProtectionError`](super::ProtectionError) attached to the
+    /// response's extensions.
+    pub fn with_rejection_response<R>(self, rejection_response: R) -> CsrfLayer<R> {
```
jlizen: nit: probably worth adding the `Clone` bound to cause it to fail here instead of layer at the `layer()` call, will be clearer what's wrong

tower-http/src/csrf/mod.rs
```diff
@@ -0,0 +1,850 @@
...
+    fn contains(&self, origin: &str) -> bool {
+        self.0.contains(origin)
+    }
+
+    fn insert(&mut self, origin: impl Into<String>) {
+        Arc::make_mut(&mut self.0).insert(origin.into());
+    }
+}
+
+#[cfg(test)]
```
jlizen, line 255: New tests are good, thanks. Probably I'd add one more capturing the "happy path" with custom rejection closure (where it doesn't fire), for completeness.

jlizen (commented):
New changes are great, thanks, the custom rejection closure impl looks good.

The remaining bit is undoing the canonicalize for the comparisons, like we discussed, and optionally tweaking the error type to be more defensive semver.

tower-http/src/csrf/service.rs
```diff
@@ -0,0 +1,199 @@
...
+            inner,
+            insecure_bypass,
+            rejection_response,
+            trusted_origins,
+        }
+    }
+
+    pub(super) fn verify<Body>(&self, req: &Request<Body>) -> Result<(), ProtectionError> {
+        if matches!(
+            req.method(),
```
Tremoneck, line 46:
```suggestion
            req.method().is_safe()
```
We could just use the function of Method to see if it is safe. I think it would be clearer. However it would also allow ```TRACE```.

Tremoneck (commented):
Just something I noticed when scrolling through. It would however not follow the Blog Post or the Go implementation.

yawn:
I'll need a few days for the next round

yawn:
All remarks have been addressed and I've added two tracking issues (as proposed): #701 and #702 .

One  one more request for clarification: I am adding errors [now in the service](https://github.com/yawn/tower-http/blob/aac7ca7b92ace24846b85bc8a2afa35961089523/tower-http/src/csrf/service.rs#L189), not anymore in the default response. This is a bit more logical but this has not really a precedence (as far as I can tell) in the crate. LMK if that works for you.

Addendum: following your style remarks `kind` is now opaque so it's a bit questionable if making the error available is even worth it. I also did _not_ update the style of the config error - not 100% sure if I should have done so or if the remarks on your end were deliberately scoped to just the protection error.

jlizen (approved):
Thanks for cutting #701 / #702 !

> One one more request for clarification: I am adding errors [now in the service](https://github.com/yawn/tower-http/blob/aac7ca7b92ace24846b85bc8a2afa35961089523/tower-http/src/csrf/service.rs#L189), not anymore in the default response. This is a bit more logical but this has not really a precedence (as far as I can tell) in the crate. LMK if that works for you.

I agree with this choice.


> Addendum: following your style remarks kind is now opaque so it's a bit questionable if making the error available is even worth it. I also did not update the style of the config error - not 100% sure if I should have done so or if the remarks on your end were deliberately scoped to just the protection error.

I think it's useful, yeah, it still exposes `kind()` which is nice for metrics or other more granular handling.

jlizen:
Great stuff! Thanks for the contribution @yawn (and the follow up issues). Will get this out next release, sometime next week.

PR #703, opened 2026-06-03
fix: Make SetMultiple*Header Clone for !Clone http bodies

Jesse-Bakker:
Replace some derived `Clone` implementations with manual ones.

## Motivation

For a couple structs, `Clone` was previously derived, creating a `Clone` bound
on the http request and therefore the request body. This meant that it wasn't
possible to use `SetMultipleRequestHeader` and `SetMultipleResponseHeader` for, for example, tonic.

## Solution

Manually implement `Clone` for the relevant types, to avoid the 
`impl<M: Clone> Clone for T<M>` bounds that caused the resulting service to not be Clone.

### DIFF
--- a/tower-http/src/set_header/request/multiple_headers.rs
+++ b/tower-http/src/set_header/request/multiple_headers.rs
@@ -15,9 +15,16 @@ use crate::set_header::{HeaderInsertionConfig, HeaderMetadata, InsertHeaderMode}
 /// Layer that applies [`SetMultipleRequestHeader`] which adds multiple request headers.
 ///
 /// See [`SetMultipleRequestHeader`] for more details.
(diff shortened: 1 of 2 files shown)


tower-http/src/set_header/request/multiple_headers.rs
```diff
@@ -361,4 +379,18 @@ mod tests {
         let debug_str = format!("{:?}", layer);
         assert!(debug_str.contains("SetMultipleRequestHeadersLayer"));
     }
+
+    #[test]
```
jlizen, line 383: Thanks for adding this test!

jlizen (approved):
Thanks for the contribution!

Issue #525, opened 2024-10-02
Support for larger sizes in `compression::predicate::SizeAbove`


liss-h:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Currently, `SizeAbove` has a `u16` inside of it to limit the size where compression is active.
It would be handy if a larger type (like `u64`) was used instead.

https://github.com/tower-rs/tower-http/blob/4bca529738aab78660d4d64a1da50dd2eb3c34b0/tower-http/src/compression/predicate.rs#L147-L149

### Motivation
Since the size is in bytes `2^16` bytes is still quite small (64KiB), when the internet connection is fast enough it does not really make sense to compress payloads that small. So if you don't want to compress payloads that small you currently just have to reimplement the same thing, which is not ideal.

### Proposal
Increase the type size from `u16` to `u64`. There are no obvious drawbacks apparent for me.
If you would like to make this change I can implement it if you want.

jlizen:
Closed by #704

PR #704, opened 2026-06-05
feat(compression)!: upgrade SizeAbove threshold from u16 to u64

Isvane:
## Motivation

I was looking through some older issues and noticed this issue #525 that have been stale for awhile now.

Right now, `SizeAbove` uses a `u16` internally to hold the byte threshold for response compression. Because of that, the absolute maximum limit anyone can configure is about 64 KiB. 

While that works great for tiny payloads, 64 KiB is pretty small for modern high-speed connections. If a service is heavily CPU-bound and wants to skip compressing payloads under 1 MB or 2 MB to save processing power, they're kind of stuck. They have to completely copy-paste and reimplement `SizeAbove` just to change the integer type, which feels a bit redundant since the rest of the predicate ecosystem here is so nice to use.

## Solution

* Bumped `SizeAbove(u16)` to `SizeAbove(u64)`.
* Updated `DEFAULT_MIN_SIZE` and the `new` function to expect a `u64`.
* Cleaned up the `Predicate` implementation for `SizeAbove` by removing the `self.0 as u64` cast, since the types natively match up now.

### Question

I wanted to ask from a maintainer's perspective. is there a drawback to doing this that I'm completely missing? I was trying to figure out if `u16` was originally picked for a specific reason. Since these predicates are usually created once and live for the lifetime of the server, I figured the extra bytes wouldn't hurt anything, but I'd love to know if there's any historical context or edge cases I should learn about!

### DIFF
--- a/tower-http/src/compression/predicate.rs
+++ b/tower-http/src/compression/predicate.rs
@@ -146,17 +146,17 @@ impl Predicate for DefaultPredicate {
 
 /// [`Predicate`] that will only allow compression of responses above a certain size.
 #[derive(Clone, Copy, Debug)]
-pub struct SizeAbove(u16);
+pub struct SizeAbove(u64);
 
 impl SizeAbove {
-    pub(crate) const DEFAULT_MIN_SIZE: u16 = 32;
+    pub(crate) const DEFAULT_MIN_SIZE: u64 = 32;
 
     /// Create a new `SizeAbove` predicate that will only compress responses larger than
     /// `min_size_bytes`.
     ///
     /// The response will be compressed if the exact size cannot be determined through either the
     /// `content-length` header or [`Body::size_hint`].
-    pub const fn new(min_size_bytes: u16) -> Self {
+    pub const fn new(min_size_bytes: u64) -> Self {
         Self(min_size_bytes)
     }
 }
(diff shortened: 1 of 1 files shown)


seanmonstar:
Probably a good idea. I think the only thing is that changing the argument type in the constructor is a breaking change.

Isvane:
> Probably a good idea. I think the only thing is that changing the argument type in the constructor is a breaking change.

Thank you so much for the review! Yes, I deliberately marked it as a breaking change in the title because of that. But I heard tower-http might be releasing a new major version soon? Or maybe make a new constructor or something if we want to make it backward compatible?

jlizen:
That's right, next release will be breaking.

I don't think a new constructor is warranted given that this is unlikely to force code changes in consumers, even though it is technically breaking. Any opinion @seanmonstar ?

jlizen (approved):
Seems sensible to me, and I don't see a need for a new constructor, but interested in your opinion @seanmonstar

seanmonstar:
No, I don't see a need for two constructors. (If someone wants them, `From<T>` is possible, but with many of those, I prefer to wait until someone needs them, instead of thinking some might eventually need it.)

Issue #251, opened 2022-04-25
Add regression test for directory traversal attack fixed in #204


davidpdrsn:
Back in https://github.com/tower-rs/tower-http/pull/204 we fixed a directory traversal attack on Windows. Everything is safe now so no need to worry but it seems we forgot to add a test for it. We should do that!

junghwan16:
Hi @davidpdrsn, I opened a small PR for this: #705.

It adds a focused regression test for the Windows drive-prefix path case fixed in #204. 
When you have time, could you take a look?

jlizen:
Thanks! Adding this to my review queue (unless David gets to it first)

PR #705, opened 2026-06-07
Add Windows drive-prefix path regression test

junghwan16:
## Motivation

#251 notes that #204 fixed a Windows directory traversal issue without adding a regression test.

## Solution

Add a regression test for Windows drive-prefix path components in `ServeDir` path validation. (#251)

### DIFF
--- a/tower-http/src/services/fs/serve_dir/tests.rs
+++ b/tower-http/src/services/fs/serve_dir/tests.rs
@@ -1117,31 +1117,42 @@ fn test_build_and_validate_path_reserved_dos_names() {
     }
 }
 
-#[test]
-fn test_build_and_validate_path_windows_drive_prefixes() {
-    use super::ServeVariant;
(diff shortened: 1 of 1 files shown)


jlizen (changes requested):
This test looks great. Can we also add a case for percent-encoded? Eg `/anypath/c%3A/windows/win.ini`?

jlizen:
Also I'd have a slight preference for a full `ServeDir::oneshot` test in `tower-http/src/services/fs/serve_dir/tests.rs`... testing the helper *should* be fine, but it's always nice to not assume it is properly invoked. I'd approve either way though if we can add a probe for the percent-encoded path.

junghwan16:
@jlizen Yeah, I agree that would be better. I'll update it.

jlizen (approved):
Thanks for the contribution!

PR #707, opened 2026-06-12
chore(release): prepare 0.7.0

jlizen:
Bump `tower-http` to 0.7.0

Breaking changes in this release:

- preserve request extensions across redirects
- add configurable Backend trait for ServeDir, bump MSRV 1.65 (#684)
- upgrade SizeAbove threshold from u16 to u64 (#704)
- handle wildcard * in Accept-Encoding per RFC 9110 §12.5.3 (#693)
- remove no-op tokio and async-compression features (#681)
- include grpc error message in tracing (#422)
- reject trailing slash for file paths (#678)

### DIFF
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 
-- `csrf`: add cross-origin protection middleware ([#699])
+- `csrf`: add cross-site request forgery (CSRF) protection middleware, porting the cross-origin protection scheme introduced in Go 1.25 ([#699])
 
   ```rust
   use tower::ServiceBuilder;
--- a/tower-http/CHANGELOG.md
+++ b/tower-http/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 # 0.7.0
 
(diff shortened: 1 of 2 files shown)


tower-http/CHANGELOG.md
```diff
@@ -5,57 +5,134 @@ All notable changes to this project will be documented in this file.
...
+  // Rejects cross-origin state-changing requests using `Sec-Fetch-Site`,
+  // an `Origin` allow-list, and an `Origin`/`Host` fallback. No per-request
+  // token state required.
+  let layer = CsrfLayer::new().add_trusted_origin("https://example.com")?;
+
+  let service = ServiceBuilder::new().layer(layer).service_fn(handler);
+  ```
+- `timeout`: add `DeadlineBody` for non-resetting body timeouts, applied via the new `RequestBodyDeadlineLayer` and `ResponseBodyDeadlineLayer` ([#688])
+
+  Unlike `TimeoutBody`, which resets its deadline on every frame, `DeadlineBody` caps the total wall-clock time of a body transfer. A slow client trickling one byte at a time never trips an idle timeout but will trip a deadline.
```
jplatte:
```suggestion
  Unlike `TimeoutBody`, which resets its deadline on every frame, `DeadlineBody` caps the total time of a body transfer. A slow client trickling one byte at a time never trips an idle timeout but will trip a deadline.
```

I think "wall-clock" is more likely confusing than helpful, it made me think of `std::time::SystemTime`.

jplatte (approved):

seanmonstar (approved):

PR #709, opened 2026-07-01
docs(examples): add custom either middleware example

Isvane:
## Description

Hey! Opening this PR to add a new example showing how to handle conditional, short-circuiting responses in custom `tower` middleware.

I’ve set it up under `examples/custom-middleware-either` with its own `Cargo.toml`, `README.md`, and `src/main.rs` to match the layout of the existing examples.

**Note**: The code is working but a bit rough around the edges, and my doc comments and the `README.md` are definitely lacking right now. I couldn't quite find the right words to make it super clear. I'd love some feedback on both the wording and the code during the review.

## Motivation

This comes from a recent Discord discussion from Tony, Darius and @jlizen. Because `Service` forces you to have one concrete `Response` associated type, you run into a wall if your middleware wants to either bounce a request early (with an error body) or let it pass through to the inner service (with whatever body that service returns)

## Solution

I added the custom-middleware-either example. Which contains:
- Using `Either<B, Full<Bytes>>` to gracefully unify the happy-path body and the early-rejection error body.
- Short-circuiting the future using `std::future::ready` for the error path.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -2,5 +2,7 @@
 
 This folder contains various examples of how to use tower-http.
 
+- `axum-key-value-store`: A key/value store with an HTTP API built with axum.
 - `warp-key-value-store`: A key/value store with an HTTP API built with warp.
 - `tonic-key-value-store`: A key/value store with a gRPC API and client built with tonic.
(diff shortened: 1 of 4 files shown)


Isvane:
closing in favor of #711

PR #710, opened 2026-07-05
feat(trace): add abstract `Clock` trait supporting custom timing impl

### LINKED ISSUE #657: Feature Proposal: Shall we replace some API calls from `std::time` with `web_time` to make `TraceLayer` (and more services) WASM compatible?

### Motivation

I'm developing a qqbot webhook callback handler targeting WASM (to be ran on Supabase Edge Functions' Runtime), and would like to use [`tower_http::trace::TraceLayer`](https://docs.rs/tower-http/0.6.8/tower_http/trace/struct.TraceLayer.html) to enable tracing for my application. However, as some time related APIs from standard library (like [`std::time::Instant::now()`](https://doc.rust-lang.org/std/time/struct.Instant.html#method.now)) panics on `wasm-unkown-unkown`, `TraceLayer` and some other service couldn't be used in WASM context.

By replacing those time related API calls with WASM compatible ones, we can make this crate more accessible to those who wish to use them in WASM context (for example, [`worker`](https://lib.rs/crates/worker) with [`axum`](https://lib.rs/crates/axum) for Cloudflare Workers and edge computing).

### Proposal

We could replace `std::time` with [`web-time`](https://lib.rs/crates/web-time), and gate this behind a cargo feature (for compatibility and those who don't want wasm stuff). I've tried vibe coding with Copilot and it seems to be working, so this should be a viable solution.

Possible drawback is the user will have to download an extra dependency if they enable `full`.

If possible, I'm willing to do the implementation myself (I would do it by hand and won't vibe code for the PR of course).

### Alternatives

An alternative would be not use [`tower_http::trace::TraceLayer`](https://docs.rs/tower-http/0.6.8/tower_http/trace/struct.TraceLayer.html) and instead manually implement the tracing stuff. The drawback is the user would have to write a lot more code to get it working.

I chose the proposed solution because it's relatively simple to implement ([`web-time`](https://lib.rs/crates/web-time) should be a drop-in replacement for those APIs) and could make more `tower-http` services available for WASM (edge computing scenario).

Saplyn:
## Motivation

Adding a way to allow user to supply their own `Clock` (time-recording) implementation if the default one (from `std`) isn't desired. Enabling deterministic testing with fake clocks and WASM-compatibility if the user has a WASM-compatible clock implementation.

Closes #657

## Solution

Added an abstract `Clock` trait for timing, replacing the original implementation.

I decided to go with static dispatch (despite @jlizen [suggest otherwise](https://github.com/tower-rs/tower-http/issues/657#issuecomment-4520828209)) because adding a `dyn Clock` would make trace layer and service non-copy, resulting a potential breaking API change.

## TODOs

- [x] test cases covering the changes
- [ ] whether [`Clock::now()` and `Clock::elapsed()`](https://github.com/Saplyn/tower-http/blob/1632fb2e8927f0fe5b4810cfed663c89c031f8e0/tower-http/src/trace/clock.rs#L39-L51) should take `&self` or `&mut self`


tower-http/src/trace/clock.rs
```diff
@@ -0,0 +1,77 @@
+use std::time::{Duration, Instant};
```
jlizen, line 1: I don't think that the clock belongs in `trace` but rather in a root module (as evidenced by `on_early_drop` importing it). You could re-export it from `trace` if desired.

tower-http/src/trace/clock.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Extension trait for duration types that need to be formatted as latencies.
+///
+/// The [`Clock`] trait's `Duration` associated type must implement this trait,
+/// which provides conversion methods used by [`Latency`](super::Latency).
+pub trait DurationExt {
```
jlizen, line 7: This `Duration` trait creates a lot of ergonomic issues... it forces all of the downstream callbacks to specify generics, and I don't think it is particularly needed.

As best as I can tell, the wasm support issues are only about `std::time::Instant/SystemTime`. `Duration` itself seems fine (web-time` just re-exports the `core::time:Duration`. Similarly other common clocks like `tokio::time::Instant` and `quanta` use it).

Is there a reason we can't simplify to:
```
fn elapsed(&self, instant: Self::Instant) -> std::time::Duration
```

tower-http/src/trace/clock.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+/// An abstract clock for measuring time.
+///
+/// Used by the trace middleware to record latencies and stream durations.
+/// The default implementation ([`DefaultClock`]) uses [`std::time::Instant`]
+/// and [`std::time::Duration`], but custom clocks can be provided — for
+/// example, to mock time in tests or to use a WASM-compatible time source.
+pub trait Clock: Copy {
```
jlizen, line 39: Why do we need `Copy` here? That makes it dyn-incompatible and also restricts all implementors to be `Copy` (which we can see the irritation in our own `DummyClock` mock, and any other stateful clock).

And then anyway `TraceLayer` and `Trace` derive `Copy`, so if the `Clock` impl is `Copy`, then those structs also remain `Copy`.

jlizen: And then, I would probably require `Clone` on the two use sites that need it, and use `&self` for access to make life better for stateful clocks.

tower-http/src/trace/clock.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// An instant in time.
+    type Instant: Copy + Send;
```
jlizen, line 41: I would probably keep `Send` on the use sites rather than the trait

tower-http/src/trace/clock.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn elapsed(&self, instant: Self::Instant) -> Self::Duration;
+}
+
+/// The default [`Clock`] implementation, backed by [`std::time::Instant`].
+#[derive(Debug, Clone, Copy, Default)]
+pub struct DefaultClock;
+
+impl DefaultClock {
+    /// Create a new `DefaultClock`.
+    pub fn new() -> Self {
```
jlizen, line 59: nit: seems redundant with the `Default` derive?

tower-http/src/trace/clock.rs
jlizen: Also without `Copy` we can probably add some blanket implementations:
```
impl<C: Clock + ?Sized> Clock for &C, for Arc<C>, for Box<C>
```

tower-http/src/on_early_drop/early_drops_as_failures.rs
```diff
@@ -25,30 +24,52 @@ use tracing::Span;
 ///
 /// [`OnEarlyDropLayer`]: super::OnEarlyDropLayer
 #[derive(Debug, Clone, Copy)]
-pub struct EarlyDropsAsFailures<F> {
+pub struct EarlyDropsAsFailures<F, Clk = DefaultClock>
+where
+    Clk: Clock,
```
jlizen, line 29: I don't think we need this bound here?

tower-http/src/on_early_drop/early_drops_as_failures.rs
```diff
@@ -25,30 +24,52 @@ use tracing::Span;
 ///
 /// [`OnEarlyDropLayer`]: super::OnEarlyDropLayer
 #[derive(Debug, Clone, Copy)]
-pub struct EarlyDropsAsFailures<F> {
+pub struct EarlyDropsAsFailures<F, Clk = DefaultClock>
```
jlizen, line 27: nit: I would probably prefer either `C` or `Clock`

tower-http/src/trace/mod.rs
```diff
@@ -852,6 +854,226 @@ mod tests {
...
+        let _res = svc
+            .ready()
+            .await
+            .unwrap()
+            .call(Request::new(Body::from("foobar")))
+            .await
+            .unwrap();
+
+        assert!(
+            ON_RESPONSE_LATENCY.load(Ordering::SeqCst) > 0,
```
jlizen, line 989: Here and in other places, we should be able to use exact assertions, right? A regular clock would also pass this, but exact assertions distinguish that we are using the deterministic clock.

tower-http/src/trace/mod.rs
```diff
@@ -852,6 +854,226 @@ mod tests {
...
+            self.0.fetch_add(1, Ordering::SeqCst)
+        }
+
+        fn elapsed(&self, instant: Self::Instant) -> Self::Duration {
+            DummyDuration(self.0.load(Ordering::SeqCst) - instant)
+        }
+    }
+
+    #[test]
+    fn dummy_clock_increments_and_elapsed() {
```
jlizen, line 898: This is just testing the fixture, I'd ditch it.

tower-http/src/trace/mod.rs
```diff
@@ -852,6 +854,226 @@ mod tests {
...
+        let d = clock.elapsed(t1);
+        assert_eq!(
+            d,
+            DummyDuration(2),
+            "elapsed from t1 at counter=3 should be 2"
+        );
+    }
+
+    #[test]
+    fn dummy_clock_latency_display() {
```
jlizen, line 923: Also would ditch this if we are dropping `DurationExt`

tower-http/src/on_early_drop/early_drops_as_failures.rs
```diff
@@ -25,30 +24,52 @@ use tracing::Span;
...
+        }
+    }
+}
+
+impl<F, Clk> EarlyDropsAsFailures<F, Clk>
+where
+    Clk: Clock,
+{
+    /// Wrap an [`OnFailure`] implementation with a custom [`Clock`].
+    pub fn with_clock(on_failure: F, clock: Clk) -> Self {
```
jlizen, line 50: This appears untested, there is enough plumbing with the callbacks that it would be good to test.

jlizen (changes requested):

Saplyn:
Got it, will work on this tomorrow and hopefully finish it within one week

jlizen:
@Saplyn think you'll have time for this in the coming week? No problem if not, just I'll wait on cutting a new release if you're trying to make it in.

Saplyn:
@jlizen Don't think I'd be able to make it, guess I'll aim for the next release after the coming one. Thanks for your patient tho.

PR #711, opened 2026-07-08
docs(example)/custom future with multiple bodies

Reza-Darius:
This PR builds on [this PR](https://github.com/tower-rs/tower-http/pull/709), which explores idomatic solutions to dealing with multiple HTTP bodies in tower services

## Motivation

@Isvane did a great job with his proposal, so i wanted to iterate on it by bringing it closer to how i solved this problem in my own project. Primarily, by introducing a custom future.

## Changes

- removed `BoxBody` alias
- removed some `service` trait bounds
- added helper functions
- replaced `Pin Box Future` with custom future.
- added asserts to `main()`

## Discussion

While i think this is a decently elegant solution, it could be argued its not idiomatic if you compare how this problem is solved in `tower-http/limit`. 

A potential foot gun im wondering about is the trait bound on `Either` which states that both arms need to have the same underlying `Data` type. Right now the example leaves the response body of the inner service entirely generic, but this could lead to a nasty compile error down the road should the `Data` types mismatch.

`tower-http/limit` imposes a `Body<Data = Bytes>` bound on its `Body` implementation for it's wrapper type and im wondering whether this example should include it too, and if so, where.

### DIFF
--- a/examples/custom-future-multiple-bodies/Cargo.toml
+++ b/examples/custom-future-multiple-bodies/Cargo.toml
@@ -9,8 +9,8 @@ license = "MIT"
 [dependencies]
 bytes = "1.12.0"
 http = "1.4.2"
+http-body = "1.0.0"
 http-body-util = "0.1.3"
-hyper = "1.10.1"
 pin-project-lite = "0.2.17"
 tokio = { version = "1.32.0", features = ["full"] }
 tower = { version = "0.5", features = ["full"] }
--- a/examples/custom-future-multiple-bodies/Cargo.toml
+++ b/examples/custom-future-multiple-bodies/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "custom-future-multiple-bodies"
+version = "0.1.0"
+authors = ["Tower Maintainers <team@tower-rs.com>"]
+edition = "2021"
+publish = false
+license = "MIT"
+
+[dependencies]
+bytes = "1.12.0"
+pin-project-lite = "0.2.17"
+hyper = "1.10.1" 
+http = "1.4.2"
+http-body-util = "0.1.3"
+tokio = { version = "1.32.0", features = ["full"] }
+tower = { version = "0.5", features = ["full"] }
--- a/examples/custom-future-multiple-bodies/Cargo.toml
+++ b/examples/custom-future-multiple-bodies/Cargo.toml
@@ -8,9 +8,9 @@ license = "MIT"
 
 [dependencies]
 bytes = "1.12.0"
-pin-project-lite = "0.2.17"
-hyper = "1.10.1" 
 http = "1.4.2"
 http-body-util = "0.1.3"
+hyper = "1.10.1"
+pin-project-lite = "0.2.17"
 tokio = { version = "1.32.0", features = ["full"] }
 tower = { version = "0.5", features = ["full"] }
--- a/examples/custom-future-multiple-bodies/README.md
+++ b/examples/custom-future-multiple-bodies/README.md
@@ -8,7 +8,7 @@ This requires wrapping the response body if the user wishes to leave the inner
 service's body untouched.
 
 This solution is geared towards application code, library author's should
-consider studying [tower_http/limit's body and future implementation](https://github.com/tower-rs/tower-http/tree/main/tower-http/src/limit).
+consider studying [tower_http/limit's body and future implementations instead](https://github.com/tower-rs/tower-http/tree/main/tower-http/src/limit).
 
 ## Running the example
 
--- a/examples/custom-future-multiple-bodies/README.md
+++ b/examples/custom-future-multiple-bodies/README.md
@@ -1,0 +1,14 @@
+# Custom future with multiple bodies
+
+This example serves to demonstrate how to build a service that returns a
+response in case of an event (in this instance, a missing HTTP header) while
+leaving the inner service's response untouched.
+
+This requires wrapping the response body if the user wishes to leave inner
+sevice's body untouched.
+
+## Running the example
+
+```
+cargo run -p custom-future-multiple-bodies
+```
--- a/examples/custom-future-multiple-bodies/README.md
+++ b/examples/custom-future-multiple-bodies/README.md
@@ -4,8 +4,8 @@ This example serves to demonstrate how to build a service that returns a
 response in case of an event (in this instance, a missing HTTP header) while
 leaving the inner service's response untouched.
 
-This requires wrapping the response body if the user wishes to leave inner
-sevice's body untouched.
+This requires wrapping the response body if the user wishes to leave the inner
+service's body untouched.
 
 ## Running the example
 
--- a/examples/custom-future-multiple-bodies/README.md
+++ b/examples/custom-future-multiple-bodies/README.md
@@ -1,11 +1,14 @@
 # Custom future with multiple bodies
 
-This example serves to demonstrate how to build a service that returns a
-response in case of an event (in this instance, a missing HTTP header) while
-leaving the inner service's response untouched.
+This example serves to demonstrate how to build a middleware service that
+returns a concrete response in case of an event (in this instance, a missing
+HTTP header) while leaving the inner service's response untouched.
 
 This requires wrapping the response body if the user wishes to leave the inner
 service's body untouched.
+
+This solution is geared towards application code, library author's should
+consider studying [tower_http/limit's body and future implementation](https://github.com/tower-rs/tower-http/tree/main/tower-http/src/limit).
 
 ## Running the example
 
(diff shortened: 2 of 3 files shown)


Isvane:
Thanks so much for iterating on this! ngl im still learning so custom futures and tower trait bounds are still kinda wizardry to me lol. your version looks way cleaner and more idiomatic.

Should i just close my pr so we can use yours?

examples/custom-future-multiple-bodies/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+authors = ["Tower Maintainers <team@tower-rs.com>"]
+edition = "2021"
+publish = false
+license = "MIT"
+
+[dependencies]
+bytes = "1.12.0"
+http = "1.4.2"
+http-body-util = "0.1.3"
+hyper = "1.10.1"
```
jlizen: hyper is not needed here

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        Ok::<_, std::convert::Infallible>(Response::new(Full::new(Bytes::from("Hello, World!"))))
+    });
+
+    let mut service = ServiceBuilder::new()
+        .layer_fn(|inner| RequireHeader::new(inner, "x-api-key"))
+        .service(inner_service);
+
+    let req_bad = Request::builder().body(Full::<Bytes>::default()).unwrap();
+    let res_bad = service.ready().await?.call(req_bad).await?;
+    assert_eq!(res_bad.status(), StatusCode::BAD_REQUEST);
```
jlizen: Probably these assertions should be in test modules? That matches the other examples.

I'd also probably add some `println!`s so that `cargo run -p custom-future-multiple-bodies` prints something, given the README guides the user to run it.

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl<S> RequireHeader<S> {
+    pub fn new(inner: S, header_name: &'static str) -> Self {
+        Self { inner, header_name }
+    }
+}
+
+impl<S, ReqBody, ResBody> Service<Request<ReqBody>> for RequireHeader<S>
+where
+    S: Service<Request<ReqBody>, Response = Response<ResBody>>,
```
jlizen, line 42: I would probably add `ResBody: http_body::Body<Data = Bytes>` here, yeah, at the cost of taking the `http-body = "1"` dependency. That moves any error out to the middleware layer, which is clearer.

examples/custom-future-multiple-bodies/README.md
```diff
@@ -0,0 +1,14 @@
+# Custom future with multiple bodies
```
jlizen, line 1: Should we mention that the `Either` is convenient but loses the concrete error type, and that you can refer to the `tower_http::limit` wrapper to preserve the type?

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        RequireHeaderFuture::Ok {
+            fut: self.inner.call(req),
+        }
+    }
+}
+
+pin_project! {
+    #[project = EnumProj]
+    pub enum RequireHeaderFuture<F> {
+        Ok{ #[pin] fut: F },
```
jlizen: nit: `Future` to match the repo conventions?

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+        HeaderNotFound,
```
jlizen: nit: `MissingHeader` to match the repo conventions?

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+            return RequireHeaderFuture::HeaderNotFound;
+        }
+        RequireHeaderFuture::Ok {
+            fut: self.inner.call(req),
+        }
+    }
+}
+
+pin_project! {
+    #[project = EnumProj]
```
jlizen: nit: `ResponseFutureProj` to match the repo conventions?

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub enum RequireHeaderFuture<F> {
```
jlizen, line 67: Would it be worth a doc comment pointing to how eg `limit/future.rs` encapsulate these types for semver safety, if used in a library? This seems fine for application usage.

jlizen: I don't particularly mind demonstrating the encapsulated pattern in general, it's not particularly more verbose, up to you.

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl<F, E, ResBody> Future for RequireHeaderFuture<F>
+where
+    F: Future<Output = Result<Response<ResBody>, E>>,
+{
+    type Output = Result<Response<ResponseBody<ResBody>>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project() {
+            EnumProj::Ok { fut } => {
```
jlizen: More readable as:
```
  fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
      let res = match self.project() {
              // `?` propagates an inner-service error. to turn it into a response
              // instead, replace `?` with
              //     .unwrap_or_else(|_| rejection(StatusCode::BAD_GATEWAY, "..."))
              // to convert every error, or `.or_else(..)` to convert only some and
              // keep propagating the rest.
          ResFutProj::Future { future } => ready!(future.poll(cx))?.map(Either::Left),
          ResFutProj::MissingHeader => rejection(StatusCode::BAD_REQUEST, "missing header"),
      };

      Poll::Ready(Ok(res))
  

```

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            EnumProj::HeaderNotFound => {
+                Poll::Ready(Ok(new_err_resp(StatusCode::BAD_REQUEST, "missing header")))
+            }
+        }
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
jlizen: We need the unwrap here since we don't have `Box<dyn Error>: From<Box<dyn Error + Send + Sync>>`.

I'd suggest instead widening the signature to `Result<(), Box<dyn Error + Send + Sync>>` (or just `Result<(), tower::BoxError`>, which is equivalent) to avoid both the unwrap and the need to coerce explicitly to work around the missing `From`.

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+// think of Either as an enum that implements Body if both arms implement Body
+// this allows us to combine multple bodies
+type ResponseBody<B> = Either<B, Full<Bytes>>;
+
+// some helper to go along with it
+fn map_resp<B>(resp: Response<B>) -> Response<ResponseBody<B>> {
+    resp.map(|body| Either::Left(body))
+}
+
+fn new_err_resp<B>(status: StatusCode, body: &'static str) -> Response<ResponseBody<B>> {
```
jlizen: nit: maybe call it `rejection`?

examples/custom-future-multiple-bodies/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+use std::error::Error;
+use tower::ServiceBuilder;
+use tower::ServiceExt;
+
+// think of Either as an enum that implements Body if both arms implement Body
+// this allows us to combine multple bodies
+type ResponseBody<B> = Either<B, Full<Bytes>>;
+
+// some helper to go along with it
+fn map_resp<B>(resp: Response<B>) -> Response<ResponseBody<B>> {
```
jlizen: I don't think this really carries its weight as a helper

jlizen (changes requested):

examples/custom-future-multiple-bodies/README.md
Reza-Darius: I dont quite understand what this means, doesnt the body stay fully preserved including the error?

jlizen: Ah sorry for the slow response.

 The body itself is preserved, but the error type is erased. `Either`'s `Body` impl declares `type Error = Box<dyn Error + Send + Sync>`. So whoever consumes the response body afterwards only sees a boxed trait object instead of, say, `hyper::Error`, and would have to downcast to inspect it.
 
Compare with `tower_http::limit's ResponseBody<B>`, which declares `type Error = B::Error` and passes through the original error untouched.

For an example this tradeoff is usually fine, I'd just like the README to mention it so readers know the alternative exists.

jlizen (approved):
Thanks for the contribution! looks great

PR #714, opened 2026-07-15
Remove Clone derivations from SetMultipleResponseHeader* types

### LINKED ISSUE #713: SetMultipleResponseHeaders still creates a clone constraint on the response body

## Bug Report

#703 's description and the changelog both say that this was fixed for Request and Response types, but the PR only actually applied the fix to the request types. For me this means that in axum I still can't use `SetMultipleResponseHeadersLayer` because axum's `Router:: layer` requires `L: Layer<Route> + Clone + Sync + 'static`.

### Version

0.7.0, main

### Platform

Darwin mac.lan 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:15 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6041 arm64 arm Darwin

### Description

Here's a reproduction:

```rust
use axum::{Router, routing::get, http::{header, HeaderValue}};
use tower_http::set_header::response::SetMultipleResponseHeadersLayer;

fn build() -> Router {
    Router::new()
        .route("/", get(|| async { "ok" }))
        .layer(SetMultipleResponseHeadersLayer::overriding(vec![
            (header::X_CONTENT_TYPE_OPTIONS, HeaderValue::from_static("nosniff")).into(),
        ]))
}
```
```console
error[E0277]: the trait bound `Response<Body>: Clone` is not satisfied
   --> src/main.rs:7:16
    |
  7 |           .layer(SetMultipleResponseHeadersLayer::overriding(vec![
    |  __________-----_^
    | |          |
    | |          required by a bound introduced by this call
  8 | |             (header::X_CONTENT_TYPE_OPTIONS, HeaderValue::from_static("nosniff")).into(),
  9 | |         ]))
    | |__________^ the trait `Clone` is not implemented for `Response<Body>`
    |
    = note: required for `Response<Body>` to implement `Clone`
    = note: 1 redundant requirement hidden
    = note: required for `SetMultipleResponseHeadersLayer<Response<Body>>` to implement `Clone`
note: required by a bound in `Router::<S>::layer`
   --> /Users/mumu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/axum-0.8.9/src/routing/mod.rs:305:27
    |
303 |     pub fn layer<L>(self, layer: L) -> Router<S>
    |            ----- required by a bound in this associated function
304 |     where
305 |         L: Layer<Route> + Clone + Send + Sync + 'static,
    |                           ^^^^^ required by this bound in `Router::<S>::layer`
help: consider borrowing here
    |
  7 |         .layer(&SetMultipleResponseHeadersLayer::overriding(vec![
    |                +
```

skeet70:
## Motivation

This allows downstream consumers to not be impacted by uneccesary Clone bounds.

Includes a regression test that simplifies the reproduction case, fails to compile without the fix, and succeeds with it.

## Solution

Apply the same fix as #703 to Response types as well, using the more narrow and accurate Clone definition vs the derived Clone one.

Fixes: #713

### DIFF
--- a/tower-http/src/set_header/response/multiple_headers.rs
+++ b/tower-http/src/set_header/response/multiple_headers.rs
@@ -18,9 +18,16 @@ use crate::set_header::{HeaderInsertionConfig, HeaderMetadata, InsertHeaderMode}
 /// Layer that applies [`SetMultipleResponseHeader`] which adds multiple response headers.
 ///
 /// See [`SetMultipleResponseHeader`] for more details.
-#[derive(Clone)]
 pub struct SetMultipleResponseHeadersLayer<M> {
     headers: Vec<HeaderInsertionConfig<M>>,
+}
+
+impl<M> Clone for SetMultipleResponseHeadersLayer<M> {
+    fn clone(&self) -> Self {
+        Self {
+            headers: self.headers.clone(),
+        }
+    }
 }
 
 impl<M> fmt::Debug for SetMultipleResponseHeadersLayer<M> {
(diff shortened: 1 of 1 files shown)


jplatte (approved):

jplatte (approved):

jlizen (approved):
Thanks!

PR #716, opened 2026-08-08
feat(request-id): Allow request-id constructor in const context

tottoto:
Allows `request-id` constructors in `const` context.


jplatte (approved):

jlizen (approved):
Thanks!

PR #718, opened 2026-08-12
Propagate ServeDir::try_call I/O errors

### LINKED ISSUE #412: `ServeDir::try_call` does not propagate result as documented
### Version
`tower-http v0.4.4`
### Platform
All

### Description

When using `ServeDir::try_call` to request a path that does not exist, it returns `Ok(404)` instead of an `Err`

```rust
use axum::http::Request;
use tracing::{error, info};

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt::init();

    let mut serve_dir = ServeDir::new("src");

    let mut req = Request::new(());
    *req.uri_mut() = "/main.rs".parse().unwrap();
    match serve_dir.try_call(req).await {
        Ok(response) => info!("response code: {}", response.status()),
        Err(e) => error!("error: {}", e),
    }

    let mut req2 = Request::new(());
    *req2.uri_mut() = "/doesnotexist.rs".parse().unwrap();
    match serve_dir.try_call(req2).await {
        Ok(response) => info!("response code: {}", response.status()),
        Err(e) => error!("error: {}", e),
    }
}
```
Expected: 
```
2023-09-23T18:09:09.141584Z  INFO rs_playground: response code: 200 OK
2023-09-23T18:09:09.141801Z  ERROR rs_playground: error: <something>
```
Actual:
```
2023-09-23T18:09:09.141584Z  INFO rs_playground: response code: 200 OK
2023-09-23T18:09:09.141801Z  INFO rs_playground: response code: 404 Not Found
```

Boulea7:
## Motivation

`ServeDir::try_call` documents that callers can handle I/O errors themselves, but expected filesystem errors are currently converted into `404 Not Found` responses before the caller can inspect them.

Fixes #412.

## Solution

- Propagate expected filesystem I/O errors from `try_call` when no fallback is configured.
- Keep existing fallback and invalid-path behavior unchanged.
- Let the infallible `Service::call` implementation convert NotFound, PermissionDenied, and Unix ENOTDIR errors to 404 responses, and other I/O errors to 500 responses.
- Preserve error-level tracing only for errors converted to 500 responses.

## Tests

- `cargo test --offline -p tower-http --all-features -- --test-threads=1`
- `cargo check --offline --workspace --all-features --all-targets`
- `cargo clippy --offline --workspace --all-features --lib --bins --examples -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/tower-http/src/services/fs/serve_dir/future.rs
+++ b/tower-http/src/services/fs/serve_dir/future.rs
@@ -144,24 +144,11 @@ where
                     }
 
                     Err(err) => {
-                        #[cfg(unix)]
-                        // 20 = libc::ENOTDIR => "not a directory
-                        // when `io_error_more` landed, this can be changed
-                        // to checking for `io::ErrorKind::NotADirectory`.
-                        // https://github.com/rust-lang/rust/issues/86442
-                        let error_is_not_a_directory = err.raw_os_error() == Some(20);
-                        #[cfg(not(unix))]
-                        let error_is_not_a_directory = false;
-
-                        if matches!(
-                            err.kind(),
-                            io::ErrorKind::NotFound | io::ErrorKind::PermissionDenied
-                        ) || error_is_not_a_directory
-                        {
+                        if super::should_return_not_found(&err) {
                             if let Some((mut fallback, request)) = fallback_and_request.take() {
                                 call_fallback(&mut fallback, request)
                             } else {
-                                break Poll::Ready(Ok(not_found()));
+                                break Poll::Ready(Err(err));
                             }
                         } else {
                             break Poll::Ready(Err(err));
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -323,9 +323,10 @@ impl<F, B: Backend> ServeDir<F, B> {
     ///
     /// By default `<ServeDir as Service<_>>::call` will handle IO errors and convert them into
     /// responses. It does that by converting [`std::io::ErrorKind::NotFound`] and
-    /// [`std::io::ErrorKind::PermissionDenied`] to `404 Not Found` and any other error to `500
-    /// Internal Server Error`. The error will also be logged with `tracing` in case the `tracing`
-    /// crate feature is enabled.
+    /// [`std::io::ErrorKind::PermissionDenied`] to `404 Not Found`. On Unix, errors indicating
+    /// that a path component is not a directory are also converted to `404 Not Found`. Any other
+    /// error is converted to `500 Internal Server Error` and will also be logged with `tracing` in
+    /// case the `tracing` crate feature is enabled.
     ///
     /// If you want to manually control how the error response is generated you can make a new
     /// service that wraps a `ServeDir` and calls `try_call` instead of `call`.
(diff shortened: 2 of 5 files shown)


tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -323,9 +323,10 @@ impl<F, B: Backend> ServeDir<F, B> {
...
-    /// [`std::io::ErrorKind::PermissionDenied`] to `404 Not Found` and any other error to `500
-    /// Internal Server Error`. The error will also be logged with `tracing` in case the `tracing`
-    /// crate feature is enabled.
+    /// [`std::io::ErrorKind::PermissionDenied`] to `404 Not Found`. On Unix, errors indicating
+    /// that a path component is not a directory are also converted to `404 Not Found`. Any other
+    /// error is converted to `500 Internal Server Error` and will also be logged with `tracing` in
+    /// case the `tracing` crate feature is enabled.
     ///
     /// If you want to manually control how the error response is generated you can make a new
     /// service that wraps a `ServeDir` and calls `try_call` instead of `call`.
```
jlizen, line 337: `try_call` now is somewhat lossy since the user needs to re-implement the "should this be a 404" check themself if all they want is a custom body.

Is it worth exposing a public helper? That feels a bit klunky... an alternative would be that the example could demonstrate the same check inline?

jlizen: Also... this is arguably a breaking behavioral change, since if somebody had copied this example, previously they would be returning `Ok(404)` but now they would return a 500.

I would consider it a bugfix that is justifiably breaking, since the caller would have expected to receive an error in the first place based on the contract described in the current docs, but we might want to highlight it in the changelog / do a minor version bump? 

Any opinion @seanmonstar ?

tower-http/src/services/fs/serve_dir/future.rs
jlizen, line 112: According to our read of `try_call`'s API contract, shouldn't we also propagate non-io failures as well rather than casting them to `Ok(404)`?

jlizen: It looks like we also swallow metadata errors in `is_dir()` in `open_file.rs`, which would result in permissions errors routing through `FileNotFound`?

tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -323,9 +323,10 @@ impl<F, B: Backend> ServeDir<F, B> {
...
     /// responses. It does that by converting [`std::io::ErrorKind::NotFound`] and
-    /// [`std::io::ErrorKind::PermissionDenied`] to `404 Not Found` and any other error to `500
-    /// Internal Server Error`. The error will also be logged with `tracing` in case the `tracing`
-    /// crate feature is enabled.
+    /// [`std::io::ErrorKind::PermissionDenied`] to `404 Not Found`. On Unix, errors indicating
+    /// that a path component is not a directory are also converted to `404 Not Found`. Any other
+    /// error is converted to `500 Internal Server Error` and will also be logged with `tracing` in
+    /// case the `tracing` crate feature is enabled.
     ///
     /// If you want to manually control how the error response is generated you can make a new
```
jlizen, line 336: We should also mention that you only will get expected errors bubbled up if you have no fallback service configured? Not clear that is the case from these docs (though I think it's probably the right API).

tower-http/src/services/fs/serve_dir/tests.rs
```diff
@@ -351,6 +351,22 @@ async fn not_found() {
     assert!(body.is_empty());
 }
 
+#[tokio::test]
```
jlizen, line 354: We should also have tests for:
- not found with a fallback service
- `ENOTDIR` case on unix
- `call` returns 500 for an unexpected error (I think that is easier to test now that we have a `Backend` trait to use?

tower-http/src/services/fs/serve_dir/future.rs
jlizen: Reading the docs again, it looks like `try_call` only claims to bubble up i/o errors, so perhaps these path shape ones are good to leave as-is, though it might be worth a note stating that explicitly?

Boulea7: I changed the directory metadata probe to keep treating NotFound and invalid filenames as absent, but return other metadata errors from try_call. A custom Backend test covers PermissionDenied from metadata. Path-shape failures still remain responses as documented.

tower-http/src/services/fs/serve_dir/tests.rs
Boulea7: Added coverage for not-found with a fallback, Unix ENOTDIR through try_call, and an unexpected Backend open error mapping to 500 through Service::call. The full tower-http test and doctest suites pass.

tower-http/src/services/fs/serve_dir/mod.rs
Boulea7: I updated the example to preserve the previous 404 behavior for NotFound, PermissionDenied, and Unix ENOTDIR, while still using 500 for unexpected I/O errors. The method docs now also state that invalid paths and unsupported methods remain responses.

tower-http/src/services/fs/serve_dir/mod.rs
Boulea7: The docs now call out that expected not-found I/O errors invoke the configured fallback and only bubble out when no fallback is present.

jlizen (approved):
Thanks for the contribution!

PR #722, opened 2026-08-13
fix(decompression): don't end the body on an empty data frame

clemenslosbichler-cloud:
Fixes the issue discussed in #719.

After the decompressor reports end-of-stream, #712 added an arm that errors when a following frame still carries bytes. A data frame with zero remaining bytes isn't caught by that check, so it fell through to `Poll::Ready(None)` and ended the body:

```rust
Some(Ok(frame)) => {
    if let Ok(bytes) = frame.into_data() {
        if bytes.has_remaining() {
            return Poll::Ready(Some(Err(/* extra bytes */)));
        }
    }
    Poll::Ready(None)   // <- empty data frame lands here
}
```

Trailers following such a frame were dropped. Because `Body`'s contract forbids polling once `None` has been returned, they couldn't be recovered either. Before #712 the frame was forwarded and polling continued.

This forwards the empty frame instead and keeps polling. The extra-bytes error from #712 is unchanged. Frames that are neither data nor trailers are now forwarded rather than swallowed, for the same reason.

Added `brotli_keeps_trailers_that_follow_an_empty_data_frame`, which yields the compressed payload, then `Frame::data(Bytes::new())`, then trailers, and asserts both the body and the trailers arrive. It fails on the current `main` with "trailers following an empty data frame should still arrive" and passes with this change.

### DIFF
--- a/tower-http/src/compression_utils.rs
+++ b/tower-http/src/compression_utils.rs
@@ -241,16 +241,24 @@ where
             Some(Ok(frame)) if frame.is_trailers() => Poll::Ready(Some(Ok(
                 frame.map_data(|mut data| data.copy_to_bytes(data.remaining()))
             ))),
-            Some(Ok(frame)) => {
-                if let Ok(bytes) = frame.into_data() {
-                    if bytes.has_remaining() {
-                        return Poll::Ready(Some(Err(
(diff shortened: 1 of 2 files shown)


jlizen (approved):
Looks great! Appreciate the fix

PR #727, opened 2026-08-22
Allow ignoring unsupported multi-range requests

### LINKED ISSUE #725: Ignore unsupported multi-range requests rather than erroring
Follow-up to #661.

`ServeDir` and `ServeFile` return `416` for valid multi-range requests because multipart responses are unsupported. `416` is misleading when the ranges are satisfiable, and the request could be served normally [according to the RFC](https://www.rfc-editor.org/rfc/rfc9110.html#section-14.2) by ignoring `Range`. 

Suggested change:
For multi-range requests, ignore `Range` and return the full representation with `200`. 

This should be opt-in within the current minor version because it changes existing `416` responses to full-file `200` responses, which can increase bandwidth unexpectedly.

Handle these cases:

- Keep `GET` and `HEAD` status and headers consistent.
- Define when malformed, overlapping, reversed, or unsatisfiable ranges are ignored versus rejected with `416`.
- Do not retain representation headers such as `Content-Type` or `Content-Encoding` on an empty or plain-text `416` body, including for precompressed files.

BreezeDelegate:
## Motivation

`ServeDir` and `ServeFile` cannot produce multipart range responses. Returning `416` for a syntactically valid multi-range request is avoidable because RFC 9110 allows a server to ignore `Range` and serve the full representation.

Fixes #725.

## Solution

Add an opt-in `ignore_multi_range_requests` setting to `ServeDir` and `ServeFile`. When enabled, syntactically valid multi-range headers are ignored before semantic range validation and the full representation is returned with `200`; malformed headers and unsatisfiable single ranges keep returning `416`.

The default remains unchanged. GET and HEAD behavior is covered, as are overlapping/reversed multi-ranges. `416` responses also stop retaining the selected representation's `Content-Type` and `Content-Encoding`, including for precompressed files.

### DIFF
--- a/tower-http/src/services/fs/serve_dir/future.rs
+++ b/tower-http/src/services/fs/serve_dir/future.rs
@@ -284,19 +284,29 @@ fn build_response(output: FileOpened) -> Response<ResponseBody> {
                 .unwrap()
         }
 
-        Some(Err(RangeError::MultipleRangesNotSupported)) => builder
-            .header(header::CONTENT_RANGE, format!("bytes */{}", size))
-            .status(StatusCode::RANGE_NOT_SATISFIABLE)
-            .body(body_from_bytes(Bytes::from(
-                "Cannot serve multipart range requests",
-            )))
-            .unwrap(),
+        Some(Err(RangeError::MultipleRangesNotSupported)) => {
+            let mut response = builder
+                .header(header::CONTENT_RANGE, format!("bytes */{}", size))
+                .status(StatusCode::RANGE_NOT_SATISFIABLE)
+                .body(body_from_bytes(Bytes::from(
+                    "Cannot serve multipart range requests",
+                )))
+                .unwrap();
+            response.headers_mut().remove(header::CONTENT_TYPE);
+            response.headers_mut().remove(header::CONTENT_ENCODING);
+            response
+        }
 
-        Some(Err(RangeError::Unsatisfiable)) => builder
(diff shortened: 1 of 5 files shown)


tower-http/src/services/fs/serve_dir/mod.rs
```diff
@@ -190,6 +194,20 @@ impl<F, B: Backend> ServeDir<F, B> {
         self
     }
 
+    /// Configure whether syntactically valid multi-range requests should be ignored.
+    ///
+    /// When enabled, a request containing multiple byte ranges is served as a normal full
```
jlizen, line 199: Good docs, thanks.

tower-http/src/services/fs/serve_dir/future.rs
```diff
@@ -284,19 +284,29 @@ fn build_response(output: FileOpened) -> Response<ResponseBody> {
...
+            response
+        }
+
+        Some(Err(RangeError::Unsatisfiable)) => {
+            let mut response = builder
+                .header(header::CONTENT_RANGE, format!("bytes */{}", size))
+                .status(StatusCode::RANGE_NOT_SATISFIABLE)
+                .body(empty_body())
+                .unwrap();
+            response.headers_mut().remove(header::CONTENT_TYPE);
```
jlizen, line 306: Thanks for fixing this

jlizen (approved):
Appreciate it!

PR #728, opened 2026-08-24
feat(services): configure directory redirects

### LINKED ISSUE #542: Handling trailing slashes for directories
## Feature Request

 I'm willing to implement this if approved

### Motivation

I want to support only non-trailing slash paths. To do this I wrap my router with `NormalizePathLayer::trim_trailing_slash().layer`. This however doesn't work well with `ServeDir` which will always redirect to a trailing slash path for directories (`/kita` -> `/kita/`). Therefore, it's not possible not to have trailing slashes on paths that correspond to directories.

### Proposal

There are 3 possibilities on how to handle paths without a trailing slash for directories:
1. redirect to trailing slash
2. return `index.html` (200)
3. respond with 404 (not found)

I think an enum of 3 variants should be offered with current behavior (redirect) as the default. Additionally, I think that the 3rd variant would make the `append_index_html_on_directories` deprecated as it would have the same effect

### Alternatives

A solution where I would rewrite URL in a layer before this service would have to access filesystem to verify that path corresponds to a directory. This would incur unnecessary IO reads so it would be best to handle it directly within `ServeDir`

I'm eager to hear other proposals on how to design this API

BreezeDelegate:
## Motivation

`ServeDir` always redirects `/dir` to `/dir/`, so directory indexes cannot be served directly at no-slash paths. This is the use case in #542. @mversic had also prototyped the no-redirect behavior on their fork.

## Solution

Add `redirect_to_trailing_slash(bool)`, defaulting to `true`. When disabled, `/dir` serves `/dir/index.html` directly when directory indexes are enabled. Existing redirect and 404 behavior stays unchanged.

Tested with the workspace all-feature test suite, the CI clippy command, and rustfmt.

Fixes #542

### DIFF
--- a/tower-http/src/services/fs/serve_dir/mod.rs
+++ b/tower-http/src/services/fs/serve_dir/mod.rs
@@ -82,6 +82,7 @@ impl ServeDir<DefaultServeDirFallback> {
             precompressed_variants: None,
             variant: ServeVariant::Directory {
                 append_index_html_on_directories: true,
+                redirect_to_trailing_slash: true,
                 html_as_default_extension: false,
             },
             fallback: None,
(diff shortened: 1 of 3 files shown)


jlizen (approved):
Looks great! Thanks for the contribution. Will work on getting out a release, hopefully this week.

PR #734, opened 2026-09-06
fix(compression): avoid re-polling body after EOF in BodyIntoStream

### LINKED ISSUE #732: CompressionLayer re-polls non-fused streaming bodies after EOF and panics

### Version

```
tower-http v0.7.0 and v0.7.1 (feature = "compression-gzip")
```

Reproduced on 0.7.0 and 0.7.1. Source inspection of current `main` shows the relevant `BodyIntoStream::poll_frame` fallback is unchanged, so it appears affected as well.

### Platform

```
Linux fedora 7.1.10-200.fc44.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Aug 23 16:15:11 UTC 2026 x86_64 GNU/Linux
```

Also reproduced on a Fedora x86_64 server (VPS).

### Crates

`compression` (`Compression` / `CompressionLayer`, `compression-gzip` feature)

### Description

Gzip-compressing a streaming response body built from `Body::from_stream(futures::stream::unfold(..))` panics when the source stream reaches EOF, after the gzip encoder has produced the body bytes.

I tried this code:

```rust
// tower-http = { version = "=0.7.0", features = ["compression-gzip"] }
// The same reproducer also fails with version = "=0.7.1".
// futures = "0.3.34"
// axum = "0.8.9"
// http-body = "1.1.0"
// http-body-util = "0.1.5"
// async-compression = "0.4.43"
// tokio = "1.53.1" (features = ["rt-multi-thread", "macros", "sync", "time"])
// tower = "0.5.3"
// bytes = "1.12.1"

use axum::body::{Body, to_bytes};
use axum::http::{Request, Response};
use bytes::Bytes;
use tower::ServiceExt;
use tower_http::compression::Compression;

#[tokio::test]
async fn repro_streaming_gzip_panic() {
    let svc = Compression::new(tower::service_fn(|_req: Request<Body>| async {
        let (tx, rx) = tokio::sync::mpsc::channel::<Result<Bytes, std::io::Error>>(64);
        tokio::spawn(async move {
            let chunk = Bytes::from(vec![b'x'; 64]);
            for _ in 0..100 {
                if tx.send(Ok(chunk.clone())).await.is_err() {
                    return;
                }
            }
            tokio::time::sleep(std::time::Duration::from_millis(15)).await;
            // sender dropped here -> the unfold returns None
        });
        let body = Body::from_stream(futures::stream::unfold(rx, |mut rx| async move {
            rx.recv().await.map(|item| (item, rx))
        }));
        Ok::<_, std::convert::Infallible>(
            Response::builder()
                .header("content-type", "application/json")
                .body(body)
                .unwrap(),
        )
    }));

    let req = Request::builder()
        .header("accept-encoding", "gzip")
        .body(Body::empty())
        .unwrap();
    let resp = svc.oneshot(req).await.unwrap();
    // Panics here:
    let _ = to_bytes(Body::new(resp.into_body()), usize::MAX).await.unwrap();
}
```

I expected the response body to emit gzip data, then a valid footer, and end cleanly at the HTTP-body level.

Instead, this happened:

```
thread 'tokio-runtime-worker' panicked at .../futures-util-0.3.34/src/stream/unfold.rs:108:21:
Unfold must not be polled after it returned `Poll::Ready(None)`
```

Reproduces 100% of the time with this compressed non-fused stream-backed body; buffered bodies are unaffected.

`WrapBody::poll_frame` polls the underlying body after the gzip encoder reaches EOF in order to forward remaining frames. In this case, that eventually causes `BodyIntoStream::poll_frame` to re-poll an underlying body that has already returned `None`.

There is an interface mismatch here: `http-body` recommends that `Body` implementations remain pollable after EOF, while `StreamBody` delegates to an arbitrary `Stream`, whose contract permits panicking after `None`. In this reproduction, `BodyIntoStream::poll_next` has already received `None` from the underlying body's `poll_frame` and sets `yielded_all_data`. However, `yielded_all_data` also represents the distinct case where a non-data frame was encountered, so it cannot by itself mean HTTP-body EOF. When no non-data frame is buffered, `BodyIntoStream::poll_frame` then falls through to `this.body.poll_frame(cx)` after its internal stream returns `None`, which violates the `Body` post-EOF contract for the wrapper when the wrapped body is backed by a non-fused `Stream`. Tracking actual underlying-body exhaustion separately, or otherwise making only the true-EOF path fused, would avoid the re-poll without suppressing trailers or other frames that legitimately follow codec EOF.

In a server, the panic aborts the connection task before the HTTP body can complete normally. In this reproducer, the gzip footer itself has already been emitted, but the transport stream is terminated abnormally; clients report a mid-stream reset (for example, `HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)` from curl, or "The socket connection was closed unexpectedly" from Bun fetch).

In production I hit this on SSE because our custom compression predicate allowed `text/event-stream`; the default tower-http predicate excludes SSE. Observed in production: 1,551 panics with this stack over ~46 hours while compressed SSE responses were enabled; after excluding `text/event-stream` from compression, no further instances were observed.

Related: [#419](https://github.com/tower-rs/tower-http/issues/419) / [#420](https://github.com/tower-rs/tower-http/issues/420) ("SSE buffers under compression", resolved in 0.5.2 by [#465](https://github.com/tower-rs/tower-http/pull/465), which excludes `text/event-stream` from the default predicate). The adjacent post-codec trailer and frame handling was later changed in [#685](https://github.com/tower-rs/tower-http/pull/685) and [#712](https://github.com/tower-rs/tower-http/pull/712); neither guards this non-fused source-body re-poll.

zfaustk:
## Motivation

Closes #732

When compressing a streaming response body backed by a non-fused stream (e.g. `Body::from_stream(futures::stream::unfold(...))`), `Compression` panics at the end of the stream with:
`Unfold must not be polled after it returned Poll::Ready(None)`.

In `BodyIntoStream`, `poll_next` drives the underlying body to read all data frames into the encoder. Once `this.body.poll_frame(cx)` returns `None`, `yielded_all_data` is set. After the encoder finishes writing all compressed output, `WrapBody::poll_frame` calls `BodyIntoStream::poll_frame` to forward any remaining frames (such as trailers).

If no trailers were buffered, `BodyIntoStream::poll_frame` unconditionally fell through to `this.body.poll_frame(cx)`. For bodies backed by non-fused streams, re-polling an already-exhausted body violates the stream contract and panics.

## Solution

Track whether the inner body has already returned `None` via a boolean field (`body_eof`). When `body_eof` is true, `BodyIntoStream::poll_frame` immediately returns `Poll::Ready(None)` instead of re-polling the inner body.


tower-http/src/compression_utils.rs
cratelyn, line 1: if we want to avoid polling the underlying body after it yields `Ready(None)`, i think that it would be more prudent to use [`http_body_util::combinators::Fuse<B>`](https://docs.rs/http-body-util/latest/http_body_util/combinators/struct.Fuse.html).

we can wrap `B` in that combinator, rather than reimplementing the logic for fusing a body here.

tower-http/src/compression_utils.rs
```diff
@@ -287,6 +288,7 @@ where
             body,
             yielded_all_data: false,
             non_data_frame: None,
+            body_eof: false,
```
cratelyn, line 291: it smells weird to me that this introduces a `body_eof` flag when there is already a `yielded_all_data` flag here. how do these differ?

cratelyn (commented):
thank you for opening a pull request @zfaustk. i appreciate that this commit includes a test case too.

i have one suggestion about how we can patch this in a more efficient manner, without introducing this `body_eof` boolean.

tower-http/src/compression_utils.rs
De-ASI-INTERFACE: Fixes #735

De-ASI-INTERFACE (approved):

tower-http/src/compression_utils.rs
zfaustk: Thanks for the suggestion. I checked `Fuse`, but it requires `B: Unpin`, which doesn’t fit this generic body, so I kept the local guard.

tower-http/src/compression_utils.rs
zfaustk: `yielded_all_data` means no more data frames, while `body_eof` means the body returned `None`. Trailers can still follow the data, so these two states need to stay separate. The added test covers this.

